PR #10, opened 2019-07-15
Dynamic method type

soutaro:
# Problem

Some methods defined in Ruby are incompatible with `super` methods, like `.new` or `#initialize`.

```
class A < Object
  def initialize(name)         # A#initialize is incompatible with Object#initialize, and so is .new
    @name = name
  end
end
```

One possible type for `Object#initialize` is `(*any) -> any`. However, it allows passing any arguments to `Object.new`, which receives no argument. It is more general than any arguments but not less general than any other method types.

Examples: [I have defined a `#hash` method](https://github.com/soutaro/strong_json/blob/master/lib/strong_json/types.rb#L103-L105) for a runtime type checking utility for `Hash` object, which is incompatible with `Object#hash` method. [ActiveRecord has `#load` method](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-load), which is also incompatible with `Kernel#load`.

# `any` type

I'd like to propose to add a new method type `any`. It is like `any` type but ranges over method types.

```
class A < Object
  def initialize: (String) -> void
                | any
end
```

The `any` method type is both a super-type and sub-type of any method type.

The reason we need another `any` is that RBS type system has two different spaces of _types of values_ and _types of methods_. We already have `any` type for _types of values_, and the new `any` method type is for _types of methods_. In typical type systems, methods are values so that `?` type covers type of methods too. In RBS type system, `any` type is not for methods and we need another _any type_ for type of methods.

# Type checking

The `any` method type should be placed at the end of method overloadings, which means it has the lowest precedence. Like `any` type, type checkers can produce an (optional) error message to warn developers that the method type resolution may be different from what they want.

```
A.new()                     # It resolves to `any` method type and causes an error
```

### DIFF
--- a/lib/ruby/signature/definition.rb
+++ b/lib/ruby/signature/definition.rb
@@ -39,7 +39,14 @@ module Ruby
         def sub(s)
           self.class.new(
             super_method: super_method&.sub(s),
-            method_types: method_types.map {|ty| ty.sub(s) },
+            method_types: method_types.map do |ty|
+              case ty
+              when MethodType
+                ty.sub(s)
+              when :any
+                ty
+              end
+            end,
             defined_in: defined_in,
             implemented_in: implemented_in,
             accessibility: @accessibility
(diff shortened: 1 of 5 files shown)


soutaro:
Closing.

Implemented `incompatible` attribute on `def` at https://github.com/ruby/ruby-signature/commit/ee29f178a64ab1a88e1388b685095fbddfd5f281.

PR #11, opened 2019-08-09
Suppress syntactic warnings

mame:
`rake` outputs a lot of "warning: `&' interpreted as argument prefix".
To suppress the warnings, this change adds explicit parentheses.


soutaro:
🙏

PR #12, opened 2019-08-09
Add `require "pathname"`

mame:
Many tests fail due to the lack.


soutaro:
Could you put to `ruby/signature.rb`? (I'm not sure why it is working...🤦‍♂)

mame:
Thank you for reviewing, updated.

PR #14, opened 2019-08-09
Use `assert_nil` instead of `assert_equal nil, ...`

mame:
to suppress deprecation warnings:
"DEPRECATED: Use assert_nil if expecting nil from ...  This will fail in Minitest 6."


soutaro:
While I don't understand the benefits of this change in Minitest 6, there seems no way to revert the _improvement_. So, I'll try using `assert_nil`. 🤖

* https://github.com/seattlerb/minitest/issues/666

PR #15, opened 2019-08-09
Do not assume the project repository name as "ruby-signature"

mame:
I'd like to clone my forked repository as "mame-ruby-signature"...


soutaro:
🎉

PR #17, opened 2019-08-09
`Array#permutation` returns self instead of Array[self]

mame:


stdlib/builtin/array.rbi
```diff
@@ -90,8 +90,8 @@ class Array[A]
 
   def pack: (String, ?buffer: String) -> String
 
-  def permutation: (?Integer) { (self) -> any } -> Array[self]
-                 | (?Integer) -> Enumerator[self, Array[self]]
+  def permutation: (?Integer) { (self) -> any } -> self
+                 | (?Integer) -> Enumerator[self, self]
```
soutaro: I think it returns an enumerator for `Array[self]` without block.

mame: `Enumerator[X, Y]`'s `X` is a type that `each` yields, and `Y` is a type that `each` returns, right?
I think that both `X` and `Y` are actually `self` (= `Array[A]`).

soutaro: Ah, right. 💦 
I was confusing with return type of `#to_a`.

PR #18, opened 2019-08-09
Add lacked instance methods of Array

mame:


stdlib/builtin/array.rbi
```diff
@@ -102,6 +106,7 @@ class Array[A]
   def product: (*Array[A]) -> Array[Array[A]]
              | (*Array[A]) { (Array[A]) -> any } -> self
 
+  def assoc: (any) - any
```
soutaro: CI looks failing because of `-`.

mame: argh

mame:
Updated :+1:

PR #24, opened 2019-08-20
Generated from Sorbet RBI

soutaro:
Automatically translated type signatures.

Probably we need to edit the signatures a lot. Several known mismatches.

* ~~`T::Boolean` or `bool`~~ ✅ 
* ~~`BasicObject` or `any`~~ ✅
* Sometimes block type is marked `T.unknown`
* ~~`Enumerable` takes only one type argument.~~ ✅

Anyway, we should review the output carefully.

### DIFF
--- a/stdlib/builtin/array.rbi
+++ b/stdlib/builtin/array.rbi
@@ -318,7 +318,7 @@
 # arr                       #=> [1, 2, 3]
 # ```
 class Array[Elem] < Object
-  include Enumerable
+  include Enumerable[Elem]
 
   def self.[]: [U] (*U arg0) -> ::Array[U]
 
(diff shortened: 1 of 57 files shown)


soutaro:
Class/module type parameter handling looks broken. 😢

lib/ruby/signature/scaffold/rbi.rb
```diff
@@ -498,6 +498,10 @@ def const_to_name(node)
               BuiltinNames::Hash.name
             when "T::Range"
               BuiltinNames::Range.name
+            when "T::Enumerator"
+              BuiltinNames::Enumerator.name
+            when "T::Enumerable"
```
connorshea: FYI, this is also missing `T::Set`, though it's not used very much in the stdlib/core sigs. The only example I can find is https://github.com/sorbet/sorbet/blob/2275752e51604acfb79b30a0a96debc996c089d9/rbi/stdlib/set.rbi#L13

soutaro:
Merging this to make the auto-generated signature the starting point to make collaboration.

PR #39, opened 2019-10-24
Fix BasicObject#instance_eval signature

hanachin:
```
irb(main):001:0> 42.instance_eval { p _1 }
42
=> 42
irb(main):002:-> 42.instance_eval(&->{})
        6: from /usr/local/var/rbenv/versions/master/bin/irb:23:in `<main>'
        5: from /usr/local/var/rbenv/versions/master/bin/irb:23:in `load'
        4: from /usr/local/var/rbenv/versions/master/lib/ruby/gems/2.7.0/gems/irb-1.1.0.pre.3/exe/irb:11:in `<top (required)>'
        3: from (irb):2
        2: from (irb):2:in `instance_eval'
        1: from (irb):2:in `block in irb_binding'
ArgumentError (wrong number of arguments (given 1, expected 0))
```

### DIFF
--- a/stdlib/builtin/basic_object.rbs
+++ b/stdlib/builtin/basic_object.rbs
@@ -91,7 +91,7 @@ class BasicObject
   def equal?: (untyped other) -> bool
 
   def instance_eval: (?String arg0, ?String filename, ?Integer lineno) -> untyped
-                   | [U] () { () -> U } -> U
+                   | [U] () { (self) -> U } -> U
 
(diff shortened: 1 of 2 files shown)


soutaro:
I will check what is happening...

soutaro:
Okay, it is because `BasicObject` does not provide some important methods like `method`, `class`, or `singleton_class`. I think skip writing tests for `BasicObject` would make sense.

hanachin:
#41 made BasicObject testable :clap:  so I added a test :v:

PR #41, opened 2019-10-24
Fix methods

soutaro:
#39 told me that `BasicObject` does not have several methods including `method`, `class`, or `singleton_class`. So, we should extract these methods beforehand.

The patch includes a fix to extract `is_a?` from `Kernel`, not `Object`. And it adds `PP` for future printf debugging.

### DIFF
--- a/lib/ruby/signature/test/hook.rb
+++ b/lib/ruby/signature/test/hook.rb
@@ -13,6 +13,7 @@ module Ruby
         CLASS = Kernel.instance_method(:class)
         SINGLETON_CLASS = Kernel.instance_method(:singleton_class)
         PP = Kernel.instance_method(:pp)
+        INSPECT = Kernel.instance_method(:inspect)
 
         module Errors
           ArgumentTypeError =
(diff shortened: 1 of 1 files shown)


hanachin:
Following test will raise `NoMethodError` when `RBS_TEST_LOGLEVEL=debug` at these lines

```ruby
class BasicObjectTest < StdlibTest
  target BasicObject
  using hook.refinement

  def test_instance_eval
    BasicObject.new.instance_eval {  }
  end
end
```
https://github.com/ruby/ruby-signature/blob/2a47654e84c0e06e67fe6f2c0b6964b498d167ec/lib/ruby/signature/test/hook.rb#L168
https://github.com/ruby/ruby-signature/blob/2a47654e84c0e06e67fe6f2c0b6964b498d167ec/lib/ruby/signature/test/hook.rb#L177
https://github.com/ruby/ruby-signature/blob/2a47654e84c0e06e67fe6f2c0b6964b498d167ec/lib/ruby/signature/test/hook.rb#L187
https://github.com/ruby/ruby-signature/blob/2a47654e84c0e06e67fe6f2c0b6964b498d167ec/lib/ruby/signature/test/hook.rb#L203

soutaro:
@hanachin No `#inspect` ... 😱 Thanks.

PR #42, opened 2019-10-25
Improve object inspect

hanachin:


lib/ruby/signature/test/hook.rb
```diff
@@ -320,6 +320,8 @@ def call(receiver, method, *args, &block)
         end
 
         def inspect_(obj)
+          obj.inspect
```
hanachin: Some classes override inspect method (Array, String, etc...) which is more readable than `Object#inspect`.

PR #43, opened 2019-10-26
Fix type check of empty block args

hanachin:
raise Test::Hook::Errors::BlockArgumentTypeError
when block args passed for `[U] () { () -> U } -> U`


soutaro:
Let me consider this carefully because ruby allows passing extra block arguments. This might be intended behavior, and the design is not trivial.

soutaro:
I got it. I was confusing and your proposal makes sense. So, block type checking should be as strict as method type checking.

soutaro:
Closing because #44 includes this change.

PR #44, opened 2019-10-27
Fix block typing

soutaro:
Related to #43.

* `yield`s should give all required arguments.
* The blocks may ignore the given arguments.


soutaro:
@hanachin Do you think this makes sense?

hanachin:
@soutaro That makes sense 👏

PR #52, opened 2019-11-02
Support empty tuple and single tuple

hanachin:


soutaro:
@hanachin The implementation looks good. 👍 

Could you give me an example to explain the reason why we need an *empty tuple*?

hanachin:
@soutaro For example, `NilClass#to_a` always return empty array.
https://docs.ruby-lang.org/en/2.6.0/NilClass.html#method-i-to_a

soutaro:
@hanachin Totally makes sense. Thanks!

PR #55, opened 2019-11-06
Update Binding signature and add some test

hanachin:


stdlib/builtin/binding.rbs
```diff
@@ -31,7 +31,7 @@
 # 
 # [Binding](Binding) objects have no class-specific
 # methods.
-class Binding < Object
+class Binding[T] < Object
```
soutaro: I think we need a variance annotation here `out T`.

soutaro:
I'm not very sure if it's a good idea to have a type of receiver in `Binding`, making it `untyped` might be sufficient. I don't mean it should be `untyped` though.

stdlib/builtin/binding.rbs
hanachin: I didn't know the variance, thank you for the comment.

hanachin:
> I'm not very sure if it's a good idea to have a type of receiver in Binding, making it untyped might be sufficient.

I'm not sure too, so I removed the commit to leave it as it is.
I added a commit to update the type of return value of Binding#receiver to untyped.

stdlib/builtin/binding.rbs
```diff
@@ -90,11 +90,17 @@ class Binding < Object
   # ```
   # 
   # if `obj` can be dumped in Ruby code.
-  def local_variable_set: (String | Symbol symbol, untyped obj) -> untyped
+  def local_variable_set: [U] (String | Symbol symbol, U obj) -> U
```
soutaro: 👍

PR #66, opened 2019-11-10
Add method signature of NilClass

kunitoo:
- `to_i`

### DIFF
--- a/stdlib/builtin/nil_class.rbs
+++ b/stdlib/builtin/nil_class.rbs
@@ -35,9 +35,9 @@ class NilClass < Object
   # Always returns zero.
   # 
   # ```ruby
-  # nil.to_i   #=> 0.0
+  # nil.to_i   #=> 0
   # ```
-  def to_i: () -> Integer
+  def to_i: () -> 0
 
   # Returns zero as a rational.
   def to_r: () -> Rational
--- a/stdlib/builtin/nil_class.rbs
+++ b/stdlib/builtin/nil_class.rbs
@@ -32,6 +32,13 @@ class NilClass < Object
   # ```
   def to_h: () -> ::Hash[untyped, untyped]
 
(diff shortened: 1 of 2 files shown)


stdlib/builtin/nil_class.rbs
```diff
@@ -32,6 +32,13 @@ class NilClass < Object
   # ```
   def to_h: () -> ::Hash[untyped, untyped]
 
+  # Always returns zero.
+  # 
+  # ```ruby
+  # nil.to_i   #=> 0.0
```
pocke: nil.to_i returns `0`, not `0.0`. Could you update this comment?

stdlib/builtin/nil_class.rbs
```diff
@@ -32,6 +32,13 @@ class NilClass < Object
...
+  # ```
+  def to_i: () -> Integer
```
hanachin: `nil.to_i` always returns 0, IMO literal type `0` is better than Integer

soutaro:
@kunitoo The comments above would make sense. Could you update?

stdlib/builtin/nil_class.rbs
kunitoo: Thank you for the review. Fixed it. https://github.com/ruby/ruby-signature/pull/66/commits/cdfe4e5709b3d58ec7c0905eec18cc55bf6c44ee

stdlib/builtin/nil_class.rbs
kunitoo: Thank you for the review. I didn't know there was a literal type.
Fixed https://github.com/ruby/ruby-signature/pull/66/commits/cdfe4e5709b3d58ec7c0905eec18cc55bf6c44ee

kunitoo:
@soutaro I updated to reflect comments.

PR #68, opened 2019-11-10
add Range#new test

YutaGoto:
- `Range#new` 
  - add `new` method with block

### DIFF
--- a/test/stdlib/Range_test.rb
+++ b/test/stdlib/Range_test.rb
@@ -1,0 +1,11 @@
+class RangeTest < StdlibTest
+  target Range
+  using hook.refinement
+
+  def test_new
+    Range.new(1, 10)
+    Range.new(11, 20, true)
+    Range.new('a', 'z', false)
+    Range.new(-1, nil)
+  end
+end


stdlib/builtin/range.rbs
```diff
@@ -90,7 +90,7 @@
 class Range[Elem] < Object
   include Enumerable[Elem, Range[Elem]]
 
-  def self.new: [U] (U from, U to, ?bool exclude_end) -> ::Range[U]
+  def self.new: [U] (U from, U to, ?bool exclude_end) ?{ () -> void } -> ::Range[U]
```
soutaro: It looks like `Range.new` doesn't receive blocks to me. Could you tell me how did you find it?

YutaGoto: Almost Rubys methods can receive no meaning blocks (e.g. `"1234".to_i {}`, `Range.new(1, 10){}`). I checked reference and Rubys C code. but I can't find..
I think it doesn't have to support this case.

soutaro: Yest, Ruby methods accepts any block if the method doesn't `yield` it. However, the mistake to passing unused block is so common, we decided not to allow passing unused blocks.

I'm sorry that I haven't explained the choice.

soutaro:
@YutaGoto could you squash the commits?

YutaGoto:
@soutaro I squashed these commits. please check it 🙇

PR #72, opened 2019-11-10
Add String#encode and #encode!

mtsmfm:


soutaro:
I found that we cannot write literal type for symbols ending with `?`. 😮

mtsmfm:
> I found that we cannot write literal type for symbols ending with `?`. 😮

True...

mtsmfm:
I fixed conflict

mtsmfm:
@soutaro Can you review?

soutaro:
#83 would have broken this? @mtsmfm

mtsmfm:
> #83 would have broken this? @mtsmfm

Ah, yes!
I've fixed

PR #74, opened 2019-11-10
Add kTRUE and kFALSE

mtsmfm:
Currently, true and false literals aren't supported despite it's written in our doc.

https://github.com/ruby/ruby-signature/blob/ba26445a4f18c26d140f201896d7d1ace7b3c717/doc/syntax.md#literal-type

I'd like to add kTRUE and kFALSE.


soutaro:
@mtsmfm Looks great!

Could update the `identifier_keywords` rule so that we can define `true` and `false` methods.

mtsmfm:
> Could update the `identifier_keywords` rule so that we can define `true` and `false` methods.

Thank you for pointing out.
Updated both the rule and the test.

PR #77, opened 2019-11-11
 Add test and update types for Kernel

pocke:
This pull request adds tests for all methods for `Kernel` module. And it updates some types because they are wrong.

I gave up writing tests for some methods, because they are difficult.
But `test_*` methods are available for all methods. I put "TODO" comment to the methods I gave up.

### DIFF
--- a/stdlib/builtin/kernel.rbs
+++ b/stdlib/builtin/kernel.rbs
@@ -18,7 +18,8 @@ module Kernel
   def caller_locations: (?Integer start_or_range, ?Integer length) -> ::Array[Thread::Backtrace::Location]?
                       | (?::Range[Integer] start_or_range) -> ::Array[Thread::Backtrace::Location]?
 
-  def catch: (?Object tag) { (Object arg0) -> untyped } -> untyped
+  def catch: [T] (T tag) { (T tag) -> untyped } -> untyped
+           | () { (Object tag) -> untyped } -> untyped
 
   # In a perfect world this should be:
   # 
@@ -53,6 +54,7 @@ module Kernel
   # try do "hello" end   #=> "hello"
   # ```
   def iterator?: () -> bool
+  alias block_given? iterator?
 
   # Returns the names of the current local variables.
   # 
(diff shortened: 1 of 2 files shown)


stdlib/builtin/kernel.rbs
```diff
@@ -124,61 +127,53 @@ module Kernel
...
   def tap: () { (untyped x) -> void } -> self
 
-  def to_enum: (?Symbol method, *untyped args) -> ::Enumerator[untyped, untyped]
-             | (?Symbol method, *untyped args) { () -> untyped } -> ::Enumerator[untyped, untyped]
-
   def to_s: () -> String
 
-  def trust: () -> self
-
-  def `undef`: (*untyped arg) -> void
```
pocke: Ruby does not have `undef` method.

stdlib/builtin/kernel.rbs
```diff
@@ -199,18 +194,20 @@ module Kernel
   # Array(nil)         #=> []
   # Array(1)           #=> [1]
   # ```
-  def Array: (NilClass x) -> untyped
-           | (untyped x) -> ::Array[untyped]
-
-  def BigDecimal: (Integer | Float | Rational | BigDecimal | String initial, ?Integer digits, ?exception: bool exception) -> BigDecimal
```
pocke: BigDecimal is a standard library. So it should be defined as an extension.

stdlib/builtin/kernel.rbs
```diff
@@ -333,13 +313,14 @@ module Kernel
   # `Exception` object or `nil`, can be specified via the `:cause`
   # argument.
   def fail: () -> bot
-          | (?String arg0) -> bot
-          | (?Class arg0, ?::Array[String] arg1) -> bot
-          | (?Class arg0, ?String arg1, ?::Array[String] arg2) -> bot
+          | (String arg0) -> bot
+          | (?Exception arg0, ?String arg1, ?::Array[String] arg2) -> bot
```
soutaro: I think the type of `fail` should be updated later, but not clear enough how it can be now.

PR #81, opened 2019-11-23
Add some hash signatures

natsuokw:
following signatures and tests of Hash are added

- `<`
- `<=`
- `>`
- `>=`
- `compact`
- `compact!`
- `filter`
- `filter!`
- `flatten`
- `replace`
- `select`
- `select!`
- `to_proc`


soutaro:
@natsuokw Thanks! 🎉

PR #84, opened 2019-11-26
Import RDoc annotations

soutaro:
Add a tool for importing/querying RDoc, `annotate-with-rdoc` and `query-rdoc`.

* You can use `bin/annotate-with-rdoc` to import RDoc and annotate RBS files.
* You can use `bin/query-rdoc` to query RDoc to be copied to RBS files.


soutaro:
Tested with `Object` and `BasicObject`.

PR #88, opened 2019-12-01
Add base64 signatures

hanachin:


lib/ruby/signature/environment_loader.rb
```diff
@@ -93,6 +93,7 @@ def load(env:)
 
         if stdlib_root
           signature_files.push(*each_signature(stdlib_root + "builtin"))
+          signature_files.push(*each_signature(stdlib_root + "base64"))
```
hanachin: `"pathname"` and `"set"` should also be added, but in a separate pull request.

soutaro: I think we need better abstraction for requiring standard libraries. Ok for now.

PR #91, opened 2019-12-02
Add SecureRandom signatures

ykpythemind:


soutaro:
@ykpythemind Thanks! 🎉

PR #94, opened 2019-12-03
Add Range several tests

YutaGoto:
add Range tests

- `inspect`
- `last`
- `max`
- `min`
- `size`
- `step`
- `to_s`
- `eql?`
- `member?`

and add `Range#size` signature to using Infinity Ranges case.


soutaro:
`size` is not an `Integer` ... 🤔

PR #97, opened 2019-12-07
Support gvar, ivar, cvar, :+ and :- symbol literal

mtsmfm:
Follow up https://github.com/ruby/ruby-signature/pull/83


soutaro:
Thanks! 🎉

PR #98, opened 2019-12-07
Revise test runner

soutaro:
* For better support loading library signatures in tests
* Common ruby test running pattern to run each test case `$ bundle exec ruby test/stdlib/String_test.rb`

### DIFF
--- a/lib/ruby/signature/environment_loader.rb
+++ b/lib/ruby/signature/environment_loader.rb
@@ -93,10 +93,6 @@ module Ruby
 
         if stdlib_root
           signature_files.push(*each_signature(stdlib_root + "builtin"))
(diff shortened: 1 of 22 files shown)


lib/ruby/signature/environment_loader.rb
```diff
@@ -93,10 +93,6 @@ def load(env:)
 
         if stdlib_root
           signature_files.push(*each_signature(stdlib_root + "builtin"))
-          signature_files.push(*each_signature(stdlib_root + "base64"))
```
soutaro: Essentially I want to delete these lines.
I believe that loading standard library (non-builtin) signatures explicitly is better.

PR #103, opened 2019-12-09
bin/test_runner.rb should require pathname

mame:


mame:
Currently the script works with `bundler exec bin/test_runner.rb` maybe because bundler itself does `require "pathname"`.  But ruby-signature must not depend upon it.

PR #109, opened 2019-12-12
Add LoadError#path signature

hanachin:


stdlib/builtin/errors.rbs
```diff
@@ -142,6 +142,7 @@ end
 # 
 #     LoadError: no such file to load -- this/file/does/not/exist
 class LoadError < ScriptError
+  def path: () -> String
```
soutaro: rurima says it can be `nil`. https://docs.ruby-lang.org/ja/latest/class/LoadError.html Do you have any opinion?

hanachin: I confirm that the `LoadError#path` returns `nil` for `eval('require_relative ""')`.
I'm updating signature and test.

PR #113, opened 2019-12-13
Add rubocop-rubycw

hanachin:
to ensure no warning, such as:
https://github.com/ruby/ruby-signature/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+warning

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -16,7 +16,7 @@ task :stdlib_test do
 end
 
 task :rubocop do
-  sh "rubocop -P"
+  sh "rubocop --parallel"
 end
 
 rule ".rb" => ".y" do |t|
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -9,12 +9,16 @@ Rake::TestTask.new(:test) do |t|
...
 
-task :default => [:test, :stdlib_test]
+task :default => [:test, :stdlib_test, :rubocop]
 
 task :stdlib_test do
   sh "ruby bin/test_runner.rb"
 end
 
+task :rubocop do
+  sh "rubocop -P"
```
pocke: Personally I prefer the long option, `--parallel`,  to `-P` in a committed file.
Because it is explicit.

I also like the short option on a console, because it needs fewer types.
But typing count is not important in a committed file. Readable code is more important.

hanachin: Totally! I updated the option.

PR #126, opened 2019-12-21
Fix FrozenError signature

soutaro:
I'm not sure why #116 passes CI... Anyway, fixing the issue.


hanachin:
The `FrozenError#initialize` was changed 2 days ago. #116 was created 4 days ago. Thus the signature was valid.
https://github.com/ruby/ruby/commit/435a4ca2a38f3be1f5d2db0f71487a52c8285e9c

PR #139, opened 2019-12-27
Fix Writer output with an empty tuple, and optional symbol literal

pocke:
This pull request fixes two bugs of the Writer. Empty tuple and optional symbol literal.

# empty tuple

Currently the Writer output an empty tuple with two spaces. This change reduces the spaces to one.

```
# input
[ ]

# expected
[ ]

# actual
[  ]
```

# optional symbol literal

The Writer generates wrong code with optional symbol literal.

```
# input
:foo ?

# expected
:foo ?

# actual
:foo?
```

Because symbol literal can include `?` at the end, like Ruby. The actual case has a different meaning with the input.

I guess wrapping with braces is also ok. like `(:foo)?`. But we already have the expected style code.
So I chose the style.

https://github.com/ruby/ruby-signature/blob/6dd573c0326450ebdc3308229027ab303657d601/stdlib/builtin/string.rbs#L716-L717

### DIFF
--- a/lib/ruby/signature/types.rb
+++ b/lib/ruby/signature/types.rb
@@ -330,7 +330,11 @@ module Ruby
         end
 
         def to_s(level = 0)
-          "[ #{types.join(", ")} ]"
+          if types.empty?
+            "[ ]"
+          else
+            "[ #{types.join(", ")} ]"
+          end
         end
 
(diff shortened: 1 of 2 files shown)


soutaro:
Thanks. 🙇

PR #146, opened 2019-12-29
[Writer] Preserve an empty line between declarations and members

pocke:
By this change, the writer will preserve an empty line between declarations and members.

I also added a smoke test for `Ruby::Signature::Writer`.
It only checks the writer generates the same meaning signature.

# Format examples

no changes

```
class A
  def foo: () -> void
  def bar: () -> void
end
class B
end
```

also no changes

```
class A
  def foo: () -> void

  def bar: () -> void
end

class B
end
```

But it removes too many empty lines. the following code will be formatted to the same as the previous example.

```
class A
  def foo: () -> void


  def bar: () -> void
end


class B
end
```

### DIFF
--- a/test/ruby/signature/writer_test.rb
+++ b/test/ruby/signature/writer_test.rb
@@ -200,6 +200,7 @@ end
     SIG
 
     assert_equal expected, format(src)
+
   end
 
   def test_smoke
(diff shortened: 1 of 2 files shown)


test/ruby/signature/writer_test.rb
```diff
@@ -136,4 +140,79 @@ class Foo[out A, unchecked B, in C] < Bar[A, C, B]
...
+module Bar
+  def foo: () -> void
+end
+    SIG
+
+    assert_equal expected, format(src)
+
+  end
+
+  def test_smoke
```
soutaro: 👍

soutaro:
@pocke Thanks!

PR #149, opened 2019-12-29
Update TrueClass signatures

hanachin:
done
https://docs.ruby-lang.org/en/2.6.0/TrueClass.html

### DIFF
--- a/stdlib/builtin/true_class.rbs
+++ b/stdlib/builtin/true_class.rbs
@@ -16,6 +16,9 @@ class TrueClass
   # otherwise.
   def ^: (bool obj) -> bool
 
+  # The string representation of `true` is "true".
+  def to_s: () -> "true"
+
   # Or—Returns `true` . As *obj* is an argument to a method call, it is
   # always evaluated; there is no short-circuit evaluation in this case.
   def |: (bool obj) -> bool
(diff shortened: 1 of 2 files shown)


stdlib/builtin/true_class.rbs
```diff
@@ -1,28 +1,46 @@
...
+  # 
+  def ^: (nil) -> true
+       | (false) -> true
+       | (untyped obj) -> bool
+
+  alias inspect to_s
+
+  # The string representation of `true` is "true".
+  # 
+  def to_s: () -> "true"
```
soutaro: I'm not sure what the type of `#to_s` should be. It may be `String` or `"true"`...
Leaving it as proposed.

stdlib/builtin/true_class.rbs
```diff
@@ -1,28 +1,46 @@
...
+  #     true || puts("logical or")
   # 
   # *produces:*
   # 
-  # ```
-  # or
-  # ```
+  #     or
+  # 
   def |: (bool obj) -> bool
```
pocke: I guess the returned value type can be `true`. And we can also modify the `!` method definition.


I think `true` is more useful than `bool` in this case. Because it is more restricted.


By the way, FalseClass's definition already has the restricted type in your pull request.
https://github.com/ruby/ruby-signature/pull/150/files#diff-7ae9c73db685d683bf00b17333e34667R8

soutaro: This is intentional. I'm not very sure if this makes more sense, but would like to keep this `bool` for now.

PR #150, opened 2019-12-30
Update FalseClass signatures

hanachin:
done
https://docs.ruby-lang.org/en/2.6.0/FalseClass.html

### DIFF
--- a/stdlib/builtin/false_class.rbs
+++ b/stdlib/builtin/false_class.rbs
@@ -9,7 +9,7 @@ class FalseClass
 
   # And—Returns `false` . *obj* is always evaluated as it is the argument to
   # a method call—there is no short-circuit evaluation in this case.
-  def &: (untyped obj) -> FalseClass
+  def &: (untyped obj) -> false
 
   # Exclusive Or—If *obj* is `nil` or `false`, returns `false` ; otherwise,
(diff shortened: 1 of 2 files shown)


stdlib/builtin/false_class.rbs
```diff
@@ -3,16 +3,30 @@
...
+
+  def ===: (false) -> true
+         | (untyped obj) -> false
 
   # Exclusive Or—If *obj* is `nil` or `false`, returns `false` ; otherwise,
   # returns `true` .
-  def ^: (untyped obj) -> bool
+  def ^: (nil) -> false
+       | (false) -> false
+       | (untyped obj) -> true
```
pocke: I'm wondering if the last returned value should be `true`.

Definitely it works, if the type checker checks the types from top to bottom.

```
# Pseudocode

if receiver =~ nil
  return_type = false
elsif receiver =~ false
  return_type = false
elsif receiver =~ untyped
  reutrn_type = true
else
  # type error
end
```


However type checker will be confused if it does not check the receiver type with the defined order.
If a type checker checks the `untyped` clause at first, it returns an incorrect type.


So, if RBS should be evaluated with the defined order, the current type definition is OK.
Otherwise we should replace the "true" with "bool".

PR #153, opened 2019-12-30
Fix `rbs prototype rb` command

pocke:
I'm trying to use `rbs prototype rb`command for a real Rails application, but I got errors.

Currently this pull request fixes two problems. The command successfully finished with the changes.
But it still has a problem. The generated signature has syntax error(s).

So I'll mark this pull request reviewable after fixing the remaining bugs.

# Problems

## arguemnt-less FCALL

FCALL node can be argument-less, but the command does not treat that case.

```ruby
# They're FCALLs, but without arguments.
foo() # (FCALL@1:0-1:5 :foo nil)
foo{} # (ITER@1:0-1:5 (FCALL@1:0-1:3 :foo nil) (SCOPE@1:3-1:5 tbl: [] args: nil body: (BEGIN@1:4-1:4 nil)))
```

I met the bug with `included` in the Rails application.

```ruby
class C
  included do
    def foo
    end
  end
end
```

## argument-less yield

`yield` also can be argument-less, but it does not treat.

```ruby
def foo
  yield
end
```

### DIFF
--- a/lib/ruby/signature/prototype/rb.rb
+++ b/lib/ruby/signature/prototype/rb.rb
@@ -337,7 +337,7 @@ module Ruby
               )
 
               yields.each do |yield_node|
-                array_content = yield_node.children[0].children.compact
+                array_content = yield_node.children[0]&.children&.compact || []
 
                 positionals, keywords = if keyword_hash?(array_content.last)
                                           [array_content.take(array_content.size - 1), array_content.last]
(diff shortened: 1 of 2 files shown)


pocke:
https://github.com/ruby/ruby-signature/issues/154

We need to modify the syntax of RBS to fix the syntax error. So I think I should work for fixing the syntax error on another pull request.
So I'll mark this pull request as ready for review.

soutaro:
@pocke Is this ready for merge? (I 👍 for the change.)

pocke:
> Is this ready for merge? (I +1 for the change.)

Yes. Thanks for reviewing!

PR #155, opened 2019-12-31
Accept keyword as var name opt

### LINKED ISSUE #154: Keyword as argument name is syntax error
# Problem

Currently, RBS does not allow a keyword as an argument name.

```ruby
class C
  # "type" is a keyword in RBS, and a keyword is not allowed at here (var_name_opt).
  # So it is syntax error.
  def foo: (untyped type) -> void

  # It is also syntax error.
  def foo: (type: untyped type) -> void

  # It is not a syntax error.
  def foo: (type: untyped) -> void
end
```

I found this problem while #153 .
We have the following Ruby code, and `rbs prototype rb` command generates a code with synax error from it.

```ruby
class C
  def foo(type:)
  end
end
```

# Solution

I have several ideas to fix it.

* Allow keyword as an argument name.
  * ```diff
    diff --git a/lib/ruby/signature/parser.y b/lib/ruby/signature/parser.y
    index 2c8475e..654b232 100644
    --- a/lib/ruby/signature/parser.y
    +++ b/lib/ruby/signature/parser.y
    @@ -941,7 +941,7 @@ rule
           }
     
       var_name_opt:
    -    | tLIDENT | tINTERFACEIDENT
    +    | tLIDENT | tINTERFACEIDENT | identifier_keywords
     
       qualified_name:
           namespace simple_name {
    ```
  * It works well and does not increase conflicts.
  * But I'm wondering if it conflicts in the future.
* Allow escaping for argument name, like method name.
  * ``def foo: (untyped `type`) -> void``
  * I like this idea.
* Fix `rbs prototype rb` command to avoid generating syntax error code.
  * e.g. replacing `type` with `_type`
  * It works for `rbs prototype rb`, but I guess we'd like to use `type` as argument name.
  * So I think it is not a good idea. 



What do you think about these solution ideas?

pocke:
Fix #154

This pull request allows you to write a keyword as an argument name with escaping by backticks.

```rbs
class C
  # ok
  def foo: (untyped `type`, void: untyped `void`) -> untyped

  # syntax error
  def bar: (untyped type, void: untyped void) -> untyped

  # syntax error
  def baz: (untyped `**`) -> untyped
end
```

This pull request also fixes the Writer to generate code with the escaping.

---

By the way, it also includes a bug fix of Writer for method name escaping.
https://github.com/ruby/ruby-signature/pull/155/files#diff-5e2ea3c3d2ea351a23bbb3cddd984c7dR206

```rbs
class C
  # Currently Writer escape `timeout` with backticks, because it ends with "out" and "out" is a keyword.
  def `timeout`: () -> Integer

  # But the escaping is unnecessary!
  def timeout: () -> Integer
end
```

### DIFF
--- a/lib/ruby/signature/writer.rb
+++ b/lib/ruby/signature/writer.rb
@@ -203,7 +203,7 @@ module Ruby
       def method_name(name)
         s = name.to_s
 
-        if s =~ Parser::KEYWORDS_RE
+        if /\A#{Parser::KEYWORDS_RE}\z/.match?(s)
           "`#{s}`"
         else
           s
--- a/lib/ruby/signature/parser.y
+++ b/lib/ruby/signature/parser.y
@@ -7,7 +7,7 @@ class Ruby::Signature::Parser
         kVOID kNIL kTRUE kFALSE kANY kUNTYPED kTOP kBOT kSELF kSELFQ kINSTANCE kCLASS kBOOL kSINGLETON kTYPE kDEF kMODULE kSUPER
         kPRIVATE kPUBLIC kALIAS
         kCOLON kCOLON2 kCOMMA kBAR kAMP kHAT kARROW kQUESTION kEXCLAMATION kSTAR kSTAR2 kFATARROW kEQ kDOT kLT
-        kINTERFACE kEND kINCLUDE kEXTEND kATTRREADER kATTRWRITER kATTRACCESSOR tOPERATOR tQUOTEDMETHOD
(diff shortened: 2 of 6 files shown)


lib/ruby/signature/writer.rb
```diff
@@ -203,7 +203,7 @@ def write_member(member)
       def method_name(name)
         s = name.to_s
 
-        if s =~ Parser::KEYWORDS_RE
+        if /\A#{Parser::KEYWORDS_RE}\z/.match?(s)
```
soutaro: 💡 🙏

PR #157, opened 2019-12-31
Support `class << self` syntax in `rbs prototype rb`

pocke:
Currently `rbs prototype rb` command ignores `class <<self` syntax.

```ruby
# test.rb
class C
  class << self
    def m() end
  end
  def m() end
end
```

```
$ ruby exe/rbs prototype rb test.rb
class C
  def m: () -> untyped

  def m: () -> untyped
end
```

This pull request will add the support of `class << self` syntax.

The generated rbs will be the following.

```bash
class C
  def self.m: () -> untyped

  def m: () -> untyped
end
```

### DIFF
--- a/lib/ruby/signature/prototype/rb.rb
+++ b/lib/ruby/signature/prototype/rb.rb
@@ -52,14 +52,14 @@ module Ruby
             end
           end
 
-          process RubyVM::AbstractSyntaxTree.parse(string), namespace: Namespace.empty, current_module: nil, comments: comments
+          process RubyVM::AbstractSyntaxTree.parse(string), namespace: Namespace.empty, current_module: nil, comments: comments, singleton: false
         end
 
         def nested_name(name)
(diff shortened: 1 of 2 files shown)


lib/ruby/signature/prototype/rb.rb
```diff
@@ -94,13 +94,22 @@ def process(node, namespace:, current_module:, comments:)
             source_decls.push mod
 
             each_node module_body do |child|
-              process child, namespace: mod.name.to_namespace, current_module: mod, comments: comments
+              process child, namespace: mod.name.to_namespace, current_module: mod, comments: comments, singleton: false
+            end
+          when :SCLASS
+            this = node.children[0]
+            if this.type != :SELF
+              Signature.logger.warn "`class <<` syntax with not-self may be compiled to incorrect code: #{this}"
```
pocke: The generator will generate incorrect rbs for the following ruby code.


```ruby
class << Foobar
  def foo() end
end
```


I think it is a super rare case, so it just warns about it.

PR #164, opened 2020-01-02
Vendoring RBSs

soutaro:
This PR is to support vendoring signatures into projects.

The vendoring allows:

1. Having the copy of standard library signatures in projects
2. Having the copy of gem signatures in projects

For 2., the `EnvironmentLoader` now accepts `gem_vendor_path`. `EnvironmentLoader` has three attributes `stdlib_root`, `gem_vendor_path`, and `no_builtin?`. We can use the three attributes to control:

1. How standard library signatures can be found?
2. Whether gem signatures are vendored or not?
3. If we load the builtin signatures?

We also add a CLI command `rbs vendor`, which can be used to vendor the signatures.

### DIFF
--- a/lib/ruby/signature/vendorer.rb
+++ b/lib/ruby/signature/vendorer.rb
@@ -1,0 +1,49 @@
+module Ruby
+  module Signature
+    class Vendorer
+      attr_reader :vendor_dir
+
+      def initialize(vendor_dir:)
+        @vendor_dir = vendor_dir
+      end
+
+      def ensure_dir
+        unless vendor_dir.directory?
+          vendor_dir.mkpath
+        end
+
+        yield
+      end
+
+      def clean!
+        ensure_dir do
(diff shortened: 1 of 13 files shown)


lib/ruby/signature/vendorer.rb
```diff
@@ -0,0 +1,49 @@
...
+      def gem!(name, version)
+        ensure_dir do
+          sig_path = EnvironmentLoader.gem_sig_path(name, version)
+          Signature.logger.debug "Checking gem signature path: name=#{name}, version=#{version}, path=#{sig_path}"
+
+          if sig_path&.directory?
+            gems_dir = vendor_dir + "gems"
+            gems_dir.mkpath unless gems_dir.directory?
+
+            gem_dir = gems_dir + name
```
soutaro: Gem signature vendoring does not support versions. One project cannot have different versions of a gem vendored.

PR #165, opened 2020-01-02
Two fixes

soutaro:
T/O

### DIFF
--- a/lib/ruby/signature/writer.rb
+++ b/lib/ruby/signature/writer.rb
@@ -250,15 +250,16 @@ module Ruby
       def preserve_empty_line(prev, decl)
         return unless prev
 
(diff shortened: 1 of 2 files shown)


lib/ruby/signature/writer.rb
```diff
@@ -250,15 +250,16 @@ def attribute(kind, attr)
       def preserve_empty_line(prev, decl)
         return unless prev
 
+        decl = decl.comment if decl.respond_to?(:comment) && decl.comment
```
soutaro: `decl.comment.location` may be `nil`.

@pocke Any comment?

pocke: This change is reasonable. Thanks :+1:

PR #166, opened 2020-01-03
Rename test_polar methods in Complex_test to avoid conflicting

pocke:


soutaro:
@pocke Could you rename the first `test_polar` method? The first one is to test *polar singleton method* and the second one is to test *polar instance method*. I think keeping both of them makes sense, and `test_singleton_polar`, `test_instance_polar`, or something else would be a good name.

pocke:
Make sense. I renamed them and force-pushed the change.
 Thanks!

PR #167, opened 2020-01-03
Fix failing stdlib_test with Ruby 3 keyword argument

pocke:
Today removing Ruby 2 keyword argument was committed to the master branch of Ruby.
https://github.com/ruby/ruby/commit/beae6cbf0fd8b6619e5212552de98022d4c4d4d4

So the stdlib test fails with the master branch.

This pull request fixes the failing.

The original idea is from @mame. Thanks!

### DIFF
--- a/test/ruby/signature/test_test.rb
+++ b/test/ruby/signature/test_test.rb
@@ -3,6 +3,8 @@ require "test_helper"
 require "ruby/signature/test"
 require "logger"
 
+return unless Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7.0')
+
 class Ruby::Signature::TestTest < Minitest::Test
   include TestHelper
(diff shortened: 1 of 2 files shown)


test/ruby/signature/test_test.rb
```diff
@@ -3,6 +3,8 @@
 require "ruby/signature/test"
 require "logger"
 
+return unless Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7.0')
```
pocke: It skips test for `Ruby::Signature::Test:Hook` if the Ruby runtime version is less than 2.7.

We also use ruby2_keywords gem to apply the polyfill instead of the skipping, but it needs https://github.com/ruby/ruby2_keywords/pull/3.

PR #170, opened 2020-01-03
Remove or rename duplicated test cases

pocke:
After #166, I found other duplicated test cases.
So this pull request renames or removes them.

This pull request contains #166.

By the way, this pull request also enables Lint/DuplicatedMethods cop for test files.
I think it is helpful to avoid test bugs. `ruby -w` also warns about it, but I guess it will be overlooked because the test prints many lines by default.

### DIFF
--- a/test/stdlib/Integer_test.rb
+++ b/test/stdlib/Integer_test.rb
@@ -137,11 +137,6 @@ class IntegerTest < StdlibTest
     3.abs2
   end
 
-  def test_digits
-    1.digits
-    1.digits(2)
-  end
-
   def test_allbits?
     1.allbits?(1)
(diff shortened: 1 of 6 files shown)


test/stdlib/Integer_test.rb
```diff
@@ -137,11 +137,6 @@ def test_abs_abs2
     3.abs2
   end
 
-  def test_digits
-    1.digits
-    1.digits(2)
-  end
```
pocke: https://github.com/ruby/ruby-signature/blob/e11bce363a1cc67125cfdbd4491ce1512b9aacc0/test/stdlib/Integer_test.rb#L189-L194
Another definition is here.

PR #174, opened 2020-01-06
[WIP] Add a signature for tempfile.rb

mame:
How can we handle `class Tempfile < Delegate(File)`?  Is `Tempfile < File` good enough?

And we need to complete `File.open` signature because `Tempfile.open` and so on delegates all arguments to `File.open`.

### DIFF
--- a/stdlib/tempfile/tempfile.rbs
+++ b/stdlib/tempfile/tempfile.rbs
@@ -1,0 +1,15 @@
+class Tempfile
+  # < Delegate(File) ???
+  def initialize : (String?, ?String?, mode: Integer, max_try: Integer?, **) # the same as File.open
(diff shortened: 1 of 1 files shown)


ybiquitous:
Hi, the RBS of `Tempfile` has been added, but this PR is still in progress?

https://github.com/ruby/rbs/blob/518440031d4a79c2b8c52ba8365f3295ab625e88/stdlib/tempfile/0/tempfile.rbs

pocke:
I'll close this issue because it has been added by another PR as @ybiquitous said.

By the way, the tempfile.rbs declares `class Tempfile < File`. I have no strong opinion of the superclass, I guess it's enough for now.

PR #186, opened 2020-01-17
Update String signatures

hanachin:


soutaro:
@hanachin I pushed my fixes. Review them please. 🙏

test/stdlib/String_test.rb
```diff
@@ -81,6 +81,8 @@ def test_concat_op
                      a, :<<, ToStr.new("world")
     assert_send_type "(Integer) -> String",
                      a, :<<, 33
+    refute_send_type "(ToInt) -> String",
```
hanachin: I didn't know that assertion, thanks!

PR #201, opened 2020-02-13
Update regexp signatures

natsuokawai:
Regexp signatues and tests.


stdlib/builtin/regexp.rbs
```diff
@@ -1,123 +1,359 @@
-class Regexp < Object
+class Regexp
+  # Alias for Regexp.new
+  #
+  def self.compile: (String arg0, ?untyped options, ?String kcode) -> Regexp
```
soutaro: I think it's better to have good argument names instead of `arg0` or something like that. I usually find the _good names_ from argslist.

stdlib/builtin/regexp.rbs
```diff
@@ -1,123 +1,359 @@
...
+  # Regexp::EXTENDED, Regexp::IGNORECASE, and Regexp::MULTILINE, *or*-ed together.
+  #  Otherwise, if `options` is not `nil` or `false`, the regexp will be case
+  # insensitive.
+  #
+  #     r1 = Regexp.new('^a-z+:\\s+\w+') #=> /^a-z+:\s+\w+/
+  #     r2 = Regexp.new('cat', true)     #=> /cat/i
+  #     r3 = Regexp.new(r2)              #=> /cat/i
+  #     r4 = Regexp.new('dog', Regexp::EXTENDED | Regexp::IGNORECASE) #=> /dog/ix
+  #
+  incompatible def initialize: (String arg0, ?untyped options, ?String kcode) -> Object
```
soutaro: Could you delete the `incompatible` and move the `initialize` to the top of instance methods. And we can delete the `private` because ruby-signature makes `initialize` private automatically as Ruby does.

stdlib/builtin/regexp.rbs
```diff
@@ -1,123 +1,359 @@
...
+                             | (Regexp arg0) -> void
 
-  def eql?: (untyped other) -> bool
+  def initialize_copy: (self object) -> self
 end
 
 Regexp::EXTENDED: Integer
```
soutaro: It looks like RDoc has entries for these constants. Might be a bug of `bin/annotate-with-rdoc`...

Could you add RDoc comments for these? The following is the comment.

```
# see Regexp.options and Regexp.new
# 
```

(It may be essentially useless, but I want to have RDoc content in RBS too.)

soutaro:
@natsuokawai Thank you so much for opening a PR. 👍 I have some requests to fix for minor issues.

stdlib/builtin/regexp.rbs
natsuokawai: Some parts of the document including `# see Regexp.options and Regexp.new` were lost because I applied `rbs prototype runtime --merge` after `bin/annotate-with-rdoc`...🙏
Then I appilied `bin/annotate-with-rdoc` again and fix them.

natsuokawai:
@soutaro 
Thank you for your review. I finished fixing them.

soutaro:
@natsuokawai Thanks!

PR #203, opened 2020-02-16
Fix warning: method redefined; discarding old *

hanachin:


soutaro:
It seems like the error message `NameError: method `fib' not defined in` is confusing. So, we may improve the implementation later...

PR #207, opened 2020-02-20
Add `rbs parse` subcommand

pocke:
This pull request will adds `rbs check` sub-command to check syntax error from .rbs files.

For example:

```bash
$ cat test.rbs
class C
  def foo: () ->
end

$ rbs check test.rbs
test.rbs:3:0: parse error on value: (kEND)
```

### DIFF
--- a/lib/ruby/signature/cli.rb
+++ b/lib/ruby/signature/cli.rb
@@ -37,7 +37,7 @@ module Ruby
         @stderr = stderr
       end
 
-      COMMANDS = [:ast, :list, :ancestors, :methods, :method, :validate, :constant, :paths, :prototype, :vendor, :version]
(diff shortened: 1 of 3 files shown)


soutaro:
The feature looks good. I'm wondering if there is a better name than `rbs check`. 

* `rbs parse`
* `rbs -C validate`
* `rbs syntax`
* ... any other?

I don't think one of the above is better...

pocke:
Hmm, I understand your concern. 
I have the following ideas, but I'm not sure they are better.

* `rbs syntax-check`
* `rbs syntax_check`
* `rbs validate`

soutaro:
I'd like to propose `rbs parse`. It parses the input and prints syntax errors if detected.

pocke:
Sure! I renamed the command to `rbs parse`. 38e36922d93808c80c416f5879d1795a7a356a2b

soutaro:
@pocke Thanks! 🎉

PR #209, opened 2020-02-21
Make `rbs prototype rb` to be aware of returned type

pocke:
`rbs prototype rb` command generates prototypes of method type definitions, but the returned types are always `untyped`.
It can decide the returned type easily if the method only contains a literal. Like `def foo() [1,2,3] end` or `def foo() {foo: 1} end`.

And I think it is useful in some cases.
I tried using this patch in our Rails application, and I found many meaningful changes.
Especially I think typing for a method that returns a Hash is useful. 
I also tried using it RuboCop's code. The result is here: https://gist.github.com/pocke/6fa639fc7549c90a852da769cbcd65b1

Note that the patch also includes Record type serializer bug fixes.

### DIFF
--- a/lib/ruby/signature/prototype/rb.rb
+++ b/lib/ruby/signature/prototype/rb.rb
@@ -339,12 +339,12 @@ module Ruby
             if lit.match?(/\A[ -~]+\z/)
               Types::Literal.new(literal: lit, location: nil)
             else
-              Types::ClassInstance.new(name: '::String', args: [], location: nil)
+              BuiltinNames::String.instance_type
             end
           when :DSTR, :XSTR
-            Types::ClassInstance.new(name: '::String', args: [], location: nil)
+            BuiltinNames::String.instance_type
           when :DSYM
-            Types::ClassInstance.new(name: '::Symbol', args: [], location: nil)
+            BuiltinNames::Symbol.instance_type
           when :DREGX
             Types::ClassInstance.new(name: '::Regexp', args: [], location: nil)
           when :TRUE
@@ -361,7 +361,7 @@ module Ruby
               if lit.match?(/\A[ -~]+\z/)
                 Types::Literal.new(literal: lit, location: nil)
               else
-                Types::ClassInstance.new(name: "::#{name}", args: [], location: nil)
+                BuiltinNames::Symbol.instance_type
               end
             when Integer
               Types::Literal.new(literal: lit, location: nil)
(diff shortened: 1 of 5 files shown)


lib/ruby/signature/prototype/rb.rb
```diff
@@ -322,13 +322,112 @@ def function_type_from_body(node)
...
+          return Types::Bases::Nil.new(location: nil) unless body
+
+          literal_to_type(body)
+        end
+
+        def literal_to_type(node)
+          case node.type
+          when :STR
+            lit = node.children[0]
+            if lit.match?(/\A[ -~]+\z/)
```
pocke: `[ -~]` means all displayable characters in ASCII.

lib/ruby/signature/prototype/rb.rb
```diff
@@ -322,13 +322,112 @@ def function_type_from_body(node)
...
+              Types::Literal.new(literal: lit, location: nil)
+            else
+              Types::ClassInstance.new(name: '::String', args: [], location: nil)
```
soutaro: How about using `Ruby::Signature::BuiltinNames::String.instance_type` ?

pocke: Thanks for the suggestion! I didn't know that.

I replaced them with BuiltinNames. 773015a4cc99419994b29b67d44a52ef57c90c4f

pocke:
The test was failed by merging #208, so I rebased this pull request on the master branch, and fixed the test in aa5db07.

Failed test: https://github.com/ruby/ruby-signature/pull/209/checks?check_run_id=463317343

lib/ruby/signature/prototype/rb.rb
```diff
@@ -325,13 +325,112 @@ def function_type_from_body(node)
...
+              Types::Literal.new(literal: lit, location: nil)
+            else
+              BuiltinNames::String.instance_type
+            end
+          when :DSTR, :XSTR
+            BuiltinNames::String.instance_type
+          when :DSYM
+            BuiltinNames::Symbol.instance_type
+          when :DREGX
+            Types::ClassInstance.new(name: '::Regexp', args: [], location: nil)
```
soutaro: Could you make `name:` be an instance of `TypeName`?

lib/ruby/signature/prototype/rb.rb
```diff
@@ -325,13 +325,112 @@ def function_type_from_body(node)
...
+          return untyped if types.include?(untyped)
+
+          Types::Union.new(types: types.uniq, location: nil)
+        end
+
+        def range_element_type(types)
+          types = types.reject { |t| t == untyped }
+          return untyped if types.empty?
+
+          type_names = types.map { |t| t.is_a?(Types::ClassInstance) ? t.name : "::#{t.literal.class.name}" }.uniq
```
soutaro: `type_names` should not be `String` too.

pocke:
@soutaro Thanks for your review!
I replaced the String type name with TypeName. c3fedf7cc7b7f106de1c2c8af28c004dcf497b93

PR #210, opened 2020-02-22
Update MatchData signatures

natsuokawai:
MatchData signatures and the tests.


soutaro:
@natsuokawai Thanks! 🎉

PR #212, opened 2020-02-23
Update Pathname type

pocke:
This pull request updates Pathname standard library types.
I think we can mark it done by this pull request.

### DIFF
--- a/stdlib/pathname/pathname.rbs
+++ b/stdlib/pathname/pathname.rbs
@@ -1,26 +1,238 @@
 class Pathname
-  def self.glob: (Pathname, ?Integer) -> Array[Pathname]
-  def +: (Pathname) -> Pathname
-       | (String) -> Pathname
+  def self.getwd: () -> untyped
 
-  def cleanpath: () -> self
-  def directory?: () -> bool
-  def file?: () -> bool
+  def self.glob: (*untyped) -> untyped
 
-  def join: (String) -> self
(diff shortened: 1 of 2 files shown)


pocke:
@soutaro I think this pull request is ready to review. Could you review it?


But I have two questions.

* Should I update types for the private methods of Pathname?
  I just left them, but I guess I need to remove them or add types to them. Which is the best way?
* I haven't written a test for `Pathname()` function because I had no idea for the test.
  I tried the following code, but actually it tested nothing because I got no error even if I removed the `Pathname()` type definition.
  ```ruby
  class PathnameKernelExtensionTest < StdlibTest
    target Kernel
    using hook.refinement

    def test_Pathname
      Pathname('foo')
      Pathname(ToStr.new('foo'))
      Pathname(Pathname('foo'))
    end
  end
  ```
  So I'd like to know how to write a test for this case.

stdlib/pathname/pathname.rbs
```diff
@@ -1,21 +1,1093 @@
...
+  #
+  # See File.binread.
+  #
+  def binread: (?Integer length, ?Integer offset) -> String
+
+  # Writes `contents` to the file, opening it in binary mode.
+  #
+  # See File.binwrite.
+  #
+  def binwrite: (String, ?Integer offset,
```
soutaro: 👏 😭

PR #218, opened 2020-02-25
Handle optional blocks in the RBI parser

### LINKED ISSUE #214: Optional blocks / nilable procs?
In a lot of Ruby code, you can use a method either with or without a block. For example, with `create_table` in Rails' ActiveRecord:

```ruby
# With block
create_table(:suppliers) do |t|
  t.string :name
end

# Without block
create_table(:suppliers)
```

In Sorbet, the signature for this methods looks like this (simplified to remove a bunch of unnecessary stuff):

```ruby
sig do
  params(
    table_name: T.any(String, Symbol),
    blk: T.nilable(T.proc.params(t: ActiveRecord::ConnectionAdapters::TableDefinition).void)
  ).void
end
def create_table(table_name, &blk); end
```

But when I run `bundle exec rbs prototype rbi foo.rbi` on a file with a nilable proc, I get this error:

```
Unexpected block type: ^(ActiveRecord::ConnectionAdapters::TableDefinition t) -> void?
(SCOPE@9:2-9:41
 tbl: [:table_name, :blk]
 args:
   (ARGS@9:19-9:35
    pre_num: 1
    pre_init: nil
    opt: nil
    first_post: nil
    post_num: 0
    post_init: nil
    rest: nil
    kw: nil
    kwrest: nil
    block: :blk)
 body: nil)
```

So, how are we supposed to represent optional blocks in rbs files?

connorshea:
Fixes #214.

Optional blocks (`T.nilable(T.proc.void)`) weren't being handled correctly by the RBI parser, and would output an error and fallback to an untyped block rather than returning the correct RBS syntax. With this change it's now handled correctly.

Input:

```ruby
# typed: true
class Foo
  sig { params(blk: T.nilable(T.proc.void)).returns(T.untyped) }
  def self.bar(&blk); end
end
```

Output:

```
# typed: true
class Foo
  def self.bar: () ?{ () -> void } -> untyped
end
```

### DIFF
--- a/lib/ruby/signature/prototype/rbi.rb
+++ b/lib/ruby/signature/prototype/rbi.rb
@@ -394,6 +394,9 @@ module Ruby
                   required: true,
                   type: Types::Function.empty(Types::Bases::Any.new(location: nil))
                 )
+              # Handle an optional block like `T.nilable(T.proc.void)`.
+              elsif type.is_a?(Types::Optional) && type.type.is_a?(Types::Proc)
+                method_block = MethodType::Block.new(required: false, type: type.type.type)
               else
                 STDERR.puts "Unexpected block type: #{type}"
                 PP.pp args_node, STDERR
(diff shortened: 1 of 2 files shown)


connorshea:
Rebased and ready for review :)

PR #228, opened 2020-03-15
Add benchmark.rbs

connorshea:
- Create `benchmark.rbs` from [`benchmark.rbi` in Sorbet](https://github.com/sorbet/sorbet/blob/fdb18367a187f0e33772e1ebabb4df6693f967bc/rbi/stdlib/benchmark.rbi)
- Update the docs using the annotate-with-rdoc script.
- Add methods from Benchmark::Tms.
- Add a test for `Benchmark.measure`.

Technically, `cstime`, `cutime`, `stime`, etc. are attr_readers, should the RBS file be changed to use the attr_reader syntax? Also, I haven't added signatures for some of the methods on Benchmark::Tms, please tell me if we need them to merge this PR.

### DIFF
--- a/stdlib/benchmark/benchmark.rbs
+++ b/stdlib/benchmark/benchmark.rbs
@@ -139,7 +139,7 @@ module Benchmark
   #     >total:   2.930000   0.000000   2.930000 (  2.932889)
   #     >avg:     0.976667   0.000000   0.976667 (  0.977630)
   #
-  def self?.benchmark: (String caption, ?Integer? label_width, ?String? format, *String labels) { (Report report) -> void } -> ::Array[Benchmark::Tms]
+  def self?.benchmark: (String caption, ?Integer? label_width, ?String? format, *String labels) { (Report report) -> (Array[Tms] | void) } -> ::Array[Benchmark::Tms]
 
   # A simple interface to the #benchmark method, #bm generates sequential reports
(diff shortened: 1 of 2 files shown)


connorshea:
I've updated this PR to include methods for Benchmark::Job and Benchmark::Report, since they're used in the blocks for `Benchmark.bm` and `Benchmark.bmbm`. Unfortunately, I don't think it's actually testing them properly since it's not trying to load the `Benchmark::Job` or `Benchmark::Report` classes/sigs, only `Benchmark`.

How should I do this so that it tests the signatures properly?

```ruby
array = (1..30).map { rand }

Benchmark.bmbm do |x|
  x.report("sort!") { array.dup.sort! } # Make sure Benchmark::Job#report has the correct signature and is being used correctly here.
  x.report("sort")  { array.dup.sort  }
end
```

soutaro:
@connorshea I'm sorry for the late reply. I pushed commits to the branch directly. 

* Add `BenchmarkReportTest` because the `StdLibTest` doesn't allow testing several targets in one test class.
* Fix some tests.

Could you take a look at these changes? Adding tests for `Benchmark::Job` or `Benchmark::Tms` may be great, but I think it's too much work. So, just skipping writing tests for them, for now, looks okay to me, or fixing the `StdLibTest` first would make more sense.

connorshea:
Looks good to me 👍

PR #232, opened 2020-03-20
(wip)Add type and test on RbConfig

taki:
## Description

I add type and test on RbConfig.

## Reference

-  https://docs.ruby-lang.org/ja/2.7.0/class/RbConfig.html#S_RUBY
- https://github.com/ruby/ruby/blob/db166290088fb7d39d01f68b9860253893d4f1a7/tool/mkconfig.rb


taki:
Since the code make spaghetti code and error, I will create new PR.

PR #233, opened 2020-03-21
Add type and test on RbConfig

taki:
## Description

I add type and test on RbConfig. If you OK, Once I want to merge this.

## Not Done

- Add RbConfig.fire_update! return nil. For now, I am wondering how to add a test.

## Reference

- https://docs.ruby-lang.org/ja/2.7.0/class/RbConfig.html#S_RUBY
- https://github.com/ruby/ruby/blob/db166290088fb7d39d01f68b9860253893d4f1a7/tool/mkconfig.rb

### DIFF
--- a/stdlib/builtin/rb_config.rbs
+++ b/stdlib/builtin/rb_config.rbs
@@ -18,7 +18,7 @@ module RbConfig
 
   def self.fire_update!: (untyped key, untyped val, ?untyped mkconf, ?untyped conf) -> untyped
 
-  def self.ruby: () -> untyped
+  def self.ruby: () -> String
 end
 
 # The hash configurations stored.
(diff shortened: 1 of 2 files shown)


stdlib/builtin/rb_config.rbs
```diff
@@ -12,7 +12,7 @@ module RbConfig
   #
   def self.expand: (String val, ?Hash[String, String] config) -> String
 
-  def self.fire_update!: (String key, String val, ?Hash[String, String] mkconf, ?Hash[String, String] conf) -> Array[String]
+  def self.fire_update!: (String key, String val, ?Hash[String, String] mkconf, ?Hash[String, String] conf) -> (Array[String] | nil)
```
soutaro: You can write `Array[String]?` instead of `| nil`. (I usually write `T?` instead of `T | nil`.)

taki: I fix by https://github.com/ruby/ruby-signature/pull/233/commits/eae3284bf88944096f1fa33b1283cd3de6e708d0.

taki:
Thank you.

PR #238, opened 2020-03-26
Update Thread::Mutex signature and add test

mtsmfm:


stdlib/builtin/thread.rbs
```diff
@@ -927,25 +927,25 @@ end
 class Thread::Mutex < Object
   # Attempts to grab the lock and waits if it isn’t available. Raises
   # `ThreadError` if `mutex` was locked by the current thread.
-  def lock: () -> untyped
+  def lock: () -> self
 
   # Returns `true` if this lock is currently held by some thread.
-  def locked?: () -> bool
+  def locked?: () -> (true | false)
```
soutaro: Could you keep it `bool`? It clearly looks a predicate, which programmers expect to be a boolean. I don't think we need to give more detail here. (The same for `owned?` and `try_lock`.)

mtsmfm: Fixed and force pushed

PR #239, opened 2020-03-27
Update Thread::Queue signature and add test

mtsmfm:


stdlib/builtin/thread.rbs
```diff
@@ -1038,10 +1038,8 @@ class Thread::Queue < Object
   # Also aliased as: [size](Queue.downloaded.ruby_doc#method-i-size)
   def length: () -> Integer
 
-  def marshal_dump: () -> untyped
```
mtsmfm: I can't find any documentation for `Thread::Queue#marshal_dump`

https://ruby-doc.org/core-2.7.0/Queue.html

stdlib/builtin/thread.rbs
```diff
@@ -1017,19 +1017,19 @@ class Thread::Queue < Object
   #           end
   #         }
   #         q.close
-  def close: () -> untyped
+  def close: () -> self
 
   # Returns `true` if the queue is closed.
-  def closed?: () -> bool
+  def closed?: () -> (true | false)
```
soutaro: Please use `bool` instead of `true | false`.

stdlib/builtin/thread.rbs
```diff
@@ -1053,21 +1051,21 @@ class Thread::Queue < Object
   # 
   # Also aliased as: [deq](Queue.downloaded.ruby_doc#method-i-deq),
   # [shift](Queue.downloaded.ruby_doc#method-i-shift)
-  def pop: (*untyped args) -> untyped
+  def pop: (?(true | false) non_block) -> untyped
```
soutaro: `bool` seems better type for this.

mtsmfm:
Updated and force pushed

PR #240, opened 2020-03-28
add more type check for FalseClass and Files

daongochieu2810:


soutaro:
Hi @daongochieu2810, 

Thank you so much for opening this pull request. Unfortunately, your change does not look correct with respect to the implementation of Ruby. For example, `File.absolute_path(nil)` raises a `TypeError` instead of returning an instance of `ArgumentError`.

I'm closing this PR, but let me know the context of the change if you believe your change is correct.

PR #242, opened 2020-03-28
Add signatures for the Coverage module.

connorshea:
Based on https://github.com/ruby/ruby/blob/v2_7_0/ext/coverage/coverage.c

### DIFF
--- a/stdlib/coverage/coverage.rbs
+++ b/stdlib/coverage/coverage.rbs
@@ -58,5 +58,5 @@ module Coverage
 
   # Enables coverage measurement.
   #
-  def self.start: (?lines: bool lines, ?branches: bool branches, ?methods: bool methods, ?oneshot_lines: bool oneshot_lines) -> NilClass
+  def self.start: (?lines: bool lines, ?branches: bool branches, ?methods: bool methods, ?oneshot_lines: bool oneshot_lines) -> nil
 end
--- a/stdlib/coverage/coverage.rbs
+++ b/stdlib/coverage/coverage.rbs
@@ -33,7 +33,7 @@
 #     p Coverage.result  #=> {"foo.rb"=>[1, 1, 10, nil, nil, 1, 1, nil, 0, nil]}
 #
 module Coverage
-  def self.line_stub: () -> untyped
+  def self.line_stub: () -> Array[Integer?]
 
   # Returns a hash that contains filename as key and coverage array as value. This
   # is the same as `Coverage.result(stop: false, clear: false)`.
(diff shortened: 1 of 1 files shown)


stdlib/coverage/coverage.rbs
```diff
@@ -0,0 +1,62 @@
...
+  def self.result: (?stop: bool stop, ?clear: bool clear) -> Hash[String, Array[Integer?]]
+
+  # Returns true if coverage stats are currently being collected (after
+  # Coverage.start call, but before Coverage.result call)
+  #
+  def self.running?: () -> bool
+
+  # Enables coverage measurement.
+  #
+  def self.start: (?lines: bool lines, ?branches: bool branches, ?methods: bool methods, ?oneshot_lines: bool oneshot_lines) -> NilClass
```
soutaro, line 61: I'd like to recommend writing `nil` instead of `NilClass`.

stdlib/coverage/coverage.rbs
```diff
@@ -0,0 +1,62 @@
...
+#     end
+#     [EOF]
+#
+#     require "coverage"
+#     Coverage.start
+#     require "foo.rb"
+#     p Coverage.result  #=> {"foo.rb"=>[1, 1, 10, nil, nil, 1, 1, nil, 0, nil]}
+#
+module Coverage
+  def self.line_stub: () -> untyped
```
mame, line 36: FWIW, it returns `Array[Integer?]`.

stdlib/coverage/coverage.rbs
```diff
@@ -0,0 +1,62 @@
...
+
+  # Returns a hash that contains filename as key and coverage array as value. This
+  # is the same as `Coverage.result(stop: false, clear: false)`.
+  #
+  #     {
+  #       "file.rb" => [1, 2, nil],
+  #       ...
+  #     }
+  #
+  def self.peek_result: () -> Hash[String, Array[Integer?]]
```
mame, line 46: Unfortuantely, it may return more complex data structure depending upon the argument of `Coverage.start`.  See https://bugs.ruby-lang.org/issues/15022.  It would be good to make it `Hash[String, untyped]`

stdlib/coverage/coverage.rbs
```diff
@@ -0,0 +1,62 @@
...
+
+  # Returns a hash that contains filename as key and coverage array as value. If
+  # `clear` is true, it clears the counters to zero. If `stop` is true, it
+  # disables coverage measurement.
+  #
+  def self.result: (?stop: bool stop, ?clear: bool clear) -> Hash[String, Array[Integer?]]
```
mame, line 52: The same as `peek_result`.

connorshea:
@mame I've resolved your comments 👍

PR #243, opened 2020-03-29
Add signatures for IPAddr.

connorshea:
Based on [`ipaddr.rb`](https://github.com/ruby/ruby/blob/cf191872fdeade823b4c49b53ec9b28d66a5c402/lib/ipaddr.rb).

I'm not sure if the error classes should be included in the RBS file?

e.g. from that file:
```ruby
class IPAddr
  # Generic IPAddr related error. Exceptions raised in this class should
  # inherit from Error.
  class Error < ArgumentError; end
end
```

I didn't include them here, but I can add them if they should be added.

### DIFF
--- a/stdlib/ipaddr/ipaddr.rbs
+++ b/stdlib/ipaddr/ipaddr.rbs
@@ -25,12 +25,12 @@ class IPAddr
   # Creates a new ipaddr containing the given network byte ordered string form of
   # an IP address.
   #
-  def self.new_ntoh: (untyped addr) -> IPAddr
+  def self.new_ntoh: (String addr) -> IPAddr
 
   # Convert a network byte ordered string form of an IP address into human
   # readable form.
   #
-  def self.ntop: (untyped addr) -> String
+  def self.ntop: (String addr) -> String
 
   # Creates a new ipaddr object either from a human readable IP address
   # representation in string, or from a packed in_addr value followed by an
@@ -58,7 +58,7 @@ class IPAddr
 
   # Returns a new ipaddr built by bitwise left shift.
   #
-  def <<: (untyped num) -> IPAddr
+  def <<: (Integer num) -> IPAddr
 
   # Compares the ipaddr with another.
   #
@@ -72,7 +72,7 @@ class IPAddr
 
   # Returns a new ipaddr built by bitwise right-shift.
   #
-  def >>: (untyped num) -> IPAddr
+  def >>: (Integer num) -> IPAddr
 
   # Checks equality used by Hash.
   #
--- a/stdlib/ipaddr/ipaddr.rbs
+++ b/stdlib/ipaddr/ipaddr.rbs
@@ -50,9 +50,6 @@ class IPAddr
   # as &, |, include? and ==, accept a string, or a packed in_addr value instead
   # of an IPAddr object.
   #
-  # # arglists 💪👽🚨 << Delete this section
-  #     new(addr = '::', family = Socket::AF_UNSPEC)
-  #
   def initialize: (?String addr, ?untyped family) -> IPAddr
 
   # Returns a new ipaddr built by bitwise AND.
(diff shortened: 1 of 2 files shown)


stdlib/ipaddr/ipaddr.rbs
```diff
@@ -0,0 +1,227 @@
...
+  # one explicitly by the optional second argument.
+  #
+  # Otherwise an IP address is generated from a packed in_addr value and an
+  # address family.
+  #
+  # The IPAddr class defines many methods and operators, and some of those, such
+  # as &, |, include? and ==, accept a string, or a packed in_addr value instead
+  # of an IPAddr object.
+  #
+  # # arglists 💪👽🚨 << Delete this section
```
soutaro: 👽

soutaro:
@connorshea Please include the error classes. The error classes are documented and I'm pretty sure that they are parts of the API.

connorshea:
Removed the arglist and added the error classes :)

stdlib/ipaddr/ipaddr.rbs
```diff
@@ -0,0 +1,247 @@
...
+  #
+  def eql?: (untyped other) -> untyped
+
+  # Returns the address family of this IP address.
+  #
+  def family: () -> untyped
+
+  # Returns a hash value used by Hash, Set, and Array classes
+  #
+  def hash: () -> untyped
```
soutaro, line 87: @connorshea Some methods are left `untyped`. Do you have a plan to give more precise types in this PR?

connorshea: @soutaro I've updated the file to type more of the methods now, there are still some with untyped but either they couldn't be typed (e.g. `==` can be compared with anything) or I couldn't figure out the correct type.

connorshea:
I've added some tests for IPAddr methods. Do you prefer this style of test or the `assert_send_type` style? Is there a reason to use one or the other?

stdlib/ipaddr/ipaddr.rbs
```diff
@@ -0,0 +1,247 @@
...
+#
+#     p ipaddr3                   #=> #<IPAddr: IPv4:192.168.2.0/255.255.255.0>
+#
+class IPAddr
+  include Comparable
+
+  # Creates a new ipaddr containing the given network byte ordered string form of
+  # an IP address.
+  #
+  def self.new_ntoh: (untyped addr) -> IPAddr
```
soutaro, line 28: How about making the type of the argument `String`?

stdlib/ipaddr/ipaddr.rbs
```diff
@@ -0,0 +1,247 @@
...
+
+  # Convert a network byte ordered string form of an IP address into human
+  # readable form.
+  #
+  def self.ntop: (untyped addr) -> String
```
soutaro, line 33: `addr` looks like`string` or `String`.

stdlib/ipaddr/ipaddr.rbs
```diff
@@ -0,0 +1,247 @@
...
+  #
+  def initialize: (?String addr, ?untyped family) -> IPAddr
+
+  # Returns a new ipaddr built by bitwise AND.
+  #
+  def &: (untyped other) -> IPAddr
+
+  # Returns a new ipaddr built by bitwise left shift.
+  #
+  def <<: (untyped num) -> IPAddr
```
soutaro, line 61: `num` might be `int`?

stdlib/ipaddr/ipaddr.rbs
```diff
@@ -0,0 +1,247 @@
...
+
+  # Returns true if two ipaddrs are equal.
+  #
+  def ==: (untyped other) -> bool
+
+  alias === include?
+
+  # Returns a new ipaddr built by bitwise right-shift.
+  #
+  def >>: (untyped num) -> IPAddr
```
soutaro, line 75: `int num` ?

stdlib/ipaddr/ipaddr.rbs
```diff
@@ -0,0 +1,247 @@
...
+  #
+  def native: () -> IPAddr
+
+  # Returns the prefix length in bits for the ipaddr.
+  #
+  def prefix: () -> Integer
+
+  # Sets the prefix length in bits
+  #
+  def prefix=: (Integer prefix) -> self
```
soutaro, line 171: 👏

connorshea:
I've added those types :)

PR #246, opened 2020-03-29
Update docs for Module.

connorshea:
And add some missing method parameters to Module methods. A lot of methods have an `inherit` parameter that wasn't in any of the signatures.

### DIFF
--- a/stdlib/builtin/module.rbs
+++ b/stdlib/builtin/module.rbs
@@ -1052,7 +1052,42 @@ class Module < Object
   #
   def to_s: () -> String
 
-  def undefMethod: (Symbol | String arg0) -> self
+  # Prevents the current class from responding to calls to the named method.
+  # Contrast this with `remove_method`, which deletes the method from the
+  # particular class; Ruby will still search superclasses and mixed-in modules for
+  # a possible receiver. String arguments are converted to symbols.
+  #
+  #     class Parent
+  #       def hello
+  #         puts "In parent"
+  #       end
+  #     end
+  #     class Child < Parent
+  #       def hello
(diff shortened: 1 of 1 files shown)


stdlib/builtin/module.rbs
```diff
@@ -1,248 +1,1069 @@
...
+  # If a method has protected visibility, it is callable only where `self` of the
+  # context is the same as the method. (method definition or instance_eval). This
+  # behavior is different from Java's protected method. Usually `private` should
+  # be used.
+  #
+  # Note that a protected method is slow because it can't use inline cache.
+  #
+  # To show a private method on RDoc, use `:doc:` instead of this.
+  #
   def protected: (*Symbol | String arg0) -> self
```
connorshea: Should this use backticks (\`) since protected is a keyword?

stdlib/builtin/module.rbs
```diff
@@ -1,248 +1,1069 @@
...
+  #
+  def singleton_class?: () -> bool
+
+  # Returns a string representing this module or class. For basic classes and
+  # modules, this is the name. For singletons, we show information on the thing
+  # we're attached to as well.
+  #
   def to_s: () -> String
 
   def undefMethod: (Symbol | String arg0) -> self
```
connorshea: I asked in the Sorbet Slack about this, but I think they accidentally typo'd this and it should be `undef_method`?

soutaro: Good find! It should be `undef_method`. I think you can push another commit to fix it in this PR.

connorshea:
@soutaro fixed 👍

PR #247, opened 2020-03-29
Update docs for Marshal and add restore method

connorshea:
Technically restore is an alias of load, but it's defined in C so I'm not sure if we should write it in the RBS as an alias?

https://github.com/ruby/ruby/blob/v2_7_0/marshal.c#L2077


soutaro:
I think you can make it alias.

connorshea:
I've made it an alias now.

PR #258, opened 2020-04-03
Remove `(Numeric) -> Numeric` rules from some Integer binary methods

mame:
for two reasons:

(1) These rules are less useful.  If a class inherits from Numeric, it
should use `extension` to extend the overloading for `Integer#+`, `#-`,
etc.  Without the extension, `1 + its_instance` returns `Numeric`, but
it would be difficult to use the result correctly.
Note that Liskov substitution principle does not matter here because
there is no `Numeric#+`, `#-`, etc.

(2) Type Profiler tries all overloaded rules that matches the argument
types at each call.  So, `1 + 1` is considered as `Integer | Numeric`
because `1` is an Integer and Integer is Numeric.  This is not wrong but
too conservative, so I'd like to avoid it.

### DIFF
--- a/stdlib/builtin/integer.rbs
+++ b/stdlib/builtin/integer.rbs
@@ -45,7 +45,6 @@ class Integer < Numeric
        | (Rational) -> Rational
        | (Complex) -> Complex
        | (Integer) -> Integer
-       | (Numeric) -> Numeric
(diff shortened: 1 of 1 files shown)


mame:
✔️

PR #261, opened 2020-04-05
Update ThreadGroup signature and add test

mtsmfm:


soutaro:
@mtsmfm The change looks good. Did you find any concrete problems with the old type? (or you just found the new type is better?)

PR #263, opened 2020-04-05
Add docs to StringIO.

connorshea:


connorshea:
Also, this is missing quite a few methods that Sorbet has. Maybe we should just import the RBI file? https://github.com/sorbet/sorbet/blob/master/rbi/stdlib/stringio.rbi

soutaro:
@connorshea Yes. How about trying to import Sorbet RBI file and then writing types and tests?

soutaro:
```
ruby-signature/lib/ruby/signature/errors.rb:19:in `check!': /__w/ruby-signature/ruby-signature/stdlib/builtin/string_io.rbs:67:59...67:79: ::Enumerator expects parameters [Elem, Return], but given args [::String] (Ruby::Signature::InvalidTypeApplicationError)
373
```

🔪

connorshea:
@soutaro I think the `Enumerator[String]` should be changed to `Enumerator[String, self]` in this PR, right?

I also just opened an issue about fixing this in the rbi converter :) https://github.com/ruby/ruby-signature/issues/276

PR #264, opened 2020-04-06
Update Exception signature and add test

mtsmfm:


stdlib/builtin/exception.rbs
```diff
@@ -124,7 +124,7 @@
 class Exception < Object
   def self.to_tty?: () -> bool
 
-  def self.exception: (?String msg) -> self
+  def self.exception: (?String msg) -> Exception
```
mtsmfm, line 127: I think `self` means `Exception` class itself so the test should be failed...?

soutaro: Right, `self` means `singleton(::Exception)` here, and your correction makes sense!

PR #265, opened 2020-04-06
Fix bundler/setup loading error in install task

Morriar:
I followed the README to install `ruby-signature` and running `bin/setup` ended with the following error:

```
+ bundle exec rake parser
racc -v -o lib/ruby/signature/parser.rb lib/ruby/signature/parser.y
	2: from /usr/local/Cellar/ruby@2.5/2.5.5_1/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:39:in `require'
	1: from /usr/local/Cellar/ruby@2.5/2.5.5_1/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:135:in `rescue in require'
/usr/local/Cellar/ruby@2.5/2.5.5_1/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:135:in `require': cannot load such file -- bundler/setup (LoadError)
rake aborted!
```

Adding `bundle exec` to the `racc -v ...` command fixed the problem.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -27,7 +27,7 @@ task :rubocop do
 end
 
 rule ".rb" => ".y" do |t|
-  sh "racc -v -o #{t.name} #{t.source}"
+  sh "bundle exec racc -v -o #{t.name} #{t.source}"
 end
 
 task :parser => "lib/ruby/signature/parser.rb"


soutaro:
@Morriar Thank you for reporting an issue. I'm not sure why adding `bundle exec` solves your problem, because we already have it in `bin/setup`.

https://github.com/ruby/ruby-signature/blob/a0791be70147e3635c80e706e9ba4f2782aa5955/bin/setup#L10

So my guess is it is another problem, and it may be because you run the command with ruby-2.5 (from the stack trace) which we don't have plan to support. (We support ruby >=2.6.)

Morriar:
You're right, all is working well on >-= 2.6.

Thanks!

PR #266, opened 2020-04-07
Add alias for Thread families

mtsmfm:


soutaro:
I think these definitions are classes, so thinking of singleton types `singleton(Thread::Mutex)` would be the correct ones?

mtsmfm:
Ah, yes, true

PR #268, opened 2020-04-08
Output test methods via test generator

mtsmfm:
`bundle exec rake 'generate:stdlib_test[Encoding]'` outputs the following result:

```ruby
require_relative "test_helper"

class EncodingTest < StdlibTest
  target Encoding
  # library "pathname", "set", "securerandom"     # Declare library signatures to load
  using hook.refinement

  def test_class_method_aliases
    # () -> ::Hash[::String, ::String]
    Encoding.aliases
  end

  def test_class_method_compatible?
    # (untyped obj1, untyped obj2) -> ::Encoding?
    Encoding.compatible?
  end

  def test_class_method_default_external
    # () -> ::Encoding
    Encoding.default_external
  end

  def test_class_method_default_external=
    # (::String arg0) -> ::String
    Encoding.default_external=
    # (::Encoding arg0) -> ::Encoding
    Encoding.default_external=
  end

  def test_class_method_default_internal
    # () -> ::Encoding
    Encoding.default_internal
  end

  def test_class_method_default_internal=
    # (::String arg0) -> ::String?
    Encoding.default_internal=
    # (::Encoding arg0) -> ::Encoding?
    Encoding.default_internal=
  end

  def test_class_method_find
    # (::String | ::Encoding arg0) -> ::Encoding
    Encoding.find
  end

  def test_class_method_list
    # () -> ::Array[::Encoding]
    Encoding.list
  end

  def test_class_method_name_list
    # () -> ::Array[::String]
    Encoding.name_list
  end

  def test_inspect
    # () -> ::String
    Encoding.new.inspect
  end

  def test_to_s
    # () -> ::String
    Encoding.new.to_s
  end

  def test_ascii_compatible?
    # () -> bool
    Encoding.new.ascii_compatible?
  end

  def test_dummy?
    # () -> bool
    Encoding.new.dummy?
  end

  def test_name
    # () -> ::String
    Encoding.new.name
  end

  def test_names
    # () -> ::Array[::String]
    Encoding.new.names
  end

  def test_replicate
    # (::String name) -> ::Encoding
    Encoding.new.replicate
  end
end
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -44,21 +44,96 @@ namespace :generate do
     path = Pathname("test/stdlib/#{klass}_test.rb")
     raise "#{path} already exists!" if path.exist?
 
-    path.write <<~RUBY
-      require_relative "test_helper"
-      
-      class #{klass}Test < StdlibTest
-        target #{klass}
-        # library "pathname", "set", "securerandom"     # Declare library signatures to load
-        using hook.refinement
+    require "erb"
+    require "ruby/signature"
 
-        # def test_method_name
-        #   # Call the method
-        #   method_name(arg)
-        #   method_name(arg, arg2)
-        # end
+    class TestTemplateBuilder
+      attr_reader :klass, :env
+
+      def initialize(klass)
+        @klass = klass
+
+        @env = Ruby::Signature::Environment.new
+        Ruby::Signature::EnvironmentLoader.new.load(env: @env)
       end
-    RUBY
+
+      def call
+        ERB.new(<<~ERB, trim_mode: "-").result(binding)
+          require_relative "test_helper"
+
+          class <%= klass %>Test < StdlibTest
+            target <%= klass %>
+            # library "pathname", "set", "securerandom"     # Declare library signatures to load
+            using hook.refinement
+          <%- class_methods.each do |method_name, definition| %>
+            def test_class_method_<%= test_name_for(method_name) %>
+            <%- definition.method_types.each do |method_type| -%>
+              # <%= method_type %>
+              <%= klass %>.<%= method_name %>
+            <%- end -%>
+            end
+          <%- end -%>
+          <%- instance_methods.each do |method_name, definition| %>
+            def test_<%= test_name_for(method_name) %>
+            <%- definition.method_types.each do |method_type| -%>
+              # <%= method_type %>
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -44,21 +44,82 @@ namespace :generate do
...
+            <%- end -%>
+            end
+          <%- end -%>
+          end
+        ERB
+      end
+
+      private
+
+      def test_name_for(method_name)
```
connorshea, line 90: This also needs `<=>`, `<=`, `>=`, `>`, `<`, `+`, `-`, `*`, `/`, `**`, etc. There's a table on this page with a bunch of the special method names: https://ruby-doc.org/core-2.4.0/doc/syntax/methods_rdoc.html

connorshea: Something like this?:
```ruby
special_methods_map = {
  '==': 'double_equal',
  '!=': 'not_equal'
  '===': 'triple_equal',
  '[]': 'square_bracket',
  '[]=': 'square_bracket_assign',
  '>': 'greater_than',
  '<': 'less_than',
  '>=': 'greater_than_equal_to',
  '<=': 'less_than_equal_to',
  '<=>': 'spaceship',
  '+': 'plus',
  '-': 'minus',
  '*': 'multiply',
  '/': 'divide',
  '**': 'power',
  '%': 'modulus',
  '&': 'and',
  '^': 'xor',
  '>>': 'right_shift',
  '<<': 'left_shift',
  '=~': 'pattern_match'
  '!~': 'does_not_match'
}

return special_methods_map[method_name] if special_methods_map.keys.include?(method_name)

return method_name
```

mtsmfm: I'm wondering if defining test methods via `define_method` is acceptable

```ruby
define_method "test==" do
end
```

soutaro: Using `define_method` would work, but I think adding renamings here for them too makes more sense. (Choosing `def` + renaming or `define_method` would look like strange to me. We already have _renaming table_ here. etc...)

soutaro:
@mtsmfm This is a great feature! I really love it. 💯 🎉

Rakefile
mtsmfm: I updated name mapping

PR #278, opened 2020-04-19
Fix typing of `fail` and add tests

alts:
Hi there. I was trying to add signature to an existing project, and found that checking through steep was failing on a few `raise` calls.

Previous the following forms would fail checking:

```ruby
raise Exception, 'message'
raise object_that_implements_exception_method
```

My understanding of the first argument to `raise` is that it can be anything that adheres to the exception interface of exposing a `exception` method. Switching `?Exception` for the interface `_Exception` allows for exception classes, exception instances, and objects that adhere to that interface to all pass checking.

### DIFF
--- a/lib/ruby/signature/test/hook.rb
+++ b/lib/ruby/signature/test/hook.rb
@@ -164,16 +164,22 @@ module Ruby
               nil
             end
             prepended = klass.ancestors.include?(hook.instance_module) || singleton_klass&.ancestors&.include?(hook.singleton_module)
-            result = if prepended
-                       method.super_method.call(*args, &block)
-                     else
-                       # Using refinement
-                       method.call(*args, &block)
-                     end
+            exception = nil
+            result = begin
+               if prepended
+                 method.super_method.call(*args, &block)
+               else
+                 # Using refinement
+                 method.call(*args, &block)
+               end
+            rescue Exception => e
+              exception = e
+              nil
+            end
 
             hook.logger.debug { "#{method_name} returns: #{hook.inspect_(result)}" }
 
-            call = CallTrace.new(method_call: ArgumentsReturn.new(arguments: args, return_value: result, exception: nil),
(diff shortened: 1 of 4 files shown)


connorshea:
The typing of `raise` probably comes from Sorbet, and the second argument can be essentially anything: https://github.com/sorbet/sorbet/pull/2859

alts:
Sounds good. I've changed the type of the second argument.

Would you recommend changing the definitions for `Exception` to match? Currently it presumes a string first argument

lib/ruby/signature/test/hook.rb
```diff
@@ -164,16 +164,22 @@ def delegation(name, method_types, method_name)
...
+               end
+            rescue Exception => e
+              exception = e
+              nil
+            end
 
             hook.logger.debug { "#{method_name} returns: #{hook.inspect_(result)}" }
 
-            call = CallTrace.new(method_call: ArgumentsReturn.new(arguments: args, return_value: result, exception: nil),
+            call = CallTrace.new(method_call: ArgumentsReturn.new(arguments: args, return_value: result, exception: exception),
```
soutaro, line 182: 👍

soutaro:
@alts Thank you for sending your patch!

I don't think revising the definition of `Exception` makes much sense. The type is too complicated and cannot be described very well in RBS. So, if the current patch works well for your project, we can use the version.

Can I confirm that you really use the duck-typed `#exception` method? (Without having `::Exception` ancestor class???) I haven't used that way before, so I'm really surprised if someone really wrote the code!

alts:
@soutaro I've never seen the duck-typed version either. I only learned that it was possible when investigating how to write the type definition to support both `Exception` and `Type[Exception]`. I didn't see a way of expressing the `Type[Exception]` form in RBS (that's how I would write it in typed Python), but while reading the documentation, learned that the expected contract is just implementing `#exception`.

I can also change the `raise` signature to look more like Sorbet's, and accept any `Class` as the first argument.

soutaro:
@alts Thanks! Merged the PR. We will revise the signature when we detected further issues.

PR #279, opened 2020-04-24
Add JSON

hanachin:


ruby-signature.gemspec
```diff
@@ -42,4 +42,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rubocop"
   spec.add_development_dependency "rubocop-rubycw"
   spec.add_development_dependency "minitest-reporters", "~> 1.3.6"
+  spec.add_development_dependency "json", "~> 2.3.0"
```
soutaro, line 45: Okay, JSON is a gem... I think the best way is to put RBS in JSON gem, but I think it's okay to have the signatures here for now.

stdlib/builtin/builtin.rbs
```diff
@@ -37,3 +37,4 @@ type int = Integer | _ToInt
 type real = Integer | Float | Rational
 
 type string = String | _ToStr
+type encoding = Encoding | string
```
soutaro, line 40: 👍

stdlib/json/json.rbs
```diff
@@ -0,0 +1,360 @@
...
+  #
+  def self.dump_default_options: () -> json_options
+
+  def self.dump_default_options=: (json_options) -> json_options
+
+  def self.fast_generate: (_ToJson obj, ?json_options opts) -> String
+
+  alias self.fast_unparse self.fast_generate
+
+  def self.generate: (_ToJson obj, ?json_options opts) -> String
```
soutaro, line 143: Could you rewrite these module functions using `def self?.generate: ...` syntax?

hanachin: Thanks for the review! I updated the signature to use the `def self?.` syntax.

hanachin:
@soutaro Done :v:

PR #286, opened 2020-05-08
Fix order of arguments in `inject`

alts:
One of the signatures for `inject` appears to have the `initial` and `obj` arguments swapped. 

From ruby-doc

```
reduce(initial) { |memo, obj| block } → obj
```

From Enumerable.rbs

```
[A] (A initial) { (Elem, A) -> A } -> Elem
```

I suspect that `(untyped initial) { (untyped, Elem) -> untyped } -> untyped` might be most honest, as the block doesn't have to be bound by the types of its arguments. If it is valuable to have a signature where `initial` differs from `Elem`, I think the one I've added is more correct.

Would anyone be able to help me write a test case for this? I wasn't able to set up a test case that demonstrated the argument swapping, and instead have been using steep with my branch as submodule to confirm my change works.

```ruby
items = ['a', 'b', 'c']
items.reduce(0) do |memo, item|
  memo ^ item.hash
end
```

### DIFF
--- a/stdlib/builtin/enumerable.rbs
+++ b/stdlib/builtin/enumerable.rbs
@@ -133,7 +133,7 @@ module Enumerable[unchecked out Elem, out Return]: _Each[Elem, Return]
 
   def inject: (untyped init, Symbol method) -> untyped
             | (Symbol method) -> untyped
-            | [A] (A initial) { (Elem, A) -> A } -> Elem
+            | [A] (A initial) { (A, Elem) -> A } -> A
             | () { (Elem, Elem) -> Elem } -> Elem
 
   # Returns the object in *enum* with the maximum value. The first form
--- a/test/stdlib/Enumerable_test.rb
+++ b/test/stdlib/Enumerable_test.rb
@@ -1,4 +1,5 @@
 require_relative "test_helper"
+require "rbs/test/test_helper"
 
 class EnumerableTest < StdlibTest
   target Enumerable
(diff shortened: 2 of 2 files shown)


soutaro:
Thank you @alts for pointing out this issue!

I believe `[A] (A initial) { (A, Elem) -> A } -> A` is the best type for the case with an initial value. (The original definition is simply wrong...)

Test with `StdlibTest` doesn't work well because of the lack of verification around type variables. Writing tests with `assert_send_type` would help (it requires some setup).

```rb
class EnumerableInstanceTest < Minitest::Test
  include RBS::Test::TypeAssertions

  testing "::Enumerable[::Integer, void]"

  def enumerable
    ...
  end

  def test_inject
    assert_send_type "(::String) { (::String, ::Integer) -> ::String } -> ::String",
                     enumerable, :inject, "" do |x, y| x + y.to_s end
  end
end
```

alts:
Ahh, thank you @soutaro. That helped a lot.

I've kept the type variables in place for the `assert_send_type` calls.

```ruby
assert_send_type("[A] (A initial) { (A, String) -> A } -> A", TestEnumerable.new, :inject, 0) ...
```

Let me know if you think it'd be better to make them concrete (like `String`).

soutaro:
@alts My opinion here is concrete type is better than type variable because the runtime type checker currently doesn't check any correctness against type variables. So, could you revise the method types?

https://github.com/ruby/ruby-signature/blob/0829bdf6c0510f9d9042cbd7b78cbb622ca4706b/lib/rbs/test/type_check.rb#L229-L230

We can implement the check by collecting constraints on type variables and doing checks after method call, but I haven't tried it yet.

alts:
@soutaro Ah! I didn't realize the variable type check was just `true`. Ok, I've switched the signatures to be concrete

PR #292, opened 2020-05-30
Improve Fiber typing

pocke:
This pull request improves Fiber typing.

It changes three things.

First, it updates `FIber#initialize` type.
The returned value of the block that is passed to `Fiber#initialize` can be used. So it should not be `void`.

```ruby
Fiber.new do
  42
end.tap do |f|
  p f.resume # => 42
end
```

Second, it adds `Fiber#raise` typing. The method is available since Ruby 2.7.
https://bugs.ruby-lang.org/issues/10344
Note that the returned value should be `untyped`, not `bot` like `Kernel.raise`. Because it returned a value if the raised exception is rescued in the child fiber.

```ruby
Fiber.new do
  Fiber.yield
rescue
  42
end.tap do |f|
  f.resume
  p f.raise # => 42
end
```

Third, it replaces the old-style test with `assert_send_type`.

### DIFF
--- a/stdlib/builtin/fiber.rbs
+++ b/stdlib/builtin/fiber.rbs
@@ -62,7 +62,11 @@
 class Fiber < Object
   def self.yield: (*untyped args) -> untyped
 
-  def initialize: () { () -> void } -> Fiber
+  def initialize: () { () -> untyped } -> void
 
   def resume: (*untyped args) -> untyped
+
+  def raise: () -> untyped
+           | (string message) -> untyped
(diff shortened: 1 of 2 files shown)


stdlib/builtin/fiber.rbs
```diff
@@ -62,7 +62,11 @@
 class Fiber < Object
   def self.yield: (*untyped args) -> untyped
 
-  def initialize: () { () -> void } -> Fiber
+  def initialize: () { () -> untyped } -> void
```
soutaro, line 65: 👏

PR #302, opened 2020-06-10
Workaround: just skip anonymous superclasses and included modules

### LINKED ISSUE #301: `rbs prototype runtime` didn't work with a class including anonymous modules
With class `C` in `./c.rb`

```ruby
class C
  include Module.new
end
```

this command fails.

```
$ bundle2.7 exec rbs prototype runtime --require ./c C
	13: from /home/tadashi/git/rbs/vendor/bundle/ruby/2.7.0/bin/rbs:23:in `<main>'
	12: from /home/tadashi/git/rbs/vendor/bundle/ruby/2.7.0/bin/rbs:23:in `load'
	11: from /home/tadashi/git/rbs/exe/rbs:7:in `<top (required)>'
	10: from /home/tadashi/git/rbs/lib/rbs/cli.rb:85:in `run'
	 9: from /home/tadashi/git/rbs/lib/rbs/cli.rb:455:in `run_prototype'
	 8: from /home/tadashi/git/rbs/lib/rbs/prototype/runtime.rb:40:in `decls'
	 7: from /home/tadashi/git/rbs/lib/rbs/prototype/runtime.rb:40:in `each'
	 6: from /home/tadashi/git/rbs/lib/rbs/prototype/runtime.rb:43:in `block in decls'
	 5: from /home/tadashi/git/rbs/lib/rbs/prototype/runtime.rb:323:in `generate_class'
	 4: from /home/tadashi/git/rbs/lib/rbs/prototype/runtime.rb:64:in `each_mixin'
	 3: from /home/tadashi/git/rbs/lib/rbs/prototype/runtime.rb:64:in `each'
	 2: from /home/tadashi/git/rbs/lib/rbs/prototype/runtime.rb:66:in `block in each_mixin'
	 1: from /home/tadashi/git/rbs/lib/rbs/prototype/runtime.rb:329:in `block in generate_class'
/home/tadashi/git/rbs/lib/rbs/prototype/runtime.rb:53:in `to_type_name': undefined method `split' for nil:NilClass (NoMethodError)
```

This is triggered by the below line because anonymous modules have a name `nil`
https://github.com/ruby/rbs/blob/2646a386f5d412259fd7df601c9a6ad0a0f2b691/lib/rbs/prototype/runtime.rb#L53

but I don't know what to do, so I submitted this issue.

Side-note: This always happened when I applied `rbs prototype runtime SomeController` to a Rails application controller. Since controllers, at least with Rails 6.0.3.1, seem to always include [an anonymous module](https://github.com/rails/rails/blob/v6.0.3.1/actionpack/lib/action_dispatch/routing/route_set.rb#L481), so the execution fails all the time.

tadd:
Fixes #301 and https://github.com/ruby/rbs/issues/301#issuecomment-641747375 based on the direction of https://github.com/ruby/rbs/issues/301#issuecomment-641726359.

For this `./c.rb`,

```ruby
class C < Class.new
  include Module.new
  def x; end
end
```

the command `bundle exec rbs prototype runtime --require ./c C` generates warning messages and  signatures:

```
W, [2020-06-10T15:39:12.732355 #28227]  WARN -- rbs: Skipping anonymous superclass #<Class:0x000055b123e491a0>
W, [2020-06-10T15:39:12.732475 #28227]  WARN -- rbs: Skipping anonymous module #<Module:0x000055b123e49010>
class C
  public

  def x: () -> untyped
end
```

### DIFF
--- a/lib/rbs/prototype/runtime.rb
+++ b/lib/rbs/prototype/runtime.rb
@@ -281,6 +281,11 @@ module RBS
           value = mod.const_get(name)
 
           next if value.is_a?(Class) || value.is_a?(Module)
+          unless value.class.name
+            RBS.logger.warn("Skipping constant #{name} #{value} of #{mod} as an instance of anonymous class")
+            next
+          end
+
           type = case value
                  when true, false
                    Types::Bases::Bool.new(location: nil)
@@ -306,6 +311,9 @@ module RBS
         type_name = to_type_name(mod.name)
         super_class = if mod.superclass == ::Object
                         nil
+                      elsif mod.superclass.name.nil?
+                        RBS.logger.warn("Skipping anonymous superclass #{mod.superclass} of #{mod}")
+                        nil
                       else
                         AST::Declarations::Class::Super.new(name: to_type_name(mod.superclass.name), args: [])
                       end
@@ -321,6 +329,11 @@ module RBS
         )
 
         each_mixin(mod.included_modules, *mod.superclass.included_modules, *mod.included_modules.flat_map(&:included_modules)) do |included_module|
+          unless included_module.name
+            RBS.logger.warn("Skipping anonymous module #{included_module} included in #{mod}")
+            next
+          end
+
           module_name = to_type_name(included_module.name)
           if module_name.namespace == type_name.namespace
             module_name = TypeName.new(name: module_name.name, namespace: Namespace.empty)
@@ -343,6 +356,11 @@ module RBS
       end
 
       def generate_module(mod)
+        unless mod.name
+          RBS.logger.warn("Skipping anonymous module #{mod}")
+          return
+        end
+
         type_name = to_type_name(mod.name)
 
         decl = AST::Declarations::Module.new(
@@ -356,6 +374,11 @@ module RBS
         )
 
         each_mixin(mod.included_modules, *mod.included_modules.flat_map(&:included_modules), namespace: type_name.namespace) do |included_module|
+          unless included_module.name
+            RBS.logger.warn("Skipping anonymous module #{included_module} included in #{mod}")
+            next
+          end
+
           module_name = to_type_name(included_module.name)
           if module_name.namespace == type_name.namespace
             module_name = TypeName.new(name: module_name.name, namespace: Namespace.empty)
--- a/test/rbs/runtime_prototype_test.rb
+++ b/test/rbs/runtime_prototype_test.rb
@@ -155,4 +155,30 @@ end
       end
     end
   end
+
+  module TestForAnonymous
+    class C < Class.new
+    end
+
+    class C2
+      include Module.new
+    end
+
+    module M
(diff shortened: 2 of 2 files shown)


pocke:
I guess `generate_module` method has the same problem (I don't test it).
https://github.com/ruby/rbs/blob/76f6b9eef89a03eea3f5ce46e007bda41c3aef74/lib/rbs/prototype/runtime.rb#L359


```ruby
module M
  include Module.new
end
```

tadd:
@pocke Thanks, you're right. I fixed it with 67ecdce.

In addition, though this may be a corner case, I fixed a similar case with 23c3574 for constants.

```
module M
   C = Class.new.new # instance of an anonymous class
end
```

Because it yielded useless output.

```
M::C: #<Class:0x00005642f1feeea0>
```

lib/rbs/prototype/runtime.rb
```diff
@@ -321,6 +329,11 @@ def generate_class(mod)
         )
 
         each_mixin(mod.included_modules, *mod.superclass.included_modules, *mod.included_modules.flat_map(&:included_modules)) do |included_module|
+          unless included_module.name
+            RBS.logger.warn("Skipping anonymous module #{included_module}")
+            next
+          end
```
tadd, line 335: Should I move this and [unless clause below](https://github.com/ruby/rbs/pull/302/files#diff-d15f6899dad56f0e659ff044e5c71b36R377-R381) to the inside of `def each_mixin`?

soutaro:
@tadd Could you add tests to confirm that the RBI generation doesn't result in an exception?

lib/rbs/prototype/runtime.rb
soutaro: My opinion is to keep the test here.

tadd:
I'm trying to add tests.

tadd:
a0e37db: A test added for `decls` method.

At first, I tried to add tests for `generate_{constants,class,module}` separately, but it seemed to require some unnatural things:

* Manipulating instance vars with `p.instance_variable_set(:@decls, [])` before each tests, or
* Changing signatures of `generate_*` to return an Array
  * This may yield (a bit of?) slowdown

So I chose the current style but any comments welcome.

0cc0afa: Warning messages improved. Now warns like

```
Skipping anonymous superclass #<Class:0x00005623fddad4b0> of A::C
Skipping anonymous module #<Module:0x00005623fddad320> included in A::C2
Skipping constant CONST #<#<Class:0x00005623fddad258>:0x00005623fddad168> of A::M as an instance of anonymous class
Skipping anonymous module #<Module:0x00005623fddad028> included in A::M2
```

for a file of

```ruby
module A
  class C < Class.new
  end

  class C2
    include Module.new
  end

  module M
    CONST = Class.new.new
  end

  module M2
    include Module.new
  end
end
```

tadd:
@soutaro Thanks. I will squash all commits into one.

PR #303, opened 2020-06-14
`rbs prototype rb` guess return value type even if "def" has multiple statements

pocke:
# Background 

Since #209, `rbs prototype rb` has been aware of returned type. But it only can be aware of the method body has only one statement and it is literal.

For example:

```ruby
# It will be: def good: () -> "foo"
def good
  "foo"
end

# But it will be: def bad: () -> untyped
def bad
  something
  "foo"
end
```

# What is this pull request

This pull request solves the `bad: () -> untyped`. It will be `bad: () -> "foo"`.

Previously `rbs prototype rb` only guessed returned types for methods that contain only one literal.
But it will guess types for `return` statements and the last statement of the method.

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -405,7 +405,7 @@ module RBS
             value_types << literal_to_type(v)
           end
 
-          if key_types.all? { |t| t.is_a?(Types::Literal) }
+          if !key_types.empty? && key_types.all? { |t| t.is_a?(Types::Literal) }
             fields = key_types.map { |t| t.literal }.zip(value_types).to_h
             Types::Record.new(fields: fields, location: nil)
           else
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -328,7 +328,19 @@ module RBS
         body = node.children[2]
         return Types::Bases::Nil.new(location: nil) unless body
 
-        literal_to_type(body)
+        if body.type == :BLOCK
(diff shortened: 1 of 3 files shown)


lib/rbs/prototype/rb.rb
```diff
@@ -408,9 +420,11 @@ def literal_to_type(node)
 
       def types_to_union_type(types)
         return untyped if types.empty?
-        return untyped if types.include?(untyped)
```
pocke, line 411: Previously `Something | untyped` is squashed to `untyped`. But I think `Something | untyped` is useful for humans when reviewing the generated types. (The original idea is the type profiler.)
So it will not be squashed by this pull request.

lib/rbs/prototype/rb.rb
```diff
@@ -393,7 +405,7 @@ def literal_to_type(node)
             value_types << literal_to_type(v)
           end
 
-          if key_types.all? { |t| t.is_a?(Types::Literal) }
+          if !key_types.empty? && key_types.all? { |t| t.is_a?(Types::Literal) }
```
pocke, line 408: An empty record is syntax error, so it should be `Hash[untyped, untyped]` instead.

test/test_helper.rb
```diff
@@ -128,6 +128,9 @@ def assert_write(decls, string)
     writer.write(decls)
 
     assert_equal string, writer.out.string
+
+    # Check syntax error
+    RBS::Parser.parse_signature(writer.out.string)
```
soutaro, line 133: 👍

PR #306, opened 2020-06-15
Add Mutex_m type

pocke:
This pull request adds Mutex_m module (mutex_m library) types.

### DIFF
--- a/stdlib/mutex_m/mutex_m.rbs
+++ b/stdlib/mutex_m/mutex_m.rbs
@@ -1,0 +1,31 @@
+module Mutex_m
+  def self.append_features: (untyped cl) -> untyped
+
+  def self.define_aliases: (untyped cl) -> untyped
+
+  def self.extend_object: (untyped obj) -> untyped
+
+  public
+
+  def mu_extended: () -> untyped
+
+  def mu_lock: () -> untyped
+
+  def mu_locked?: () -> untyped
+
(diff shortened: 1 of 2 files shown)


stdlib/mutex_m/mutex_m.rbs
```diff
@@ -0,0 +1,77 @@
...
+
+  # See Mutex#sleep
+  #
+  def sleep: (?Numeric timeout) -> Integer
+
+  alias locked? mu_locked?
+  alias lock mu_lock
+  alias unlock mu_unlock
+  alias try_lock mu_try_lock
+  alias synchronize mu_synchronize
```
pocke, line 68: These aliases are defined when Mutex_m is included.
https://github.com/ruby/ruby/blob/d1fb446b62b1e114252606dcf040dd9392f25170/lib/mutex_m.rb#L45-L53

stdlib/mutex_m/mutex_m.rbs
```diff
@@ -0,0 +1,77 @@
...
+module Mutex_m
+  def self.append_features: (Module cl) -> untyped
+
+  def self.define_aliases: (Module cl) -> untyped
+
+  def self.extend_object: (Object obj) -> untyped
+
+  public
+
+  def mu_extended: () -> untyped
```
pocke, line 38: These four methods are internal API that are called when it is included/extended.
So I didn't add any tests for them.

PR #314, opened 2020-06-27
Update generate:stdlib_test

pocke:
This pull request contains several fixes for `generate:stdlib_test` rake task.

First, it raises an error for accessing unknown constant.
Because it requires `rbs` library but it accesses to `Ruby::Signature` constant.
So I replaced `Ruby::Signature` with `RBS`.

Second, it generates the old-style test cases, which uses `StdlibTest`. So I replaced it with the new style test cases.

Third, it isn't aware of libraries. It means the generator does not work for the standard libraries. For example, `rake generate:stdlib_test[Pathname]` doesn't work.

So I updated the generator to load all standard libraries.
It also loads unnecessary libraries, but I think it is the easiest way.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -26,3 +26,5 @@ jobs:
         bin/setup
         bundle exec rake
         bundle exec rake build
+        RBS_GENERATE_TEST_PATH=/tmp/Array_test.rb bundle exec rake 'generate:stdlib_test[Array]'
+        ruby -c /tmp/Array_test.rb
--- a/Rakefile
+++ b/Rakefile
@@ -66,34 +66,55 @@ namespace :generate do
       def initialize(klass)
         @klass = klass
 
-        @env = Environment.from_loader(Ruby::Signature::EnvironmentLoader.new).resolve_type_names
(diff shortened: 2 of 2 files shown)


soutaro:
I'd suggest adding a simple test for the task, like running `rake generate:stdlib_test[TESTTESTTEST]` and see if it finished without error. @pocke

pocke:
> I'd suggest adding a simple test for the task, like running `rake generate:stdlib_test[TESTTESTTEST]` and see if it finished without error. @pocke


Thanks for your suggestion! It is a good idea :+1: 


I've added a simple test in https://github.com/ruby/rbs/pull/314/commits/40724994fff19889a4017e77d7aa813261cfbd16.

.github/workflows/ruby.yml
```diff
@@ -26,3 +26,5 @@ jobs:
         bin/setup
         bundle exec rake
         bundle exec rake build
+        RBS_GENERATE_TEST_PATH=/tmp/Array_test.rb bundle exec rake 'generate:stdlib_test[Array]'
```
soutaro, line 29: 😄

PR #316, opened 2020-06-28
Add logger library types

pocke:
This pull request adds logger library types.


stdlib/logger/log_device.rbs
```diff
@@ -0,0 +1,39 @@
+class Logger
+  class LogDevice
+    # TODO: Write type signature for MonitorMixin
+    # include MonitorMixin
```
pocke, line 4: RBS doesn't have MonitorMIxin type, so I commented out it with a TODO comment.

PR #319, opened 2020-06-28
Remove unnecessary 💪👽🚨

pocke:


connorshea:
Might be a good idea to add a script that'd search all the RBS files for `💪👽🚨 << Delete this section`? Then we can run it in CI and make sure these don't get added in the future :)

PR #323, opened 2020-07-04
Add sampling option for arrays

Afront:


lib/rbs/test/setup.rb
```diff
@@ -19,6 +19,12 @@
   STDERR.puts "  [OPTIONAL] RBS_TEST_OPT: options for signatures (`-r` for libraries or `-I` for signatures)"
   STDERR.puts "  [OPTIONAL] RBS_TEST_LOGLEVEL: one of debug|info|warn|error|fatal (defaults to info)"
   STDERR.puts "  [OPTIONAL] RBS_TEST_RAISE: specify any value to raise an exception when type error is detected"
+  STDERR.puts "  [OPTIONAL] RBS_TEST_ARRAY_NO_SAMPLE: if set, all the values of an Array object would be type-checked"
+  STDERR.puts "  [OPTIONAL] RBS_TEST_ENUMERATOR_NO_SAMPLE: if set, all the values of an Enumerator object would be type-checked"
+  STDERR.puts "  [OPTIONAL] RBS_TEST_HASH_NO_SAMPLE: if set, all the values of a Hash object would be type-checked"
+  STDERR.puts "  [OPTIONAL] RBS_TEST_NO_SAMPLING: if set, all the values of any type of collection would be type-checked"
```
soutaro, line 25: Having only `RBS_TEST_NO_SAMPLING` would work fine (without `RBS_TEST_ABC_SAMPLE`).

lib/rbs/test/type_check.rb
```diff
@@ -4,9 +4,10 @@ class TypeCheck
       attr_reader :self_class
       attr_reader :builder
 
-      def initialize(self_class:, builder:)
+      def initialize(self_class:, builder:, array_no_sample:)
```
soutaro, line 7: Changing the keyword name fro `no_sample:` would be less confusing.

lib/rbs/test/type_check.rb
```diff
@@ -151,6 +152,10 @@ def zip_args(args, fun, &block)
         end
       end
 
+      def values(vals, type)
```
soutaro, line 155: I'm not sure if extracting the _test with sampling_ to a method makes sense.
I think the `Hash` sampling would be like:

```rb
val.keys.sample(size).all? {|k| v = val[k]; value(...) }
```

Which cannot be used with the `values` method.

I'd like to suggest that trying writing sampling for `Hash` and then do refactoring.

test/stdlib/Time_test.rb
```diff
@@ -174,7 +174,7 @@ def test_subsec
   end
 
   def test_succ
-    Time.new.succ
+    Time.new + 1
```
soutaro, line 177: @Afront This is not good... We need a test for `#succ`, not `+1`.

soutaro:
Thank you @Afront. 🎉 🎆 🚀 👏

PR #333, opened 2020-07-12
-114 reduce/reduce conflicts

soutaro:
🤯

### DIFF
--- a/lib/rbs/parser.y
+++ b/lib/rbs/parser.y
@@ -520,7 +520,7 @@ rule
 
   method_name:
       tOPERATOR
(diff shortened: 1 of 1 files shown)


lib/rbs/parser.y
```diff
@@ -520,7 +520,7 @@ rule
 
   method_name:
       tOPERATOR
-    | kAMP | kHAT | kSTAR | kLT | kEXCLAMATION | kSTAR2 | kBAR | kOUT | kIN
+    | kAMP | kHAT | kSTAR | kLT | kEXCLAMATION | kSTAR2 | kBAR
```
soutaro, line 523: `kOUT` and `kIN` are duplicated with https://github.com/ruby/rbs/pull/333/files#diff-10b5d394fc77a4c862e8f34a520bf417R551.

PR #336, opened 2020-07-15
Add sample size option

Afront:
This PR replaces `RBS_TEST_NO_SAMPLE` with an option to change the amount of values in a collection to be type-checked called `RBS_TEST_SAMPLE_SIZE`.

### DIFF
--- a/lib/rbs/test/setup.rb
+++ b/lib/rbs/test/setup.rb
@@ -1,8 +1,9 @@
 require "rbs"
 require "rbs/test"
-
 require "optparse"
 require "shellwords"
+
+include RBS::Test::SetupHelper
 
 logger = Logger.new(STDERR)
 
@@ -10,15 +11,19 @@ begin
   opts = Shellwords.shellsplit(ENV["RBS_TEST_OPT"] || "-I sig")
   filter = ENV.fetch('RBS_TEST_TARGET').split(',').map! { |e| e.strip }
   skips = (ENV['RBS_TEST_SKIP'] || '').split(',').map! { |e| e.strip }
-  sampling = !ENV.key?("RBS_TEST_NO_SAMPLE")
   RBS.logger_level = (ENV["RBS_TEST_LOGLEVEL"] || "info")
-rescue
+  sample_size = get_sample_size(ENV['RBS_TEST_SAMPLE_SIZE'] || '')
+rescue InvalidSampleSizeError => exception
+  RBS.logger.error exception.message
+  exit 1
+rescue Exception => e
+  raise e.message
   STDERR.puts "rbs/test/setup handles the following environment variables:"
   STDERR.puts "  [REQUIRED] RBS_TEST_TARGET: test target class name, `Foo::Bar,Foo::Baz` for each class or `Foo::*` for all classes under `Foo`"
   STDERR.puts "  [OPTIONAL] RBS_TEST_SKIP: skip testing classes"
   STDERR.puts "  [OPTIONAL] RBS_TEST_OPT: options for signatures (`-r` for libraries or `-I` for signatures)"
   STDERR.puts "  [OPTIONAL] RBS_TEST_LOGLEVEL: one of debug|info|warn|error|fatal (defaults to info)"
-  STDERR.puts "  [OPTIONAL] RBS_TEST_NO_SAMPLE: if set, the type checker tests all the values of a collection"
+  STDERR.puts "  [OPTIONAL] RBS_TEST_SAMPLE_SIZE: sets the amount of values in a collection to be type-checked (Set to `ALL` to type check all the values)"
   exit 1
 end
 
@@ -32,7 +37,8 @@ env = RBS::Environment.from_loader(loader).resolve_type_names
 
 def match(filter, name)
   if filter.end_with?("*")
-    name.start_with?(filter[0, filter.size - 1]) || name == filter[0, filter.size-3]
+    size = filter.size
+    name.start_with?(filter[0, size - 1]) || name == filter[0, size-3]
   else
     filter == name
   end
@@ -52,7 +58,7 @@ TracePoint.trace :end do |tp|
       if tester.checkers.none? {|hook| hook.klass == tp.self }
         if env.class_decls.key?(class_name)
           logger.info "Setting up hooks for #{class_name}"
-          tester.install!(tp.self, sampling: sampling)
+          tester.install!(tp.self, sample_size: sample_size)
         end
       end
     end
--- a/lib/rbs/test/setup_helper.rb
+++ b/lib/rbs/test/setup_helper.rb
@@ -1,0 +1,29 @@
+module RBS
+  module Test
+    module SetupHelper
+      class InvalidSampleSizeError < StandardError
+        attr_reader :string
+        
+        def initialize(string)
+          @string = string
+          super("Sample size should be a positive integer: `#{string}`")
+        end
+      end
+      
+      DEFAULT_SAMPLE_SIZE = 100
+      
+      def get_sample_size(string)
+        case string
+        when ""
+          DEFAULT_SAMPLE_SIZE
+        when 'ALL'
+          nil
+        else
+          int_size = string.to_i
+          raise InvalidSampleSizeError.new(string) unless int_size.positive?
+          int_size
+        end
+      end
+    end
+  end
+end
--- a/lib/rbs/test/tester.rb
+++ b/lib/rbs/test/tester.rb
@@ -17,14 +17,14 @@ module RBS
         @builder ||= DefinitionBuilder.new(env: env)
       end
 
-      def install!(klass, sampling:)
+      def install!(klass, sample_size:)
         RBS.logger.info { "Installing runtime type checker in #{klass}..." }
 
         type_name = factory.type_name(klass.name).absolute!
 
         builder.build_instance(type_name).tap do |definition|
           instance_key = new_key(type_name, "InstanceChecker")
-          Observer.register(instance_key, MethodCallTester.new(klass, builder, definition, kind: :instance, sampling: sampling))
+          Observer.register(instance_key, MethodCallTester.new(klass, builder, definition, kind: :instance, sample_size: sample_size))
 
           definition.methods.each do |name, method|
             if method.implemented_in == type_name
@@ -36,7 +36,7 @@ module RBS
 
         builder.build_singleton(type_name).tap do |definition|
           singleton_key = new_key(type_name, "SingletonChecker")
-          Observer.register(singleton_key, MethodCallTester.new(klass.singleton_class, builder, definition, kind: :singleton, sampling: sampling))
+          Observer.register(singleton_key, MethodCallTester.new(klass.singleton_class, builder, definition, kind: :singleton, sample_size: sample_size))
 
           definition.methods.each do |name, method|
             if method.implemented_in == type_name || name == :new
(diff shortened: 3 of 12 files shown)


soutaro:
@Afront Thanks, it's a good idea!

Let's clarify the API first. My understanding is as follows:

| `$RBS_TEST_SAMPLE_SIZE` | sampling |
|-------|-----------|
| `"ALL"` | Disable sampling (pass `nil` to `tester.install!`) |
| `/\d+/` | Sampling with the number (pass `#.to_i` to `tester.install!`) |
| not set | Sampling with default size (pass `100` to `tester.install!`) |
| Anything else | Error! |

Is it okay?

Afront:
Yes, it's correct. I also had an option to explicitly set the sample size to the default value (via `"DEFAULT"`), but that can be removed since not setting a value already sets the sample size to the default value, which is `100`.

soutaro:
Okay.

I don't think passing `"DEFAULT"` to `install!` makes sense. It's better to keep the default value only in `setup.rb` (because we don't need to know if it is the default value in type checker), and pass an integer to `install!`.

Afront:
Okay, I understand. I forgot to note another thing to the API. Passing a value that is not a `String`, `Integer`, or a `Float` object would result in an error since the value of an environment variable is always a `String` (or nil) and some people might pass a `Float` or `Integer` to `install!` as well, but passing a non-numeric string or a negative number would not result in an error, and it would just warn the user that the value is invalid and then use the default value, which is `100`.

soutaro:
If a value is specified bu. it's not an integer nor ’all’ it should be an error and just exit with an error message. Validating with a regexp would work.

soutaro:
Ah we should try with #to_i and detect an error. (Negative or zero.)

lib/rbs/test/setup.rb
```diff
@@ -1,24 +1,29 @@
...
+  exit 1
+rescue Exception => e
+  raise e.message
   STDERR.puts "rbs/test/setup handles the following environment variables:"
   STDERR.puts "  [REQUIRED] RBS_TEST_TARGET: test target class name, `Foo::Bar,Foo::Baz` for each class or `Foo::*` for all classes under `Foo`"
   STDERR.puts "  [OPTIONAL] RBS_TEST_SKIP: skip testing classes"
   STDERR.puts "  [OPTIONAL] RBS_TEST_OPT: options for signatures (`-r` for libraries or `-I` for signatures)"
   STDERR.puts "  [OPTIONAL] RBS_TEST_LOGLEVEL: one of debug|info|warn|error|fatal (defaults to info)"
-  STDERR.puts "  [OPTIONAL] RBS_TEST_NO_SAMPLE: if set, the type checker tests all the values of a collection"
+  STDERR.puts "  [OPTIONAL] RBS_TEST_SAMPLE_SIZE: sets the amount of values in a collection to be type-checked (Set to `ALL` to type check all the values or to `DEFAULT` to set to the default value)"
```
soutaro, line 26: No `Default`, right?

lib/rbs/test/tester.rb
```diff
@@ -17,14 +17,14 @@ def builder
         @builder ||= DefinitionBuilder.new(env: env)
       end
 
-      def install!(klass, sampling:)
+      def install!(klass, sample_size: nil)
```
soutaro, line 20: Make the `sample_size` a required keyword.

lib/rbs/test/tester.rb
```diff
@@ -66,22 +66,22 @@ class MethodCallTester
         attr_reader :definition
         attr_reader :builder
         attr_reader :kind
-        attr_reader :sampling
+        attr_reader :sample_size
 
-        def initialize(self_class, builder, definition, kind:, sampling:)
+        def initialize(self_class, builder, definition, kind:, sample_size: nil)
```
soutaro, line 71: Make `sample_size` a required keyword here too.

lib/rbs/test/type_check.rb
```diff
@@ -3,11 +3,14 @@ module Test
     class TypeCheck
       attr_reader :self_class
       attr_reader :builder
+      attr_reader :sample_size
 
-      def initialize(self_class:, builder:, sampling:)
+      DEFAULT_SAMPLE_SIZE = 100
+
+      def initialize(self_class:, builder:, sample_size: nil)
```
soutaro, line 10: Same here about optional keyword.

lib/rbs/test/type_check.rb
```diff
@@ -176,12 +179,8 @@ def zip_args(args, fun, &block)
         end
       end
 
-      def sampling?
-        !!@sampling
-      end
-
       def sample(array)
-        array.size > 100 && sampling? ? array.sample(100) : array
+        @sample_size && (array.size > @sample_size) ? array.sample(@sample_size) : array
```
soutaro, line 183: Use `sample_size` method instead of instance variable.

test/rbs/test/setup_test.rb
```diff
@@ -0,0 +1,29 @@
+require "test_helper"
+require "rbs/test"
+
+class SetupTest < Minitest::Test
```
soutaro, line 4: The test name should be `SetupHelperTest`.

test/rbs/test/setup_test.rb
```diff
@@ -0,0 +1,29 @@
...
+include RBS::Test::SetupHelper
```
soutaro, line 5: This is a nitpick but you need two spaces before `include`.

test/rbs/test/tester_test.rb
```diff
@@ -24,7 +24,7 @@ def move: (?x: Integer, ?y: Integer) -> void
         builder = RBS::DefinitionBuilder.new(env: env)
         definition = builder.build_instance(type_name("::Hello"))
 
-        checker = RBS::Test::Tester::MethodCallTester.new(Object, builder, definition, kind: :instance, sampling: true)
+        checker = RBS::Test::Tester::MethodCallTester.new(Object, builder, definition, kind: :instance, sample_size: 'DEFAULT')
```
soutaro, line 27: `sample_size: nil`

test/rbs/test/type_check_test.rb
```diff
@@ -120,19 +119,68 @@ def test_sampling_handling
     SignatureManager.new do |manager|
       manager.build do |env|
         builder = DefinitionBuilder.new(env: env)
+        
+        begin assert '#sample'
```
soutaro, line 123: These `begin`s are useless.

soutaro: You can clean up the tests because you need only two cases, `nil` and an Integer.

test/test_helper.rb
```diff
@@ -150,6 +156,26 @@ def assert_write(decls, string)
...
   end
+
+  def assert_raises_with_message(error_type, expected_message)
+    silence_errors do
+      e = assert_raises(error_type) { yield }
+      assert_match expected_message, e.message
+    end
+  end
+
+  def assert_raises_from_sampling_check(builder, sample_size)
```
soutaro, line 167: I don't think you need these helpers. We can skip testing these errors.

test/test_helper.rb
```diff
@@ -150,6 +156,26 @@ def assert_write(decls, string)
...
+    assert_raises_with_message(ArgumentError, /comparison of .+ with .+ failed/) do
+      sampling_check = RBS::Test::TypeCheck.new(self_class: Integer, builder: builder, sample_size: sample_size)
+      sampling_check.sample([0,1,2,3,4])
+    end
+  end
+
+  def assert_sampling_check(builder, sample_size, obj_to_be_sampled)
```
soutaro, line 174: I'm not sure if we need this helper...

test/test_helper.rb
```diff
@@ -150,6 +156,26 @@ def assert_write(decls, string)
...
+    assert sampling_check = RBS::Test::TypeCheck.new(self_class: Integer, builder: builder, sample_size: sample_size)
```
soutaro, line 175: This `assert` is meaningless. What do you want to test with these assertions?

test/rbs/test/runtime_test_test.rb
```diff
@@ -1,14 +1,13 @@
 require "test_helper"
-
 require "rbs/test"
 require "logger"
 
 return unless Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7.0')
 
-class RBS::Test::RuntimeTestTest < Minitest::Test
+module TestSetupHelper
```
soutaro, line 7: I don't think we need to extract this module. Define the helper methods in `RuntimeTestTest`.

test/rbs/test/runtime_test_test.rb
```diff
@@ -1,14 +1,13 @@
...
   include TestHelper
 
-  def test_runtime_test
+  def do_runtime_session(other_env: {})
```
soutaro, line 10: I think we can organize the test helpers as follows:

* `def run_runtime_test(other_env:)` to run the test and returns `[err, status]`
* `def assert_test_success(other_env = {})` which asserts `assert_operator status, :success?` and returns stderr output.
* `def assert_test_failure(other_env = {})` which asserts `refute_operator status, :success?` and returns stderr output.

Using the three helpers, we can define the tests like:

```
def test_runtime_test_success
  assert_test_success()
  assert_test_success("RBS_TEST_SAMPLE_SIZE" => "30")
  assert_test_success("RBS_TEST_SAMPLE_SIZE" => "ALL")
end

def test_runtime_test_error_with_mulformed_sample_size
  error_message = assert_test_failure("RBS_TEST_SAMPLE_SIZE" => "yes")
  assert_match /.../, error_message

  ...
end
```

lib/rbs/test/setup_helper.rb
```diff
@@ -0,0 +1,28 @@
+module RBS
```
soutaro, line 1: 👍

lib/rbs/test/setup.rb
```diff
@@ -37,7 +37,8 @@
 
 def match(filter, name)
   if filter.end_with?("*")
-    name.start_with?(filter[0, filter.size - 1]) || name == filter[0, filter.size-3]
+    size = filter.size
```
soutaro, line 40: 👍

test/rbs/test/runtime_test_test.rb
```diff
@@ -37,24 +54,32 @@ def move(x: 0, y: 0)
...
-        refute_operator status, :success?
-        assert_match(/Setting up hooks for ::Hello$/, err)
-        assert_match(/TypeError: \[Hello#move\] ArgumentError:/, err)
+        return [err, status]
       end
     end
   end
+
+  def assert_test_success(other_env: {})
+    silence_warnings do
```
soutaro, line 72: I don't think we need `silence_warnings` here.

soutaro:
The rest of the code is okay. 💪

test/rbs/test/runtime_test_test.rb
```diff
@@ -37,24 +54,28 @@ def move(x: 0, y: 0)
...
-        refute_operator status, :success?
-        assert_match(/Setting up hooks for ::Hello$/, err)
-        assert_match(/TypeError: \[Hello#move\] ArgumentError:/, err)
+        return [err, status]
       end
     end
   end
+
+  def assert_test_success(other_env: {})
+      result = run_runtime_test(other_env: other_env)
```
soutaro, line 72: Please fix the indentation. 🙏

PR #342, opened 2020-07-21
Introduce `...` syntax

soutaro:
This PR contains an update of _syntax_. We are introducing `...` method type syntax and deprecating `overload def` syntax.

```
def foo: () -> String | ...           # New syntax
overload def foo: () -> String        # Deprecated
```


soutaro:
No semantics update implemented yet while we are planning to change the semantics to allow _overloading_ super class's method types.

PR #346, opened 2020-07-28
Add PTY module

hargoniX:
I do have a test in my local repository, however I'm not quite sure whether I should add it since it has to spawn a shell as well as (obviously) a PTY in order to run it's methods properly. However that would require us to choose a shell for the tests, for Linux and UNIX based Operating Systems that'd be `sh` I presume? On windows cmd?

If we don't want to mess with such details I can alternatively delete the tests as well, the type signatures aren't all too complex.

### DIFF
--- a/stdlib/pty/pty.rbs
+++ b/stdlib/pty/pty.rbs
@@ -1,0 +1,15 @@
+module PTY
+  def self.check: (*untyped) -> untyped
+
+  def self.getpty: (*untyped) -> untyped
+
+  def self.open: () -> untyped
+
+  def self.spawn: (*untyped) -> untyped
+
+  private
+
+  def getpty: (*untyped) -> untyped
+
+  def spawn: (*untyped) -> untyped
+end
--- a/stdlib/pty/pty.rbs
+++ b/stdlib/pty/pty.rbs
@@ -1,15 +1,159 @@
+# Creates and manages pseudo terminals (PTYs).  See also
+# http://en.wikipedia.org/wiki/Pseudo_terminal
+#
+# PTY allows you to allocate new terminals using ::open or ::spawn a new
+# terminal with a specific command.
+#
+# ## Example
+#
+# In this example we will change the buffering type in the `factor` command,
+# assuming that factor uses stdio for stdout buffering.
+#
+# If IO.pipe is used instead of PTY.open, this code deadlocks because factor's
+# stdout is fully buffered.
+#
+#     # start by requiring the standard library PTY
+#     require 'pty'
+#
+#     master, slave = PTY.open
+#     read, write = IO.pipe
+#     pid = spawn("factor", :in=>read, :out=>slave)
+#     read.close     # we dont need the read
+#     slave.close    # or the slave
+#
+#     # pipe "42" to the factor command
+#     write.puts "42"
+#     # output the response from factor
+#     p master.gets #=> "42: 2 3 7\n"
+#
+#     # pipe "144" to factor and print out the response
+#     write.puts "144"
+#     p master.gets #=> "144: 2 2 2 2 3 3\n"
+#     write.close # close the pipe
+#
+#     # The result of read operation when pty slave is closed is platform
+#     # dependent.
+#     ret = begin
+#             master.gets     # FreeBSD returns nil.
+#           rescue Errno::EIO # GNU/Linux raises EIO.
+#             nil
+#           end
+#     p ret #=> nil
+#
+# ## License
+#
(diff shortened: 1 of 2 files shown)


soutaro:
Thank you @hargoniX !

Could you try running `bin/annotate-with-rdoc stdlib/pty/pty.rbs` to import RDoc as markdown. (And delete the `arglists` section.)

stdlib/pty/pty.rbs
```diff
@@ -0,0 +1,158 @@
...
+  # In the non-block form this returns an array of size three,
+  # <tt>[r, w, pid]</tt>.
+  #
+  # In the block form these same values will be yielded to the block:
+  #
+  # +r+:: A readable IO that contains the command's
+  #       standard output and standard error
+  # +w+:: A writable IO that is the command's standard input
+  # +pid+:: The process identifier for the command.
+  def self.getpty: (*String command) -> [IO, IO, Integer]
```
soutaro, line 90: Could you add block accepting patterns?

stdlib/pty/pty.rbs
```diff
@@ -0,0 +1,158 @@
...
+  #		    terminal device is available via +slave_file.path+
+  #
+  # IO#raw! is usable to disable newline conversions:
+  #
+  #   require 'io/console'
+  #   PTY.open {|m, s|
+  #     s.raw!
+  #     ...
+  #   }
+  def self.open: () -> [IO, File]
```
soutaro, line 129: Seems like `open` also accepts block.

stdlib/pty/pty.rbs
```diff
@@ -0,0 +1,158 @@
...
+  # In the non-block form this returns an array of size three,
+  # <tt>[r, w, pid]</tt>.
+  #
+  # In the block form these same values will be yielded to the block:
+  #
+  # +r+:: A readable IO that contains the command's
+  #       standard output and standard error
+  # +w+:: A writable IO that is the command's standard input
+  # +pid+:: The process identifier for the command.
+  def self.spawn: (*String command) -> [IO, IO, Integer]
```
soutaro, line 157: This also accepts block!

soutaro:
For testing, maybe we can use `ruby` itself? (https://github.com/ruby/ruby/blob/master/test/test_pty.rb)

hargoniX:
Regarding the rdoc part, if i run that script it will just clear all docs for me...I suspect that this is due to the fact that pty is a C module? At least that's where i got the docs that are above the functions right now from.

For the rest, I'll get to that!

soutaro:
Okay. I think you installed ruby without docs. (RVM doesn't build RDoc by default. https://rvm.io/rubies/docs)

After you generated RDoc, `$ ri PTY` will show you a doc.

hargoniX:
So I did add my tests, however the block testing confuses me slightly, for example in `test_getpty` I use
```ruby
assert_send_type  "(*::String command) { (::IO r, ::IO w, ::Integer pid) -> untyped } -> untyped",
                   PTY, :getpty, "echo" do |r, w, pid| 1 end
```
Which is the proper type to my understanding of the documentation (https://ruby-doc.org/stdlib-2.7.0/libdoc/pty/rdoc/PTY.html#method-c-getpty) However this assertion tells me that that's apparently not at all what is happening, instead it says I'm getting some sort of array type? So I'm wondering, is the true signature for this method actually
```
(*::String command) { (Array[::IO | ::Integer] args) -> untyped } -> untyped
```
instead of having all three as separate arguments? Or am I missing out on something?

soutaro:
@hargoniX I think the document is incorrect. The implementation passes an array of three objects, not three arguments.

Could you try the type `(*String command) { ([IO, IO, Integer]) -> untyped } -> untyped`? (We may need to fix the doc or the implementation.)

hargoniX:
@soutaro That did indeed do the trick. However looking at the signatures the docs provide I believe this behaviour is kind of predictable since the docs state
```
spawn(command_line) { |r, w, pid| ... }
spawn(command_line) => [r, w, pid]
spawn(command, arguments, ...) { |r, w, pid| ... }
spawn(command, arguments, ...) => [r, w, pid]
```
and as we can see from the 2nd and 4th way to call the method they do return an array so the `|r, w, pid|` syntax they show just unpacks the array already and the docs are fine?

soutaro:
When single array is yielded and the block has multiple parameters, Ruby automatically unpacks the array. So, the examples in the doc works.

We can see the difference when we receive it with single block parameter.

```rb
PTY.spawn("ls") {|x| pp x }
```

hargoniX:
@soutaro Cleaned the history up!

PR #360, opened 2020-08-08
Remove unused constants

pocke:
They are no longer used since 2a5ab09eb3d5ad40d993462e08e3a82a1b6689c3


soutaro:
I think it's an issue of runtime type checker...

PR #361, opened 2020-08-09
CMath

moritzbuhl:
I am not sure about the location of the `cmath.rbs` file.

Tests are not passing on my machine yet but I don't know what is needed to make them pass:

```
moritz@pink rbs % ruby bin/test_runner.rb test/stdlib/CMath_test.rb
	9: from bin/test_runner.rb:12:in `<main>'
	8: from bin/test_runner.rb:12:in `each'
	7: from bin/test_runner.rb:13:in `block in <main>'
	6: from bin/test_runner.rb:13:in `load'
	5: from test/stdlib/CMath_test.rb:4:in `<top (required)>'
	4: from test/stdlib/CMath_test.rb:8:in `<class:CMathTest>'
	3: from /Users/moritz/Desktop/rbs/test/stdlib/test_helper.rb:155:in `testing'
	2: from /Users/moritz/.rvm/gems/ruby-2.7.0/gems/rbs-0.9.1/lib/rbs/definition_builder.rb:397:in `build_instance'
	1: from /Users/moritz/.rvm/gems/ruby-2.7.0/gems/rbs-0.9.1/lib/rbs/definition_builder.rb:1021:in `try_cache'
/Users/moritz/.rvm/gems/ruby-2.7.0/gems/rbs-0.9.1/lib/rbs/definition_builder.rb:398:in `block in build_instance': Unknown name for build_instance: ::CMath (RuntimeError)
```

### DIFF
--- a/stdlib/cmath/cmath.rbs
+++ b/stdlib/cmath/cmath.rbs
@@ -3,121 +3,42 @@ module CMath
 
   RealMath: untyped
 
-  #
-  # Math::E raised to the +z+ power
-  #
-  #   CMath.exp(1.i * Math::PI) #=> (-1.0+1.2246467991473532e-16i)
   def exp: (untyped z) -> untyped
 
-  #
-  # Returns the natural logarithm of Complex. If a second argument is given,
-  # it will be the base of logarithm.
-  #
-  #   CMath.log(1 + 4i)     #=> (1.416606672028108+1.3258176636680326i)
-  #   CMath.log(1 + 4i, 10) #=> (0.6152244606891369+0.5757952953408879i)
   def log: (untyped z, ?untyped b) -> untyped
 
-  #
-  # Returns the base 2 logarithm of +z+
-  #
-  #   CMath.log2(-1) => (0.0+4.532360141827194i)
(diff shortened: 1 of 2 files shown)


soutaro:
Thank you @moritzbuhl for opening this PR. It's complicated...

1. `cmath` is no longer bundled with `ruby` since 2.7. So, we don't think bundling the `.rbs`s for cmath in `rbs` is the best way.
2. The best way should be bundling `.rbs`s with `cmath` gem, but it's not very clear yet how can we do it.

So, I think I should take care of these complications myself. (Do you want RBSs for cmath for an application?)

moritzbuhl:
>(Do you want RBSs for cmath for an application?)

No, I just picked it at random from that list because I already hate libm.

Should I link my current efforts to the [CMath repository](https://github.com/ruby/cmath) in some way?

soutaro:
@moritzbuhl Ok, thanks.

It seems like a bit complicated. (I need to define a way to test the RBS in CMath repository, for example.)
Could you let me continue this work on CMath repository? I will find out the best way to give RBS for CMath, and take care of this.

PR #363, opened 2020-08-10
Debgug print  “No type checker” message

soutaro:
The `at_exit` block may be called twice for `RUBYOPT="-rrbs/test/setup" rake test`, one for unit tests and another for `rake`. This is confusing.

This PR is to make it `debug` print to stop printing it by default.


soutaro:
When you are using `rake`, you can move the `RUBYOPT` definition after `rake` commandline to avoid this issue.

```
$ rake test RUBYOPT="-rrbs/test/setup"
```

PR #366, opened 2020-08-14
Add `rbs test` command

Afront:
This PR adds the `rbs test` command to RBS, which allows the users to easily run the runtime type checker instead of manually setting the environment variables every time like before.

The API is as follows:
 rbs [RBS OPTIONS] test [TEST OPTIONS] COMMAND

`RBS OPTIONS` are the options that RBS currently uses, such as `-I DIR` and `-r LIB`

TEST OPTIONS:
| Command | Description|
| :-: | :-: |
| --target \<TARGET\> |  the module, class, or anything else that would be tested
|   --sample-size \<SIZE\>\|ALL  |   the amount of values in a collection to be type-checked

### DIFF
--- a/lib/rbs/cli.rb
+++ b/lib/rbs/cli.rb
@@ -1,4 +1,5 @@
 require "optparse"
+require "shellwords"
 
 module RBS
   class CLI
@@ -36,7 +37,7 @@ module RBS
       @stderr = stderr
     end
 
-    COMMANDS = [:ast, :list, :ancestors, :methods, :method, :validate, :constant, :paths, :prototype, :vendor, :parse]
+    COMMANDS = [:ast, :list, :ancestors, :methods, :method, :validate, :constant, :paths, :prototype, :vendor, :parse, :test]
 
     def library_parse(opts, options:)
       opts.on("-r LIBRARY", "Load RBS files of the library") do |lib|
@@ -749,6 +750,53 @@ Examples:
         last = namespace.path.last
         TypeName.new(name: last, namespace: namespace.parent)
       end
+    end
+
+    def test_opt options
+      opt_string = options.dirs.map { |dir| "-I #{Shellwords.escape(dir)}"}.concat(options.libs.map { |lib| "-r#{Shellwords.escape(lib)}"}).join(' ')
+      opt_string.empty? ? nil : opt_string
+    end
+
+    def run_test(args, options)
+      targets = []
+      sample_size = nil
+
+      (opts = OptionParser.new do |opts|
+        opts.banner = <<EOB
+Usage: rbs [rbs options...] test [test options...] COMMAND
+
+Examples:
+
+  $ rbs test rake test
+  $ rbs --log-level=debug test --target SomeModule::* rspec
+  $ rbs test --target SomeModule::* --target AnotherModule::* --target SomeClass rake test
+
+Options:
+EOB
+        opts.on("--target TARGET", "Sets the runtime test target") do |target|
+          targets << target
+        end
+
+        opts.on("--sample-size SAMPLE_SIZE", "Sets the sample size") do |size|
+          sample_size = size
+        end
+      end).order!(args)
+
+      if args.length.zero?
+        stdout.puts opts.help
+        exit 1
+      end
+
+      env_hash = {
+        'RBS_TEST_OPT' => test_opt(options),
+        'RBS_TEST_TARGET' => (targets.join(',') unless targets.empty?),
+        'RBS_TEST_SAMPLE_SIZE' => sample_size,
+        'RBS_TEST_LOGLEVEL' => RBS.logger_level,
+        'RUBYOPT' => "#{ENV['RUBYOPT']} -rrbs/test/setup"
+      }
+
+      system(env_hash, *args)
+      $?
     end
   end
 end
--- a/test/rbs/cli_test.rb
+++ b/test/rbs/cli_test.rb
@@ -1,6 +1,5 @@
 require "test_helper"
 require "stringio"
-
 require "rbs/cli"
 
 class RBS::CliTest < Minitest::Test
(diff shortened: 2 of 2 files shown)


lib/rbs/cli.rb
```diff
@@ -750,5 +751,55 @@ def parse_type_name(string)
...
+Examples:
+
+  $ rbs test rake test
+  $ rbs --log-level=debug test --target SomeModule::* rspec
+  $ rbs test --target SomeModule::* --target AnotherModule::* --target SomeClass rake test
+
+Options:
+EOB
+        opts.on("--target TARGET", "Sets the runtime test target") do |target|
+          targets << Shellwords.escape(target)
```
soutaro, line 778: The escape for `target` is unnecessary.

lib/rbs/cli.rb
```diff
@@ -750,5 +751,55 @@ def parse_type_name(string)
...
+      if args.length.zero?
+        stdout.puts help
+        exit 1
+      end
+
+      targets_string = targets.uniq.join(',')
+
+      env_hash = {
+        'RBS_TEST_OPT' => test_opt(options),
+        'RBS_TEST_TARGET' => (targets_string unless targets_string.empty?),
```
soutaro, line 797: I would write `targets.join(",") unless targets.empty?` here.

* It works well even without `uniq`.
* Testing `targets` emptiness seems better because we can omit `targets_string` variable completely.

lib/rbs/cli.rb
```diff
@@ -750,5 +751,55 @@ def parse_type_name(string)
...
+EOB
+        opts.on("--target TARGET", "Sets the runtime test target") do |target|
+          targets << Shellwords.escape(target)
+        end
+
+        opts.on("--sample-size SAMPLE_SIZE", "Sets the sample size") do |size|
+          sample_size = size
+        end
+
+        help = opts.help
```
soutaro, line 785: You can assign the OptionParser instance and use it to print the help message.

```
opts = OptionParser.new do ... end

if args.empty?
  stdout.puts opts.help
  exit 1
end
```

lib/rbs/cli.rb
```diff
@@ -750,5 +751,55 @@ def parse_type_name(string)
         TypeName.new(name: last, namespace: namespace.parent)
       end
     end
+
+    def test_opt options
+      opt_string = options.dirs.map { |dir| "-I #{dir}"}.concat(options.libs.map { |lib| "-r{lib}"}).join(' ')
```
soutaro, line 756: Escape the options before `join`. `Shellescape.join` would be the best method for our purpose!

lib/rbs/cli.rb
```diff
@@ -750,5 +751,50 @@ def parse_type_name(string)
         TypeName.new(name: last, namespace: namespace.parent)
       end
     end
+
+    def test_opt options
+      opt_string = options.dirs.map { |dir| Shellwords.escape("-I #{dir}")}.concat(options.libs.map { |lib| Shellwords.escape("-r{lib}")}).join(' ')
+      opt_string.empty? ? nil : Shellwords.escape(opt_string)
```
soutaro, line 757: I don't think we need `Shellwords.escape(opt_string)` again here. Do we need it?

lib/rbs/cli.rb
```diff
@@ -750,5 +751,50 @@ def parse_type_name(string)
...
+      end).order!(args)
+
+      if args.length.zero?
+        stdout.puts opts.help
+        exit 1
+      end
+
+      env_hash = {
+        'RBS_TEST_OPT' => test_opt(options),
+        'RBS_TEST_TARGET' => (targets.join('') unless targets.empty?),
```
soutaro, line 792: `.join(',')` ?

test/rbs/cli_test.rb
```diff
@@ -266,4 +266,33 @@ def cli.has_parser?
...
+          end
+        end
+      RBS
+
+      with_cli do |cli|
+        assert_raises(SystemExit) { cli.run(%w(test)) }
+        assert_raises(SystemExit) { cli.run(%W(-I #{dir} test)) }
+        assert_raises(SystemExit) { cli.run(%W(-I #{dir} test --target ::Foo)) }
+        
+        # assert_instance_of Integer, cli.run(%W(-I #{dir} test --target ::Foo ls))
```
soutaro, line 290: I want there are some test cases here...

test/rbs/cli_test.rb
```diff
@@ -1,6 +1,6 @@
 require "test_helper"
 require "stringio"
-
+require "thread"
```
soutaro, line 3: No need to `require` thread explicitly??

test/rbs/cli_test.rb
```diff
@@ -266,4 +266,39 @@ def cli.has_parser?
...
+        assert_rbs_test_no_errors(cli, dir, %w(--target Bar::Baz rbs version))
+        assert_rbs_test_no_errors(cli, dir, %w(--target ::Bar::Baz rake stdlib_test))
+        assert_rbs_test_no_errors(cli, dir, %w(--target ::Foo --target Bar::* rbs test --target ::Bar::Baz rbs version))
+      end
+    end
+  end
+
+  def assert_rbs_test_no_errors cli, dir, arg_array
+    args = ['-I', dir.to_s, 'test', *arg_array] 
+    assert_instance_of Integer, Thread.new { cli.run(args) }.value
```
soutaro, line 302: I don't think we need to use thread here.

lib/rbs/cli.rb
```diff
@@ -750,5 +751,50 @@ def parse_type_name(string)
...
+      end
+
+      env_hash = {
+        'RBS_TEST_OPT' => test_opt(options),
+        'RBS_TEST_TARGET' => (targets.join(',') unless targets.empty?),
+        'RBS_TEST_SAMPLE_SIZE' => sample_size,
+        'RBS_TEST_LOGLEVEL' => RBS.logger_level
+      }
+
+      Process.wait Process.spawn(env_hash, args.join(' '))
```
soutaro, line 797: `Kernel#system` would be better.

```ruby
system(env_hash, *args)
$?
```

soutaro (approved):
Looks good to me. 💯 

Could you try to fix up the commits? I would squash to one commit.

PR #367, opened 2020-08-16
Add date/datetime library types

waka:
Add signatures of Date class and DateTime class.
This PR is based on [the contributon guide](https://github.com/ruby/rbs/blob/master/docs/CONTRIBUTING.md).

### DIFF
--- a/stdlib/date/date.rbs
+++ b/stdlib/date/date.rbs
@@ -136,7 +136,27 @@
 class Date
   include Comparable
 
-  def initialize: (*untyped) -> untyped
+  # Creates a date object denoting the given calendar date.
+  #
+  # In this class, BCE years are counted astronomically.  Thus, the year before
+  # the year 1 is the year zero, and the year preceding the year zero is the year
+  # -1.  The month and the day of month should be a negative or a positive number
+  # (as a relative month/day from the end of year/month when negative).  They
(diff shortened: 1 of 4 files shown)


stdlib/date/date.rbs
```diff
@@ -0,0 +1,1057 @@
...
+  # Date::JULIAN (the proleptic Julian calendar) can be specified as a day of
+  # calendar reform.
+  #
+  #     Date.new(2001)            #=> #<Date: 2001-01-01 ...>
+  #     Date.new(2001,2,3)        #=> #<Date: 2001-02-03 ...>
+  #     Date.new(2001,2,-1)       #=> #<Date: 2001-02-28 ...>
+  #
+  # See also ::jd.
+  #
+  def self.new: (?Integer year, ?Integer month, ?Integer mday, ?Integer start) -> Date
```
soutaro, line 157: Could you delete the `.new` method definition? It's automatically generated from `#initialize` type. (Same for `DateTime.new` too.)

(And the return type of `initialize` can be `void`.)

waka: Oh, it's convenient. I understand and deleted it (date and datetime).

soutaro (approved):
Thanks! 🎉

PR #369, opened 2020-08-19
_parameter_ typo fix

sbeckeriv:
Dearest Reviewer,

This is a a typo fix.  While I am here I have a question about _parameter_. Why would a parameter include a var-name if it was not in the name parameter format `(x: Int)`? 

```
  def bar: (untyped `type`, void: untyped `void`) -> untyped
```
In this test what does `type` do? https://github.com/ruby/rbs/blob/master/test/rbs/signature_parsing_test.rb#L786

I am mostly reading the syntax document. Is there another document I should be reading for reference? 

Thanks for your time,

Dictated but not reviewed,
Becker

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -212,7 +212,7 @@ _method-type_ ::= `(` _parameters_ `) ->` _type_                                
 
 _parameters_ ::= _required-positionals_ _optional-positionals_ _rest-positional_ _trailing-positionals_ _keywords_
 
-_paramater_ ::= _type_ _var-name_                                  # Parameter with var name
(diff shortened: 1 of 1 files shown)


soutaro:
Thank you @sbeckeriv !

   def foo: (String name, email: String address) -> void

Let's think about the example above. The `name` is the name of parameter and `email` is the name of keyword. I think your question is about `address`. It's just because I reuse the syntax rule for parameter in keyword type too. 

I agree that there is much of usage for the name in keyword arguments, but causing some confusion... Just removing it might make more sense.

PR #372, opened 2020-08-20
[docs] Add comments to sytnax doc

sbeckeriv:
Dearest Reviewer,

I was shocked to see comments in rbs files only because it was not listed in the syntax file. I assumed that post line comments are not allowed. I am unsure if that is a correct statement.  I also assumed that heredocs/docstring/<<- and =begin/=end docs are not supported. 

Please let me know if I am incorrect in these assumptions and I will update the diff.

Thank you for work and time on this project,

Dictated but not reviewed,
Becker

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -529,3 +529,16 @@ You can declare a global variable.
 $LOAD_PATH: Array[String]
 ```
 
+### Comments
+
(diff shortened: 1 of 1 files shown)


soutaro (approved):
Thanks! 👏

PR #380, opened 2020-08-25
Add Double API to RBS

Afront:
As of now, there are some issues with doubles for RBS. In order to mitigate those issues, the Double API was made so that users can declare the double suite they are using and ignore unwanted classes.

### DIFF
--- a/lib/rbs/test/setup_helper.rb
+++ b/lib/rbs/test/setup_helper.rb
@@ -24,6 +24,24 @@ module RBS
           int_size
         end
       end
+
+      def to_double_class(double_suite)
+        return nil unless double_suite
+
+        double_class = ENV['RBS_TEST_DOUBLE_CLASS']
+        RBS.logger.warn "Both the double class #{double_class} and double suite (#{double_suite}) are set!" if double_class == double_suite
+
+        case double_suite.downcase.strip
+        when 'rspec'
+          '::RSpec::Mocks::Double'
+        when 'minitest'
+          '::Minitest::Mock'        
+        else
+          RBS.logger.warn "Unknown test suite - defaults to nil"
+          nil
+        end
+      end
+
     end
   end
 end
--- a/lib/rbs/test/setup_helper.rb
+++ b/lib/rbs/test/setup_helper.rb
@@ -28,14 +28,11 @@ module RBS
       def to_double_class(double_suite)
         return nil unless double_suite
 
-        double_class = ENV['RBS_TEST_DOUBLE_CLASS']
-        RBS.logger.warn "Both the double class #{double_class} and double suite (#{double_suite}) are set!" if double_class == double_suite
-
         case double_suite.downcase.strip
         when 'rspec'
-          '::RSpec::Mocks::Double'
+          ['::RSpec::Mocks::Double']
         when 'minitest'
-          '::Minitest::Mock'        
+          ['::Minitest::Mock'] 
         else
           RBS.logger.warn "Unknown test suite - defaults to nil"
           nil
--- a/lib/rbs/test/type_check.rb
+++ b/lib/rbs/test/type_check.rb
@@ -4,15 +4,17 @@ module RBS
       attr_reader :self_class
       attr_reader :builder
       attr_reader :sample_size
+      attr_reader :double_suite
 
       attr_reader :const_cache
 
       DEFAULT_SAMPLE_SIZE = 100
 
-      def initialize(self_class:, builder:, sample_size:)
+      def initialize(self_class:, builder:, sample_size:, double_suite:)
         @self_class = self_class
         @builder = builder
         @sample_size = sample_size
+        @double_suite = double_suite
         @const_cache = {}
       end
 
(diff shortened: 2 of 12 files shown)


lib/rbs/test/setup_helper.rb
```diff
@@ -24,6 +24,24 @@ def get_sample_size(string)
           int_size
         end
       end
+
+      def to_double_class(double_suite)
+        return nil unless double_suite
+
+        double_class = ENV['RBS_TEST_DOUBLE_CLASS']
```
soutaro, line 31: `RBS_TEST_IGNORED_CLASSES` ?

lib/rbs/test/setup_helper.rb
```diff
@@ -24,6 +24,21 @@ def get_sample_size(string)
           int_size
         end
       end
+
+      def to_double_class(double_suite)
+        return nil unless double_suite
+
+        case double_suite.downcase.strip
+        when 'rspec'
+          '::RSpec::Mocks::Double'
```
soutaro, line 33: Returning an array would be better.

soutaro:
It may be better to install `rspec` for testing through `Gemfile`, and running real specs from tests.

lib/rbs/test/type_check.rb
```diff
@@ -203,7 +205,13 @@ def get_class(type_name)
         const_cache[type_name] ||= Object.const_get(type_name.to_s)
       end
 
+      def is_double?(value)
+        unchecked_classes.any? { |unchecked_class| Test.call(value, IS_AP, Object.const_get(unchecked_class))}
+      end
+
       def value(val, type)
+        return true if is_double?(val) && RBS.logger.info("A double (#{val.inspect}) is detected!")
```
soutaro, line 213: It's better to write `info` in `if` block explicitly.

```rb
if is_double?(val)
  RBS.logger.info(...)
  return true
end
```

test/rbs/test/type_check_test.rb
```diff
@@ -357,6 +363,54 @@ def test_typecheck_args
     end
   end
 
+  def test_is_double
```
soutaro, line 366: 👍

soutaro:
Could you add test cases in `runtime_test_test.rb` that confirms unit test (Minitest) and specs (RSpec) are working as we expect.

```rb
def test_minitest
  assert_test_success(..., rbs_content: <<RBS, ruby_content: <<RUBY)
class Foo
  def foo: (Integer) -> void
end
RBS
require "minitest"
require "minitest/mocks"

class FooTest < Minitest::Test
  def test_foo_mock
    # Confirm if mock is correctly ignored.
    Foo.new.foo(::Minitest::Mock.new)
  end

  def test_no_foo
    # Confirm if RBS runtime test raises errors when unexpected object is given.
    assert_raises RBS::Test::Tester::TypeError do
      Foo.new.foo("")
    end
  end
end
RUBY
end
```

test/rbs/test/runtime_test_test.rb
```diff
@@ -126,4 +126,53 @@ def world: () -> void
...
     assert_match /TypeError: \[Hello#world\]/, output
   end
+
+  def test_minitest
+    assert_test_success(rbs_content: <<RBS, ruby_content: <<RUBY)
+class Foo
+  def foo: (Integer) -> void
+end
+RBS
+require "minitest"
```
soutaro, line 136: `require "minitest/test"`

test/rbs/test/runtime_test_test.rb
```diff
@@ -126,4 +126,53 @@ def world: () -> void
 
     assert_match /TypeError: \[Hello#world\]/, output
   end
+
+  def test_minitest
+    assert_test_success(rbs_content: <<RBS, ruby_content: <<RUBY)
```
soutaro, line 131: You need to specify `RBS_TEST_TARGET` and `RBS_TEST_DOUBLE_SUITE` here.

test/rbs/test/runtime_test_test.rb
```diff
@@ -126,4 +126,53 @@ def world: () -> void
...
+    assert_raises RBS::Test::Tester::TypeError do
+      Foo.new.foo("")
+    end
+  end
+end
+RUBY
+  end
+
+  def test_rspec
+    assert_test_success(rbs_content: <<RBS, ruby_content: <<RUBY)
```
soutaro, line 155: I guess the spec is not running here too.

test/rbs/test/runtime_test_test.rb
```diff
@@ -8,18 +8,21 @@ class RBS::Test::RuntimeTestTest < Minitest::Test
   include TestHelper
 
   def test_runtime_success
+    skip
```
soutaro, line 11: Unskip other tests!

soutaro (approved):
@Afront Looks good!

* Revert `skip`s in `runtime_test_test.rb`
* I think the commits can be organized in a better way... But it's okay to merge the 29 commits.

PR #381, opened 2020-08-25
Update post line comment

sbeckeriv:
Dearest Reviewer,

This is the only post line comment I have found so far. Per the syntax file they are not allowed. I have moved it up with the comments above. I think this
comment is more suited to this level of comments. If we were  to show the comment above inline with code the monomorphic detail would be useful.

Thank you for your time and work on this project.
Dictated but not reviewed, 
Becker


soutaro (approved):
Thanks!

(It was intentional because the comment is not included in the original RDoc. But making it in the doc would make sense!)

PR #383, opened 2020-08-29
Fix test command in README

mihyaeru21:
`rake test` didn't work for my environment. Added `bundle exec` to `rake test` according to `.github/workflows/ruby.yml`.


soutaro (approved):
👍

PR #386, opened 2020-08-31
Improve bigdecimal signatures

mihyaeru21:
Improve signatures for `BigDecimal` in `bigdecimal`. And create signatures for `BigMath` and `bigdecimal/math`.

Remove `BigDecimal` from `builtin/builtin.rbs`. Replace `Integer | Float | Rational | BigDecimal` in `builtin/math.rbs` to `Numeric`  to adapt to their implementations. Built in math functions can convert their args from `Numeric` to `Float`.


mihyaeru21:
> RBS::NoTypeFoundError: /__w/rbs/rbs/stdlib/builtin/math.rbs:16:47...16:57: Could not find ::BigDecimal

It seems `BigDecimal` in `builtin/math` should be put into `bigdecimal/math`. `BigDecimal` isn't a built in class.

stdlib/builtin/builtin.rbs
```diff
@@ -35,9 +35,6 @@ interface _Exception
                | (String arg0) -> Exception
 end
 
-class BigDecimal
```
soutaro, line 38: How about keeping the `BigDecimal` definition here? It will make `math.rbs` valid even without `bigdecimal` library. 

```
class BigDecimal < Numeric
  # You can find the definition of BigDecimal in bigdecimal/bigdecimal.rbs.
  # This is a dummy definition for Math module.
end
```

(Another way is to edit `math.rbs` to remove all `BigDecimal` references, and add `Math` methods overloadings in `bigdecimal` directory. I don't think this makes sense.)

mihyaeru21: I replaced `Integer | Float | Rational | BigDecimal` in `math.rbs` to `Numeric` on unpushed commit. [These args are applied `numeric_to_float` function](https://github.com/ruby/ruby/blob/0b81a48/object.c#L3829) in [their implementations](https://github.com/ruby/ruby/blob/0b81a48/math.c#L78). So I think `Numeric` is suitable for these types. How is this?

mihyaeru21:
I will add rbs for remaining sub libraries in `bigdecimal` next PR. Would you review this PR with current changes?

stdlib/builtin/builtin.rbs
soutaro: Good! It makes more sense! 👏

soutaro (approved):

soutaro:
@mihyaeru21 Thanks! I'd like to merge this PR. Opening another PR would be great if you have another proposal.

Feel free to fix up the commits if you want.

mihyaeru21:
Thank you for reviewing! There is nothing I want to fix in this PR any more.

PR #388, opened 2020-09-02
Add type signatures for some URI classes from stdlib

ahazem:
Hi 👋

I just read @soutaro's [article](https://developer.squareup.com/blog/the-state-of-ruby-3-typing/) and thought it would be great to get myself acquainted with typing in Ruby 3 if I actually get to add type signatures to some classes of `stdlib` myself, so here goes:

- [x] Added signatures for `URI::Generic`
- [x] Added signatures for `URI::File`

I'm planning to do more classes (hopefully finish the rest of classes in `URI` module) but I wanted to get some feedback first in case I'm doing something wrong or if there's anything missing.

**Please note:** some methods were left without typing (and tests) because they depend on other `URI::` based classes (like `URI::HTTP` or `URI::Parser`). As soon as we add those, I'll update them with the correct typing.

Cheers! 🥂

### DIFF
--- a/stdlib/uri/file.rbs
+++ b/stdlib/uri/file.rbs
@@ -1,0 +1,167 @@
+# URI is a module providing classes to handle Uniform Resource Identifiers
+# ([RFC2396](http://tools.ietf.org/html/rfc2396)).
+#
+# ## Features
+#
+# *   Uniform way of handling URIs.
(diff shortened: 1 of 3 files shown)


soutaro:
Hello @ahazem !

Thank you for adding RBS for URI! It is so old and big that we couldn't start working on it. 🙏

PR #403, opened 2020-09-23
Fix VarianceCalculator exception when passing too many type arguments

### LINKED ISSUE #401: NoMethodError if too many type arguments passed for a type
If you define a class which takes type arguments, and then pass it too many type arguments, RBS throws a `NoMethodError` from something being nil when using `rbs validate`.

For example, using this RBS:

```
module Example
  class Foo[X]
  end

  def example: (Foo[String, Integer] x) -> void
end
```

Raises this error:

```
NoMethodError: undefined method `variance' for nil:NilClass
  /home/aaron/src/rbs/lib/rbs/variance_calculator.rb:123:in `block in type'
```

Passing too few type arguments (such as by just using `Foo`) will result in a proper `InvalidTypeApplicationError` type error, which is probably what passing too many type arguments should also do.

AaronC81:
Fixes #401.

This uses safe traversal so that running the variance calculator when too many type arguments are passed doesn't throw an exception. The variance of that type parameter instead gets left as `:unused`.

Please let me know if there's anything else I need to add to this PR! Thanks for all your work on RBS, I'm really enjoying getting familiar with it :)

### DIFF
--- a/lib/rbs/variance_calculator.rb
+++ b/lib/rbs/variance_calculator.rb
@@ -120,7 +120,7 @@ module RBS
 
         type.args.each.with_index do |ty, i|
           var = type_params.params[i]
-          case var.variance
+          case var&.variance
           when :invariant
             type(ty, result: result, context: :invariant)
           when :covariant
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thank you so much @AaronC81 !! 🎉

PR #404, opened 2020-09-28
fix: typo in the paths inline documentation

andrewmcodes:
Corrects what appears to be a typo in the `rbs paths` inline documentation.


soutaro (approved):
😱 🎉 🙏

PR #409, opened 2020-10-02
Add more types for URI module

ahazem:
Hi 👋

As promised, to continue on my earlier work from #388, here are a few more classes from the `URI` module:

- [x] Added `URI::HTTP`
- [x] Added `URI::HTTPS`
- [x] Added `URI::LDAP`
- [x] Added `URI::LDAPS`

Also, broke down `URI_test` into multiple files and re-arranged things accordingly.

Please let me know if you have any feedback or comments.

Have a nice weekend! 🥂


stdlib/uri/http.rbs
```diff
@@ -0,0 +1,158 @@
...
+    #
+    # If the URI contains a query, the full path is URI#path + '?' + URI#query.
+    # Otherwise, the path is simply URI#path.
+    #
+    # Example:
+    #
+    #     uri = URI::HTTP.build(path: '/foo/bar', query: 'test=true')
+    #     uri.request_uri #  => "/foo/bar?test=true"
+    #
+    def request_uri: () -> (nil | String)
```
soutaro, line 156: It seems like the `request_uri` method doesn't return `nil`.

soutaro:
@ahazem Thank you for this great work! I have one question about the return type of `Net::HTTP#request_uri` method.

stdlib/uri/http.rbs
ahazem: Good catch! I thought the implicit return here:

https://github.com/ruby/ruby/blob/master/lib/uri/http.rb#L78

would return `nil` but it seems that `@path` is `""` even if it was not passed to `URI::HTTP.build`:

Screenshot 2020-10-04 at 23 51 34

I have updated the signature to reflect that.

soutaro:
Merged! Thanks! 🎉

PR #413, opened 2020-10-13
Commit parser code

soutaro:
It seems like there is no way to `rake parser` when installing from Git repo via bundler.


marcandre:
I find it annoying that we can't (easily) run an install script when a gem is installed from a git repo.

FWIW, the [`parser` gem](https://github.com/whitequark/parser/issues/446) does not support it, and I copied the same setup in [`rubocop-ast`](github.com/rubocop-hq/rubocop-ast/).

soutaro:
Thanks @marcandre! I now feel I'm not doing something badly...

PR #414, opened 2020-10-13
Update README.md

marcandre:
Most Ruby documentation follows the convention `code \n # => result`, not `# result \n code`. I didn't change the README, but at least providing the output seems very useful.


marcandre:
Note: I don't understand why it prints `::string` sometimes, not `::String`. Is it a bug?

soutaro (approved):
👏 🙏

soutaro:
In fact, `::string` is another type, an alias to `::String | ::_ToStr`.

PR #417, opened 2020-10-16
Allow keywords ending with `?` or `!`

soutaro:
The problem is when a method accepts options ending with `?` or `!`.

```rb
def foo(**opts) = opts
foo(a?: 1, b!: 2)        # OK

def bar(a?:, b!:) = ...         # Syntax error
```

So, this PR let keyword names in RBS end with `?` and `!`.


marcandre:
Note that keyword options currently accept just about anything, not just symbols

PR #426, opened 2020-10-16
Fix trailing whitespace

marcandre:
This PR removes trailing whitespaces in the code base.

It tweaks `bin/annotate-with-rdoc` to avoid outputing some.


marcandre:
Note: many programmers like me have their editor set to automatically trim trailing whitespace. Editing codebases with trailing whitespaces is a nightmare.

soutaro (approved):
Exactly... Thanks!

PR #428, opened 2020-10-17
_Reader, _Writer, _Rewindable, and _ToIO interfaces.

### LINKED ISSUE #425: Reader/Writer/Rewindable interfaces
There are a couple of agreed-upon interfaces in stream scenarios, which have been supported by ruby since always, and might break with the recent typing.

For instance, `IO.select` works with any object that responds to `#.to_io`, not with [IO objects, like the typing info suggests](https://github.com/ruby/rbs/blob/c64d1f5ba5a4fbc95ffb1d72d783ee39b17d60fe/stdlib/builtin/io.rbs#L525). Likewise, `IO.copy_stream` works if the first argument responds to `read(size, buffer)`, and the second responds to `write(data) -> size`, and [not only with IO objects](https://github.com/ruby/rbs/blob/c64d1f5ba5a4fbc95ffb1d72d783ee39b17d60fe/stdlib/builtin/io.rbs#L517). Both usages will break under typing if these typing declarations are to be kept.

On the former, I suspect that usage of `OpenSSL::SSLSocket` in `IO.select` will break if under typing.

Alternatively, I'd like to suggest the introduction of  `_Reader`,  `_Writer`,  `_Rewindable`, and `_ToIO` interfaces, which can be used in these functions, but also in others where the same criteria applies.

I've been using and supporting such APIs for a long time and they are very convenient, and have also [written about it](https://github.com/ruby/rbs/blob/c64d1f5ba5a4fbc95ffb1d72d783ee39b17d60fe/stdlib/builtin/io.rbs#L517).

HoneyryderChuck:
These enhance certain methods which expect duck-types of IO, in certain cases methods which work as well with StringIO and Tempfile.

Fixes #425

### DIFF
--- a/stdlib/builtin/builtin.rbs
+++ b/stdlib/builtin/builtin.rbs
@@ -30,6 +30,25 @@ interface _Each[out A, out B]
   def each: { (A) -> void } -> B
 end
 
+interface _Reader
+  def read: (?int length, ?string outbuf) -> String?
+end
+
+interface _Writer
+  # Writes the +data+ string. Returns the number of bytes written
+  def write: (*_ToS data) -> Integer
+end
+
+interface _Rewindable
+  # Positions the stream to the beginning of input, resetting `lineno` to zero.
+  #
+  def rewind: () -> Integer
+end
+
+interface _ToIO
+  def to_io: () -> IO
+end
+
 interface _Exception
   def exception: () -> Exception
                | (String arg0) -> Exception
@@ -40,3 +59,5 @@ type real = Integer | Float | Rational
 
 type string = String | _ToStr
 type encoding = Encoding | string
+
+type io = IO | _ToIO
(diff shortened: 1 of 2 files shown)


soutaro:
Thanks @HoneyryderChuck!

I like the idea of `_Reader`, `_Writer`, `_Rewindable`, and `_ToIO`. One concern is if using the interfaces to define `IO` and `StringIO` is better or not. It may be better to have `def`s and RDoc inline.

HoneyryderChuck:
> It may be better to have defs and RDoc inline.

I understand the reasoning, but I don't understand how the rbs files came to be used as the RDoc repository. Is there any context worth sharing? IMO it doesn't make sense, and generates this conflict you mentioned, right?

soutaro:
@HoneyryderChuck The docs in RBS are used by Steep to generate inline documentation vis LSP. I don't think using interface to define the methods makes much sense, because the interfaces are extracted from the implementations while it makes difficult to understand the definition reading the RBS files.

It's very subtle and difficult to explain to me, but I'd like to keep the `defs` in the classes.

HoneyryderChuck:
@soutaro went ahead with the updates. Let me know what you think.

soutaro (approved):
Great. Thanks!

Issue #133, opened 2019-12-23
Add a shorthand for "TrueClass | FalseClass"


connorshea:
I see [in syntax.md there's a `bool` type]( https://github.com/ruby/ruby-signature/blob/7392f411f3fb97d2949e906c9f8f28e980c3863b/doc/syntax.md#bool-or-trueclass--falseclass), but it allows "`nil` or any other values", which seems to defeat the purpose a bit.

I'd expect passing anything other than `true` or `false` to warn, because otherwise the type signature doesn't seem to have a point?

```ruby
# rbs file
def foo: (bool) -> String
```

```ruby
# ruby file
baz = 'string'

foo(true) #=> good
foo(baz) #=> this should make the typechecker warn?
```

If changing `bool` isn't going to happen, I'd like a simple shorthand for `TrueClass | FalseClass` that would make passing `nil` or a string to a method expecting a boolean give a warning. Maybe `bool!`? I don't think `!` has any meaning in the signature syntax right now, but it might not be a good idea to use `!` if you plan on using it for something else at some point.

```ruby
# rbs file
def foo: (bool) -> String
def bar: (bool!) -> String
```

```ruby
# ruby file
baz = 'string'

foo(true) #=> good
foo(baz) #=> ok
bar(baz) # warning
```

Thanks :)

soutaro:
@connorshea 

Hi, thank you very much for your proposal and sorry for late reply.

Your proposal makes sense. It keeps the syntax of RBS, and we know there is some cases we need the `true | false` types. But I'm not sure how many of Ruby developers want this feature or how many of Ruby developers can choose one from the two types correctly.

So, my intuition is that:

* Some Ruby program need that, but the programmer can easily define the type like: `type boolean = true | false`.
* Adding extra checks about *implicit conversion to bool types* can be done by type checker or other tools like RuboCop.
* I want to see how many Ruby program wants the *strict bool* type.

Do these make sense?

connorshea:
That makes sense to me, though I still think a lot of users are going to want a strict bool type.

I'm fine with waiting and seeing if people actually need it before implementing it :)

sandstrom:
For this use-case, plus for regular in ruby, e.g. `my_var.is_a?(Boolean)` it would be great with a proper boolean super-class in Ruby.

This has been requested before:
https://bugs.ruby-lang.org/issues/12515

This is what Sorbet does:
https://sorbet.org/docs/class-types

marcandre:
[...]

PR #430, opened 2020-10-17
Add strict bool type

soutaro:
This PR is to add a new type `bool!` which is equivalent to `true | false`.

https://github.com/ruby/rbs/issues/133


soutaro:
https://github.com/ruby/rbs/pull/456

PR #432, opened 2020-10-18
Array signature and self

marcandre:
This PR fixes uses of `self` for array

It also uses explicit scope (`::Array` instead of `Array`)


soutaro:
Thanks! 🎉

PR #436, opened 2020-10-19
Doc tweaks

marcandre:
This PR
- fixes the doc syntax file for `unchecked`
- mentions the `out` and `in` parameters

I could not figure out what the `unchecked` does exactly so I did not add documentation for that.

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -433,8 +433,10 @@ _global-name_ ::= /$[a-zA-Z]\w+/ | ...
 _module-type-parameters_ ::=                                                  # Empty
                            | `[` _module-type-parameter_ `,` ... `]`
 
-_module-type-parameter_ ::= _variance_ _type-variable_
+_module-type-parameter_ ::= _check_ _variance_ _type-variable_
 _variance_ ::= `out` | `in`
+_check_ ::= # Empty
+          | `unchecked`
 ```
 
 ### Class declaration
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -450,6 +450,29 @@ class Ref[A] < Object
 end
 ```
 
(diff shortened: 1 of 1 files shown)


docs/syntax.md
```diff
@@ -448,6 +450,16 @@ class Ref[A] < Object
 end
 ```
 
+For classes with type parameters, you may specify if they are "invariant" (default), "covariant" (`out`) or "contravariant" (`in`). See [this definition of covariance and contravariance](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)).
+
+For example, an `Array` of `String` can almost be considered to be an `Array` of `Object`, but not the reverse, so we have:
+
+```
+class Array[out T]
```
soutaro, line 458: I'm not very sure if `Array[T]` is a good example here because it is `unchecked`... How about using `Range[T]` because it is truly covariant?

marcandre: Could you explain the purpose of `unchecked`?
Why is `Array/Hash` unchecked but `Range` is not?
I could be introduced that notion at the same time in the doc.

soutaro: `Array` and `Hash` are mutable and thus making the type parameters covariant breaks the type system.

```rb
a = [1,2,3]       # Assume type of the variable `a` is Array[Integer]
b = a             # Assume type of the variable `b` is Array[Integer | String]

b.push("foo")     # This is okay because b: Array[Integer | String]

a.last + 1       # 💥 
```

However, they are treated as covariant in popular languages (example: Java and TypeScript) because it's more useful and keeping them invariant confuses users.

`unchecked` is to skip the variance validation on type parameters.

marcandre:
Thanks for the explanation. I changed the doc to reflect that.

PR #437, opened 2020-10-19
Fix signature for `Enumerable#find`.

marcandre:
I wanted to put the `interface` declaration just above `find`, but I was getting an error:

```
/Users/mal/ruby/rbs/lib/rbs/errors.rb:122:in `check!': /Users/mal/ruby/rbs/stdlib/builtin/enumerable.rbs:323:26...323:30: Could not find Elem (RBS::NoTypeFoundError)
```

Is this a bug?

### DIFF
--- a/stdlib/builtin/enumerable.rbs
+++ b/stdlib/builtin/enumerable.rbs
@@ -307,11 +307,17 @@ module Enumerable[unchecked out Elem, out Return]: _Each[Elem, Return]
   # ```
   def to_h: () -> ::Hash[untyped, untyped]
 
-  def each_slice: (Integer n) { (::Array[Elem] arg0) -> untyped } -> NilClass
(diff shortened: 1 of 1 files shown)


soutaro:
@marcandre Generics parameters are local in `class`, `module`, and `interface`. So you anyway need to write `_NotFound[T]` even if it is written inside `Enumerable`.

marcandre:
> @marcandre Generics parameters are local in `class`, `module`, and `interface`. So you anyway need to write `_NotFound[T]` even if it is written inside `Enumerable`.

Yes, I understand that. I opened #440 to show you.

I rebased this PR to avoid conflict.

marcandre:
PR changed now that bug #440 is fixed.

soutaro (approved):
🎉

PR #438, opened 2020-10-19
Make Enumerable's predicate blocks return bool instead of untyped

mame:
For example, the return value of the block of Enumerable#all? is handled
as bool. In this case, it will bring no difference, but I think it is a
good habit to use a more explicit name instead of "untyped".

Note, if the definition of `bool` is changed according to #133, they
should be changed again.


marcandre (approved):

marcandre:
Agreed 👍 
I was waiting for #133 too to make a PR, but this could be merged beforehand anyways.

soutaro:
Thanks! @mame

PR #440, opened 2020-10-19
Potential bug with interface within generic class

marcandre:
Why does this not work?

Looks like a bug. Note that error message is really not informative (happens after the `interface`)

Followup to #437


soutaro:
Thanks. This should be a bug.

marcandre:
Closed, thanks for fixing the bug

PR #444, opened 2020-10-20
added [] and []= operators to the list of potential methods

HoneyryderChuck:
This fixes:

```
syntax error, unexpected '(', expecting ';' or '\\n' (SyntaxError)
def []__with__RBS_TEST_7414c4(*args)
                             ^
```

when having an annotation for a `def []` of r`def[]=` methods.


soutaro (approved):
Good catch! 👏

PR #449, opened 2020-10-26
Kernel#enum_for accepts no argument

mame:


mame:
Oops, there is another declaration of `Object#enum_for` too: https://github.com/ruby/rbs/blob/282b93300e93eb8c217973dc3a2e9527721e0915/stdlib/builtin/object.rbs#L208-L209

Maybe `Kernel#enum_for` should be removed?

mame:
I've created a PR to remove all Kernel's public methods that are defined in Object class. https://github.com/ruby/rbs/pull/450

PR #450, opened 2020-10-26
Remove public methods from Kernel module

mame:
These methods were doubly defined in both Kernel and Object.
They are technically defined in the Kernel module (in Ruby semantics),
but according to @soutaro, RBS adds the public methods to Object, not
Kernel.


soutaro:
🙇‍♂️

PR #451, opened 2020-10-28
Fix proc type parsing

soutaro:
This allows parsing the following RBS correctly.

```rbs
class Foo
  def foo: -> ^-> Integer
end
```


soutaro:
Thanks @mame for reporting this!

PR #456, opened 2020-11-01
Bool semantics

soutaro:
1. Make `bool` strict (an alias of `true | false`)
2. Introduce `boolish` type, an alias of `top`
3. Replace `bool` in RBS to `boolish` if applicable

See https://github.com/ruby/rbs/issues/133.


marcandre (approved):
👍

PR #457, opened 2020-11-01
Support alias in `rbs prototype rb`

pocke:
This pull request makes `rbs prototype rb` to support `alias`.

Currently `rbs prototype rb` just ignores `alias` syntax and `alias_method`. It will output `alias` by this change.

For example:

```ruby
# test.rb

class C
  def m() end
  alias m2 m
end
```

```bash
# Before
$ rbs prototype rb test.rb
class C
  def m: () -> nil
end

# after
$ rbs prototype rb test.rb
class C
  def m: () -> nil

  alias m2 m
end
```

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -178,9 +178,9 @@ module RBS
             end
           when :attr_reader
             args.each do |arg|
-              if arg&.type == :LIT && arg.children[0].is_a?(Symbol)
+              if arg && (name = literal_to_symbol(arg))
                 decls << AST::Members::AttrReader.new(
-                  name: arg.children[0],
+                  name: name,
                   ivar_name: nil,
                   type: Types::Bases::Any.new(location: nil),
                   location: nil,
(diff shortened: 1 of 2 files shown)


lib/rbs/prototype/rb.rb
```diff
@@ -166,9 +178,9 @@ def process(node, decls:, comments:, singleton:)
             end
           when :attr_reader
             args.each do |arg|
-              if arg&.type == :LIT && arg.children[0].is_a?(Symbol)
+              if arg && (name = literal_to_symbol(arg))
```
pocke, line 181: It is not the main part of this pull request, but `attr_*` accept Strings as a method names. It is the same behavior with `alias_method`.
So I changed them to accept Strings.


```ruby
class C
  attr_reader 'a'
  attr_writer 'a'
  attr_accessor 'b'
end

c = C.new

c.a = 1
c.b = 2
p c.a, c.b
# => 1
     2
```

https://github.com/ruby/rbs/pull/457/commits/f8accb0b82865f63ad0266c66fdee4f3974f0b8b

soutaro (approved):
💯

PR #462, opened 2020-11-07
Update Enumerable#to_h type

pocke:
`Enumerable#to_h` actually receives a block, but the type wasn't aware of the block.
So this pull request adds a method type to be aware of the block argument.

By the way, other `to_h`s support the block argument.

https://github.com/ruby/rbs/blob/967a421db6ff6907536235f2ae744791455a354e/core/hash.rbs#L857-L858
https://github.com/ruby/rbs/blob/967a421db6ff6907536235f2ae744791455a354e/core/array.rbs#L1855-L1856


marcandre (approved):

soutaro (approved):
💯

PR #470, opened 2020-11-09
Add more types for Random

zofrex:
This is my first time writing RBS so I am not sure it is right, especially the tests.

I am not sure how to fully describe `rand` because it can return whatever type you give it, for example:

```
> Random.rand(Tally.new("|")..Tally.new("|||||"))
=> #<Tally:0x00007fedf725f9f8 @string="|">
```

In other words, `rand: (Range[SomeType]) -> SomeType` but I don't know how to write that in RBS.

(Also, I think `SomeType` might have to be a descendent of `Numeric`, but the documentation is not clear)

### DIFF
--- a/core/random.rbs
+++ b/core/random.rbs
@@ -19,14 +19,19 @@
 # of 2**19937-1.
 #
 class Random < Object
-  include Random::Formatter
-
   # Creates a new PRNG using `seed` to set the initial state. If `seed` is
   # omitted, the generator is initialized with Random.new_seed.
   #
   # See Random.srand for more information on the use of seed values.
   #
   def initialize: (?Integer seed) -> void
+
+  include Random::Formatter
+
+  # Returns a random binary string. The argument `size` specifies the length of
+  # the returned string.
+  #
+  def self.bytes: (Integer size) -> String
 
   # Returns an arbitrary seed value. This is used by Random.new when no seed value
   # is specified as an argument.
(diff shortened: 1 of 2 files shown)


core/random.rbs
```diff
@@ -81,8 +134,9 @@ class Random < Object
   # (`-`) and add (`+`)methods, or rand will raise an ArgumentError.
   #
   def rand: () -> Float
-          | (Integer | ::Range[Integer] max) -> Integer
+          | (Integer | ::Range[Integer] | Numeric max) -> Integer
```
soutaro, line 137: Seems like having `Numeric` here will return `Integer` for any numbers including `Float`.
Moving this line after next line would give better type.

core/random.rbs
```diff
@@ -81,8 +134,9 @@ class Random < Object
...
           | (Float | ::Range[Float] max) -> Float
+          | (::Range[Numeric]) -> Numeric
```
soutaro, line 139: Yeah, the type we want would be `[T <: Numeric] (Range[T]) -> T`, but RBS currently doesn't allow writing it.

I think there are two options. `[T] (Range[T]) -> T` or `(Range[Numeric]) -> Numeric`.  The first option would make sense in my opinion, because testing the code will uncover the problem anyway.

soutaro (changes requested):
@zofrex Thanks for opening this pull request! Commented inline.

PR #471, opened 2020-11-10
Make Hash#merge! strict so that it can accept only the same type

mame:
... and Hash#replace, too.

Actually, these methods can accept a hash that has arbitrary key and
value types. However, they destructively changes its receiver. Thus,
their RBS declarations should be strict so that they can accept only
`Hash[K, V]`. Otherwise, a type checker will overlook the change of
element types of the receiver hashes. This is not new; `Array#concat`
only accepts `Array[Elem]` though the method actually accepts any array
that has arbitrary elements.

In addition, TypeProf assumes that contravariant occurences of type variables `K` and `V` updates the receiver's elements. The former declarations are inconvenient for TypeProf.

### DIFF
--- a/core/hash.rbs
+++ b/core/hash.rbs
@@ -720,8 +720,8 @@ class Hash[unchecked out K, unchecked out V] < Object
   #
   # Hash#update is an alias for Hash#merge!.
   #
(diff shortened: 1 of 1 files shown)


soutaro (approved):
Thanks! 👍

PR #474, opened 2020-11-13
Drop second type parameter of `Enumerable` and `_Each`

soutaro:
* The second type parameter is not used so much.
* To make the type definition more compatible with Sorbet.


core/enumerable.rbs
```diff
@@ -66,24 +66,24 @@ module Enumerable[unchecked out Elem, out Return]: _Each[Elem, Return]
            | () { (Elem) -> boolish } -> Integer
 
   def cycle: (?Integer n) { (Elem arg0) -> untyped } -> NilClass
-           | (?Integer n) -> ::Enumerator[Elem, Return]
```
soutaro, line 69: Many `Return`s in `Enumerable` definition are incorrect.

PR #481, opened 2020-11-22
Support more meta programmings for prototype rb

pocke:
`rbs protoype rb` will be aware of `private`, `public`, and `module_function` by this pull request.

For example:

```ruby
# ruby code

module M
  def foo
  end

  module_function def bar
  end

  module_function

  def baz
  end
end

class C
  private def foo
  end

  def bar
  end

  private

  def baz
  end
end
```

```ruby
# generated rbs

module M
  def foo: () -> nil

  def self?.bar: () -> nil

  def self?.baz: () -> nil
end

class C
  private

  def foo: () -> nil

  public

  def bar: () -> nil

  private

  def baz: () -> nil
end
```

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -1,9 +1,19 @@
 module RBS
   module Prototype
     class RB
-      Context = Struct.new(:module_function_in_ruby, :module_function_in_rbs, :singleton, keyword_init: true) do
+      Context = Struct.new(:module_function, :singleton, keyword_init: true) do
         def self.initial
-          self.new(module_function_in_ruby: false, module_function_in_rbs: false, singleton: false)
+          self.new(module_function: false, singleton: false)
+        end
+
+        def method_kind
+          if singleton
+            :singleton
+          elsif module_function
+            :singleton_instance
+          else
+            :instance
+          end
         end
       end
 
@@ -115,7 +125,7 @@ module RBS
         when :DEFN, :DEFS
             if node.type == :DEFN
               def_name, def_body = node.children
-              kind = context.singleton ? :singleton : :instance
+              kind = context.method_kind
             else
               _, def_name, def_body = node.children
               kind = :singleton
@@ -154,7 +164,7 @@ module RBS
           )
           decls.push member unless decls.include?(member)
 
-        when :FCALL
+        when :FCALL, :VCALL
           # Inside method definition cannot reach here.
           args = node.children[1]&.children || []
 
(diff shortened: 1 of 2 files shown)


lib/rbs/prototype/rb.rb
```diff
@@ -1,6 +1,22 @@
 module RBS
   module Prototype
     class RB
+      Context = Struct.new(:module_function, :singleton, keyword_init: true) do
```
pocke, line 4: I introduced `context `for this feature to `prototype rb`, instead of `singleton:` keyword args passing.

Currently it has information about module_function and singleton. And I have a plan to use it for `extend self` because `extend self` needs what is `self` in the context.

lib/rbs/prototype/rb.rb
```diff
@@ -220,16 +239,55 @@ def process(node, decls:, comments:, singleton:)
...
+                if arg && (name = literal_to_symbol(arg))
+                  if i = find_def_index_by_name(decls, name)
+                    current = current_accessibility(decls, i)
+                    if current != accessibility
+                      decls.insert(i + 1, current)
+                      decls.insert(i, accessibility)
+                    end
+                  end
+                end
+              end
```
pocke, line 277: For `private :method_name` syntax.

It finds method definition member from the literal name, and inserts accessibility methods around the method definition.
For example


```ruby
# Ruby
class C
  def foo() end
  private :foo
  def bra() end
end
```

```ruby
# RBS
class C
  private # Make foo private
  def foo: () -> nil
  public # Insert `public` to avoid to make bar private
  def bar: () -> nil
end
```

lib/rbs/prototype/rb.rb
```diff
@@ -220,16 +239,55 @@ def process(node, decls:, comments:, singleton:)
...
 
-          each_child node do |child|
-            process child, decls: decls, comments: comments, singleton: singleton
+              # For `private def foo` syntax
+              current = current_accessibility(decls)
+              decls << accessibility
+              each_child node do |child|
+                process child, decls: decls, comments: comments, context: context
+              end
+              decls << current
```
pocke, line 285: It processes arguments of `private` and `public` with the accessibility for `private def foo` syntax.

lib/rbs/prototype/rb.rb
```diff
@@ -586,6 +644,59 @@ def node_type(node, default: Types::Bases::Any.new(location: nil))
...
+            current = current_accessibility(decls, idx)
+            next
+          end
+
+          current = decl if is_accessibility?(decl)
+          idx += 1
+        end
+
+        decls.pop while decls.last && is_accessibility?(decls.last)
+      end
```
pocke, line 684: `process` method may insert unnecessary accessibility methods. So it removes them when `process` method processed an entire class/module.


For example:


```ruby
# Ruby


class C
  private def foo() end
  private def bar() end
end
```

`prototype rb` will convert the Ruby code to the following RBS, if it works without `remove_unnecessary_accessibility_methods!`.

```ruby
# RBS

class C
  private
  def foo: () -> nil
  public # <- redundant

  private # <- redundant
  def bar: () -> nil
  public # <- redundant
```


So this method removes them.

soutaro (approved):
🌻

PR #482, opened 2020-11-23
Add type of Singleton module

pocke:
This pull request adds types of Singleton module.

Note that Singleton module makes `new` and `allocate` methods private, but this pull request doesn't treat it.
https://github.com/ruby/ruby/blob/53e352fd718cd2cae6d77298e6e92736dddcfeeb/lib/singleton.rb#L164

I added a test only for `Singleton.instance` method. Because tests for other methods are not useful. I guess they are almost private APIs.

### DIFF
--- a/stdlib/singleton/0/singleton.rbs
+++ b/stdlib/singleton/0/singleton.rbs
@@ -89,30 +89,23 @@
 #     p a.keep  #  => "keep this"
 #     p a.strip #  => nil
 module Singleton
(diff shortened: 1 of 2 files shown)


soutaro (approved):

soutaro:
> Note that `Singleton` module makes `new` and `allocate` methods private

Got it. Yes, RBS doesn't support it...

PR #487, opened 2020-11-23
Ignore refinements in `rbs prototype rb`

pocke:
`rbs prototype rb` will ignore refinements by this pull request.

# Problem

Currently `rbs prototype rb` command generates incorrect RBS code for Ruby code that uses refinements.
For example

```ruby
# test.rb

module M
  refine Array do
    def cool_method() end
  end
end
```

```console
module M
  def cool_method: () -> nil
end
```

The `cool_method` is not `M` module's method, but prototype rb isn't aware of it.

# Solution

Just ignore the body of `refine` method.
By this change, it just omits methods for refinements.

```console
# For the same test.rb

module M
end
```

RBS doesn't have support for refinements, so I think we can ignore it.

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -120,9 +120,7 @@ module RBS
           end
 
           ctx = Context.initial.tap { |ctx| ctx.singleton = true}
-          each_child(body) do |child|
-            process child, decls: decls, comments: comments, context: ctx
-          end
+          process_children(body, decls: decls, comments: comments, context: ctx)
 
         when :DEFN, :DEFS
             if node.type == :DEFN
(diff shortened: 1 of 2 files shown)


lib/rbs/prototype/rb.rb
```diff
@@ -306,9 +309,13 @@ def process(node, decls:, comments:, context:)
...
 
         else
-          each_child node do |child|
-            process child, decls: decls, comments: comments, context: context
-          end
+          process_children(node, decls: decls, comments: comments, context: context)
+        end
+      end
+
+      def process_children(node, decls:, comments:, context:)
```
soutaro, line 316: 👍

PR #491, opened 2020-11-26
Fix generating incorrect accessibility by prototype rb

pocke:
# Problem

I implemented generating accessibility methods in #481 , but it has a bug :bowing_man: 
It generates incorrect accessibility when methods are continually defined with multiple `private def foo() end` syntax.

For example:

```ruby
class C
  private def prv1() end
  private def prv2() end
  def pub() end
end
```

```bash
class C
  private

  def prv1: () -> nil

  def prv2: () -> nil

  def pub: () -> nil
end
```

it needs `public` before `def pub: () -> nil`, but there isn't.

# Cause

I misunderstood the usage of Array#slice. It needs to receive the size of the array if I need the entire of the array, but I passed the size - 1.

So this pull request replaces `size - 1` with just `Array#size`

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -661,7 +661,7 @@ module RBS
         @public ||= AST::Members::Public.new(location: nil)
       end
 
-      def current_accessibility(decls, index = [0, decls.size - 1].max)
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍

PR #492, opened 2020-11-26
Support `extend self` by prototype rb

pocke:
This pull request adds support for `extend self` by prototype rb.

For example, `module M; extend self; end` will be converted to `module M; extend ::M; end`. `self` that is in arguments of `extend` is converted to the module name in the current scope. 

It converts `self` to an absolute type name. It means it always generates a full path as an argument of `extend`.
Because a relative path can be incorrect.
For example:

```ruby
module M
  module M
  end

  # In this context, `extend M` means `extend ::M::M`,
  # so it should be converted to `extend ::M`.
  extend self
end
```

---

it can resolve `include self` with the same technique, but I didn't implement it because of meaningless.
`include self` raises an error, so actually no one uses this.

```bash
$ ruby -e module M; include self; end
-e:1:in `append_features': cyclic include detected (ArgumentError)
	from -e:1:in `include'
	from -e:1:in `<module:M>'
	from -e:1:in `<main>'
```

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -1,9 +1,9 @@
 module RBS
   module Prototype
     class RB
-      Context = Struct.new(:module_function, :singleton, keyword_init: true) do
-        def self.initial
-          self.new(module_function: false, singleton: false)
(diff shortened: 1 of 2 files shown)


soutaro (approved):
💯

PR #493, opened 2020-11-27
Assume that rbs files are encoded in UTF-8

mame:
Formerly, rake fails when `LANG=C` because ruby assumes the input files
are encoded in US-ASCII.

By this change, all rbs files are read as UTF-8.
Nowadays, I think a major part of ruby (and RBS) code are in UTF-8. So
this gives good first aid for the issue.

A magic comment like `# encoding: something` may be needed in RBS as
well as ruby code, but I think it requires further discussion.


soutaro (approved):
👏

PR #497, opened 2020-12-02
Remove deprecated DBM#index

nobu:


soutaro:
@nobu Should we delete `DBM#index` from `dbm.rbs` too?

nobu:
Yes, I've overlooked it.

soutaro (approved):

PR #501, opened 2020-12-03
Add support for escape sequences in double quoted rbs strings

johansenja:
I came across these earlier:
```ruby
class A
  B = "a\nb"
end
```
```ruby
class A
  B: "a\nb"
end
```
They were incompatible because rbs appeared to be handling double-quoted strings differently to ruby. This PR aims to bring support for newlines and other relatively common escape sequences
```ruby
# rbs - existing workaround (for new lines at least)
class A
  B: "a
b"
end
```


soutaro (approved):
Thanks! 🙌

PR #505, opened 2020-12-06
Support singleton attr by prototype rb

pocke:
This pull request adds to support singleton attr, which was introduced by #502, to `rbs prototype rb` command.

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -15,6 +15,14 @@ module RBS
             :instance
           end
         end
+
+        def attribute_kind
+          if singleton
+            :singleton
+          else
+            :instance
+          end
+        end
       end
 
       attr_reader :source_decls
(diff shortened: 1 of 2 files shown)


lib/rbs/prototype/rb.rb
```diff
@@ -15,6 +15,14 @@ def method_kind
             :instance
           end
         end
+
+        def attribute_kind
```
pocke, line 19: https://github.com/ruby/rbs/blob/8871f6da72bd8c4c3e08dbd52dcc9bd7820e1512/lib/rbs/parser.y#L205-L207
Attribute doesn't support `singleton_instance` syntax (`self?`), unlike method definition. Is it intentional?

soutaro: It's intentional but I'm open to adding the syntax if it's better.

I was assuming that we rarely define attributes for both singleton and instance while the presence of `module_function` implies it happens frequently for`def`s.

soutaro (approved):
🙏

lib/rbs/prototype/rb.rb
pocke: > I was assuming that we rarely define attributes for both singleton and instance while the presence of module_function implies it happens frequently for `def`s.


I completely agree. It is rare and I do not have any use cases for `singleton_instance` attributes.

PR #511, opened 2020-12-07
Add Time (extension) signatures

ybiquitous:
This change adds signatures for a `Time` class extension (not built-in) via `require "time"`.

Note:
- Generated signatures by the `rbs prototype runtime --require time Time` command conflicts
  with the existing signatures in `core/time.rbs`:
  https://github.com/ruby/rbs/blob/71b8e534e0adcc78aa76ffbd0326ffe01594d520/core/time.rbs
- So, I use the command `rbs prototype rb rbs prototype rb /usr/local/lib/ruby/2.7.0/time.rb`,
  and edit the majority of the generated file by hand (especially, remove the private APIs).

See also:
- https://github.com/ruby/ruby/blob/v2_7_2/lib/time.rb
- https://ruby-doc.org/stdlib-2.7.2/libdoc/time/rdoc/Time.html

### DIFF
--- a/stdlib/time/0/time.rbs
+++ b/stdlib/time/0/time.rbs
@@ -1,0 +1,327 @@
+class Time
+  interface _TimeLike
+    def year: () -> Integer
+    def mon: () -> Integer
+    def day: () -> Integer
+  end
+
+  #
+  # Return the number of seconds the specified time zone differs
+  # from UTC.
+  #
+  # Numeric time zones that include minutes, such as
+  # <code>-10:00</code> or <code>+1330</code> will work, as will
+  # simpler hour-only time zones like <code>-10</code> or
+  # <code>+13</code>.
+  #
+  # Textual time zones listed in ZoneOffset are also supported.
+  #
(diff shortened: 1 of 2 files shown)


stdlib/time/0/time.rbs
```diff
@@ -0,0 +1,327 @@
+class Time
+  interface _TimeLike
```
ybiquitous, line 2: [note] I cannot find a good name... 😓

test/stdlib/TimeExtension_test.rb
```diff
@@ -0,0 +1,119 @@
+require_relative "test_helper"
```
ybiquitous, line 1: [note] Compared with `Time_test.rb`, the file name `TimeExtension_test.rb` may be confusing... 🤔 
I'm open for a better name suggestion.

soutaro (approved):
👏

ybiquitous:
@soutaro Thank you! 😄

PR #513, opened 2020-12-08
Fix ruby2_keywords for Proc in Ruby <2.7

palkan:
We should return `self` for Procs.

Otherwise hook makes `wrapped_block` to be nil here: https://github.com/ruby/rbs/blob/71b8e534e0adcc78aa76ffbd0326ffe01594d520/lib/rbs/test/hook.rb#L121


soutaro (approved):
Nice catch! 👍

PR #520, opened 2020-12-16
Improve `IO.binread` and `IO.binwrite` signatures

ybiquitous:
This change aims to improve the signatures of `IO.binread` and `IO.binwrite`.

- Add missing documents.
- Add missing tests.
- Improve the argument types of the `.binwrite` method.

In creating this change, I refer to the following resources:

- https://github.com/ruby/ruby/blob/ruby_2_7/spec/ruby/core/io/binread_spec.rb
- https://github.com/ruby/ruby/blob/ruby_2_7/spec/ruby/core/io/binwrite_spec.rb
- https://github.com/ruby/ruby/blob/ruby_2_7/spec/ruby/core/io/shared/binwrite.rb
- https://ruby-doc.org/core-2.7.2/IO.html#method-c-binread
- https://ruby-doc.org/core-2.7.2/IO.html#method-c-binwrite

### DIFF
--- a/core/io.rbs
+++ b/core/io.rbs
@@ -510,9 +510,20 @@ class IO < Object
 
   def write: (*_ToS arg0) -> Integer
 
+  # Opens the file, optionally seeks to the given *offset*, then returns *length*
+  # bytes (defaulting to the rest of the file). #binread ensures the file is
+  # closed before returning.  The open mode would be `"rb:ASCII-8BIT"`.
+  #
+  #     IO.binread("testfile")           #=> "This is line one\nThis is line two\nThis is line three\nAnd so on...\n"
+  #     IO.binread("testfile", 20)       #=> "This is line one\nThi"
+  #     IO.binread("testfile", 20, 10)   #=> "ne one\nThis is line "
+  #
   def self.binread: (String name, ?Integer length, ?Integer offset) -> String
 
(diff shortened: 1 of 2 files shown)


core/io.rbs
```diff
@@ -510,9 +510,20 @@ class IO < Object
...
+  #     IO.binread("testfile", 20)       #=> "This is line one\nThi"
+  #     IO.binread("testfile", 20, 10)   #=> "ne one\nThis is line "
+  #
   def self.binread: (String name, ?Integer length, ?Integer offset) -> String
 
-  def self.binwrite: (String name, _ToS arg0, ?Integer offset, ?external_encoding: String external_encoding, ?internal_encoding: String internal_encoding, ?encoding: String encoding, ?textmode: untyped textmode, ?binmode: untyped binmode, ?autoclose: untyped autoclose, ?mode: String mode) -> Integer
+  # Same as IO.write except opening the file in binary mode and ASCII-8BIT
+  # encoding (`"wb:ASCII-8BIT"`).
+  #
+  def self.binwrite: (String name, _ToS string, ?Integer offset, ?mode: String mode) -> Integer
```
ybiquitous, line 526: [note] About the argument changes of `.binwrite`:

- I remove optional arguments (`open_args`) except `mode:` because it is a misuse to specify `encoding:`, `textmode:` or `autoclose:` etc. for this method. But this removal is potentially breaking so I’m ready to revert it if you want.
- More readable argument name: `arg0` -> `string` (from the API doc)

soutaro: Looks reasonable. Nice catch! 👏

PR #527, opened 2020-12-18
Remove `[options...]` from help message of prototype rb(i) commands

pocke:
Because they receive no options.

I've confirmed other help messages correspond with the implementations. :heavy_check_mark:


soutaro (approved):
🙇‍♂️ 🙏

PR #529, opened 2020-12-18
Add URI singleton methods

ybiquitous:
This change adds signatures of the `URI` module singleton methods (and constants).
This also adds `URI:RFC*` modules and classes to pass the test but their content are empty still.

I refer to the following resources:
- https://ruby-doc.org/stdlib-2.7.2/libdoc/uri/rdoc/URI.html
- https://github.com/ruby/ruby/blob/v2_7_2/lib/uri/common.rb
- https://github.com/ruby/ruby/blob/v2_7_2/test/uri/test_common.rb

### DIFF
--- a/stdlib/uri/0/common.rbs
+++ b/stdlib/uri/0/common.rbs
@@ -1,0 +1,401 @@
+# URI is a module providing classes to handle Uniform Resource Identifiers
+# ([RFC2396](http://tools.ietf.org/html/rfc2396)).
+#
+# ## Features
+#
+# *   Uniform way of handling URIs.
+# *   Flexibility to introduce custom URI schemes.
+# *   Flexibility to have an alternate URI::Parser (or just different patterns
+#     and regexp's).
+#
+#
+# ## Basic example
+#
+#     require 'uri'
+#
+#     uri = URI("http://foo.com/posts?id=30&limit=5#time=1305298413")
+#     #=> #<URI::HTTP http://foo.com/posts?id=30&limit=5#time=1305298413>
+#
+#     uri.scheme    #=> "http"
+#     uri.host      #=> "foo.com"
+#     uri.path      #=> "/posts"
+#     uri.query     #=> "id=30&limit=5"
+#     uri.fragment  #=> "time=1305298413"
+#
+#     uri.to_s      #=> "http://foo.com/posts?id=30&limit=5#time=1305298413"
+#
+# ## Adding custom URIs
+#
+#     module URI
+#       class RSYNC < Generic
+#         DEFAULT_PORT = 873
+#       end
+#       @@schemes['RSYNC'] = RSYNC
+#     end
+#     #=> URI::RSYNC
+#
+#     URI.scheme_list
+#     #=> {"FILE"=>URI::File, "FTP"=>URI::FTP, "HTTP"=>URI::HTTP,
+#     #    "HTTPS"=>URI::HTTPS, "LDAP"=>URI::LDAP, "LDAPS"=>URI::LDAPS,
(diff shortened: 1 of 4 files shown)


ybiquitous:
Oh, I forget to add the `Kernel::URI` method. I'll add it soon.

ybiquitous:
Done. (squashed)

stdlib/uri/0/common.rbs
```diff
@@ -0,0 +1,401 @@
...
+  #     ary = URI.decode_www_form("a=1&a=2&b=3")
+  #     ary                   #=> [['a', '1'], ['a', '2'], ['b', '3']]
+  #     ary.assoc('a').last   #=> '1'
+  #     ary.assoc('b').last   #=> '3'
+  #     ary.rassoc('a').last  #=> '2'
+  #     Hash[ary]             #=> {"a"=>"2", "b"=>"3"}
+  #
+  # See URI.decode_www_form_component, URI.encode_www_form.
+  #
+  def self.decode_www_form: (String str, ?encoding enc, ?isindex: bool, ?use__charset_: bool, ?separator: String) -> Array[[ String, String ]]
```
soutaro, line 119: It seems like `isindex` and `use__charset_` accepts any value, not only `true` and `false`.
`boolish` would be better?

soutaro (changes requested):
`boolish` looks better to me.

stdlib/uri/0/common.rbs
ybiquitous: Surely, `boolish` is more flexible. I'll fix it soon.

ybiquitous: Fixed via 708bbf4

ybiquitous:
By the way, does adding a Goodcheck rule below make sense? 👇 

```yaml
  - id: prefer_boolish_over_bool
    pattern:
      - regexp: '\([^(]*\bbool\b[^\n]*\)'
      - token: "-> bool }"
    message: |
      Prefer `boolish` over `bool` for method arguments and block return values

      See the doc below:
      https://github.com/ruby/rbs/blob/78d04a2db0f1c4925d2b13c2939868edf9465d6c/docs/syntax.md#bool-or-boolish
    glob:
      - "**/*.rbs"
    justification:
      - When the method arguments or block return values require `true` or `false` strictly.
    pass:
      - "(arg: boolish)"
      - "{ () -> boolish }"
    fail:
      - "(arg: bool)"
      - "{ () -> bool }"
```

According to <https://github.com/ruby/rbs/blob/78d04a2db0f1c4925d2b13c2939868edf9465d6c/docs/syntax.md#bool-or-boolish>.

E.g.

```console
$ be goodcheck check -R prefer_boolish_over_bool core
core/complex.rbs:168:14: Prefer `boolish` over `bool` for method arguments and block return values
  def clone: (?freeze: bool) -> self
             ^~~~~~~~~~~~~~~
core/exception.rbs:193:21: Prefer `boolish` over `bool` for method arguments and block return values
  def full_message: (?highlight: bool, ?order: :top | :bottom) -> String
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
core/float.rbs:174:14: Prefer `boolish` over `bool` for method arguments and block return values
  def clone: (?freeze: bool) -> self
             ^~~~~~~~~~~~~~~
```

stdlib/uri/0/common.rbs
ybiquitous: I've rebased `master` and squashed to one commit.

ybiquitous:
I've open PR #531 to add a Goodcheck rule as a test, but please close the PR if you don't want.

soutaro (approved):
Thanks!

PR #530, opened 2020-12-18
Fix incorrect glob in goodcheck.yml

ybiquitous:
We have `.rbs` files also in `core/`, not also `stdlib/`.

### DIFF
--- a/goodcheck.yml
+++ b/goodcheck.yml
@@ -3,7 +3,7 @@ rules:
     pattern: 💪👽🚨
     message: Do you forget to delete `arglists` section?
     glob:
-      - "stdlib/**/*.rbs"
+      - "{core,stdlib}/**/*.rbs"
     fail:
       - |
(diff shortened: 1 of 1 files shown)


goodcheck.yml
```diff
@@ -3,7 +3,7 @@ rules:
     pattern: 💪👽🚨
     message: Do you forget to delete `arglists` section?
     glob:
-      - "stdlib/**/*.rbs"
+      - "**/*.rbs"
```
soutaro, line 6: Nice catch. 👍 
Could you make it `stdlib/**/*.rbs` and `core/**/*.rbs`? The distributed RBS files are saved there.

ybiquitous: I've done it! (rebasing `master` and squashed to one commit)

soutaro (approved):
Thanks! 💯

PR #536, opened 2020-12-19
[prototype runtime] Fill type arguments with untyped

pocke:
This pull request makes `rbs prototype runtime` to fill type arguments to super classes, included modules, and constant types.

Problem
===

Currently `rbs prototype runtime` isn't ware of type parameters.
So it generates unusable RBS for Array, Hash, and so on.
For example:

```ruby
class C < Array
  ARRAY = [1,2,3]
  HASH = { foo: 42 }
end

class C2
  include Enumerable
end
```

```console
$ rbs prototype runtime -R test.rb C C2
class C < Array      # it should be Array[something]
end

C::ARRAY: Array      # it should be Array[something]

C::HASH: Hash        # it should be Hash[something, something]

class C2
  include Enumerable # it should be Enumerable[something]
end
```

It doesn't pass `rbs validate` check.

```
$ rbs -I . validate --silent
/path/to/lib/rbs/errors.rb:33:in `check!': test.rbs:2:0...3:3: ::Array expects parameters [Elem], but given args [] (RBS::InvalidTypeApplicationError)
	from /path/to/lib/rbs/definition.rb:156:in `apply'
	from /path/to/lib/rbs/definition_builder.rb:294:in `instance_ancestors'
	from /path/to/lib/rbs/definition_builder.rb:409:in `block in build_instance'
	from /path/to/lib/rbs/definition_builder.rb:1098:in `try_cache'
	from /path/to/lib/rbs/definition_builder.rb:403:in `build_instance'
	from /path/to/lib/rbs/cli.rb:423:in `block in run_validate'
	from /path/to/lib/rbs/cli.rb:421:in `each_key'
	from /path/to/lib/rbs/cli.rb:421:in `run_validate'
	from /path/to/lib/rbs/cli.rb:113:in `run'
	from /path/to/exe/rbs:7:in `<top (required)>'
	from /home/pocke/.rbenv/versions/trunk/bin/rbs:23:in `load'
	from /home/pocke/.rbenv/versions/trunk/bin/rbs:23:in `<main>'
```

Solution
====

Fill type arguments with `untyped`.

So the command generates the following RBS with this patch.

```console
$ rbs prototype runtime -R test.rb C C2
class C < Array[untyped]
end

C::ARRAY: Array[untyped]

C::HASH: Hash[untyped, untyped]

class C2
  include Enumerable[untyped]
end
```

Note
===

* It uses `env` to search type parameters information. It means it only
  knows core classes by default. But we can use `-I`, `-r` options to
  change the loaded libraries.
* Currently it always fill type arguments with `untyped`. But I think we
  can guess the type arguments for Array and Hash constants. For
  example, `[1,2,3]` is `Array[Integer]`. So I'd like to support Array
  and Hash to output specific types in the future.
* I guess we can apply the same feature to `rbs prototype rb`, but this
  patch doesn't to keep this patch simple. It is future work.

### DIFF
--- a/lib/rbs/prototype/runtime.rb
+++ b/lib/rbs/prototype/runtime.rb
@@ -80,12 +80,12 @@ module RBS
             unless const_name(mix)
               RBS.logger.warn("Skipping anonymous module #{mix} included in #{mod}")
             else
-              module_name = to_type_name(const_name(mix))
-              if module_name.namespace == type_name.namespace
-                module_name = TypeName.new(name: module_name.name, namespace: Namespace.empty)
+              module_name = module_full_name = to_type_name(const_name(mix))
+              if module_full_name.namespace == type_name.namespace
+                module_name = TypeName.new(name: module_full_name.name, namespace: Namespace.empty)
               end
 
-              yield module_name, mix
+              yield module_name, module_full_name, mix
             end
           end
         end
@@ -319,7 +319,9 @@ module RBS
                      location: nil
                    )
                  else
-                   Types::ClassInstance.new(name: to_type_name(const_name(value.class)), args: [], location: nil)
+                   value_type_name = to_type_name(const_name(value.class))
+                   args = type_args(value_type_name)
+                   Types::ClassInstance.new(name: value_type_name, args: args, location: nil)
                  end
 
           @decls << AST::Declarations::Constant.new(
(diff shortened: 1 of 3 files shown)


soutaro (approved):
Great improvement! 🎉

PR #537, opened 2020-12-20
[Writer] Escape method names that are invalid with `def` syntax

pocke:
This pull request fixes the escaping strategy of RBS::Writer to escape invalid method names for `def` syntax correctly.

# Problem

Currently `RBS::Writer` doesn't escape invalid method names for `def` syntax. For example

```ruby
require 'rbs'
require 'stringio'

rbs = RBS::Parser.parse_signature(<<~RBS)
  class C
    def `foo!=`: () -> untyped
  end
RBS

io = StringIO.new

RBS::Writer.new(out: io).write(rbs)

puts io.string
# => class C
#      def foo!=: () -> untyped
#    end

RBS::Parser.parse_signature(io.string)
# => RBS::Parser::SyntaxError
```

I expect ``def `foo!=`: () -> untyped`` as the output, but it displays without backtick escaping. So it generates invalid RBS.

By the way, we can't define `foo!=` method with `def` syntax even in Ruby. But we can use `define_method` for this method name.

```ruby
class C
  define_method(:'foo!=') {}
end

p C.instance_methods(false) # => [:"foo!="]
```

# Solution

Escape method names correctly.

### DIFF
--- a/lib/rbs/writer.rb
+++ b/lib/rbs/writer.rb
@@ -232,10 +232,11 @@ module RBS
     def method_name(name)
       s = name.to_s
 
-      if /\A#{Parser::KEYWORDS_RE}\z/.match?(s)
-        "`#{s}`"
-      else
+      if [:tOPERATOR, :kAMP, :kHAT, :kSTAR, :kLT, :kEXCLAMATION, :kSTAR2, :kBAR].include?(Parser::PUNCTS[s]) ||
+          (/\A[a-zA-Z_]\w*[?!=]?\z/.match?(s) && !/\A#{Parser::KEYWORDS_RE}\z/.match?(s))
         s
+      else
+        "`#{s}`"
       end
     end
 
--- a/lib/rbs/writer.rb
+++ b/lib/rbs/writer.rb
@@ -233,7 +233,7 @@ module RBS
       s = name.to_s
 
       if [:tOPERATOR, :kAMP, :kHAT, :kSTAR, :kLT, :kEXCLAMATION, :kSTAR2, :kBAR].include?(Parser::PUNCTS[s]) ||
-          (/\A[a-zA-Z_]\w*[?!=]?\z/.match?(s) && !/\A#{Parser::KEYWORDS_RE}\z/.match?(s))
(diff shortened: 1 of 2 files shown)


lib/rbs/writer.rb
```diff
@@ -232,10 +232,11 @@ def write_member(member)
     def method_name(name)
       s = name.to_s
 
-      if /\A#{Parser::KEYWORDS_RE}\z/.match?(s)
-        "`#{s}`"
-      else
+      if [:tOPERATOR, :kAMP, :kHAT, :kSTAR, :kLT, :kEXCLAMATION, :kSTAR2, :kBAR].include?(Parser::PUNCTS[s]) ||
+          (/\A[a-zA-Z_]\w*[?!=]?\z/.match?(s) && !/\A#{Parser::KEYWORDS_RE}\z/.match?(s))
```
pocke, line 236: `RBS::Writer` escapes keywords, for example: ``def `def`: () -> untyped`. But this escaping is not necessary because it is allowed.

I keep this escaping because of compatibility, and I'm not sure why this escaping needs to keep. https://github.com/ruby/rbs/blob/a1e62612f2057a47259f4add6d15fe2700d6bf58/test/rbs/writer_test.rb#L114


If this escaping is not necessary, I can remove the escaping easily by removing `&& !/\A#{Parser::KEYWORDS_RE}\z/.match?(s)` condition.


What do you think? @soutaro

soutaro: It makes sense. I don't think we need to keep the quoting behavior. 

Let's make it:

1. Unquoted even if it is a keyword (proposed change)
2. Unquoted if it is safe `/\A[a-zA-Z_]\w*[?!=]?\z/`
3. Quoted otherwise

pocke: Thanks for your comment. I applied this charge with https://github.com/ruby/rbs/pull/537/commits/2b032125d03453cc54b48c9d706d8cea03d5190b.

But I noticed we still need to quote `self` and `self?` as a method name because it causes syntax error without quote.
So I replace  `Parser::KEYWORDS_RE` with `/\Aself\??\z/`.

soutaro (approved):
💯

PR #538, opened 2020-12-21
Add signature for Module#const_source_location

ybiquitous:
This change aims to add a missing signature for the `Module#const_source_location` method.
See the API doc: <https://ruby-doc.org/core-2.7.2/Module.html#const_source_location-method>


soutaro (approved):
👏

PR #540, opened 2020-12-21
Add missing extended/included modules to stdlib

pocke:
This pull request adds missing `include`s and `extend`s to stdlib type definitions.

We use `rbs prototype runtime` to generate a prototype of standard library types, but it didn't output `extend` until #535.
So I guessed we have some missing `extend`s in stdlib. Actually, they have some missing `extend`s, so I added them.

# How 

I checked all RBS under `stdlib/` with `Klass.singleton_class.included_modules - Object.included_modules`.

By the way, I didn't check `core/` directory, because I was tired, and I guess we have a few (or nothing) `extend`s in core libraries.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -45,6 +45,14 @@ task :validate => :parser do
       lib << "monitor"
     end
 
+    if lib == ["csv"]
+      lib << "forwardable"
+    end
+
+    if lib == ["prime"]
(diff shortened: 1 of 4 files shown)


soutaro (approved):
🙆‍♀️

PR #549, opened 2020-12-26
Make Kernel methods to be available as module functions

pocke:
This pull request makes Kernel methods to be available as module functions. 

# Problem

Most `Kernel` methods are available as module functions, but currently any `Kernel` methods aren't.

For example, `Kernel.puts "Hello"` is a valid Ruby code but Steep complain it.
Because `Kernel.puts` is typed as a private method accidentally.

# Solution

Most methods are module functions, but it is not all methods. So I need to extract the target methods in the beginning.

I extracted the target methods with the following code on IRB.

```ruby
Kernel.methods.map { Kernel.method(_1) }.select { _1.owner == Kernel.singleton_class }.map(&:name)
# => [:pp, :Pathname, :local_variables, :rand, :srand, :sprintf, :format, :Integer, :String, :Array, :Hash, :warn, :trap, :test, :require, :raise, :fail, :global_variables, :__method__, :__callee__, :__dir__, :require_relative, :autoload, :autoload?, :system, :eval, :iterator?, :block_given?, :catch, :throw, :loop, :exec, :binding, :gets, :abort, :proc, :lambda, :spawn, :Rational, :caller, :caller_locations, :set_trace_func, :exit!, :trace_var, :untrace_var, :at_exit, :Complex, :fork, :URI, :select, :load, :exit, :`, :syscall, :open, :printf, :print, :putc, :puts, :readline, :readlines, :p, :Float, :sleep]
```

And I changed the methods kinds to `self?` for the above methods.

# Note

I found some missing methods while I updated the RBS.
The following methods are available as Kernel's singleton methods, but RBS doesn't have any definitions. 

* trap
* spawn
* set_trace_func
* trace_var
* untrace_var

So we need to add them to `kernel.rbs` as module functions, but this pull request doesn't add them because of out of scope.
I'll work on them with another pull request.

--

I noticed this problem with this bloc post. Thanks! https://narazaka.hatenablog.jp/entry/2020/12/26/Ruby_3.0%E3%81%AE%E9%9D%99%E7%9A%84%E5%9E%8B%E5%AE%9A%E7%BE%A9%E3%82%92TypeScript%E3%81%BF%E3%81%9F%E3%81%84%E3%81%AB%E3%83%A9%E3%82%A4%E3%83%96%E3%83%A9%E3%83%AA%E3%81%AB%E6%9B%B8%E3%81%84%E3%81%A6

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -13,14 +13,14 @@
 module Kernel : BasicObject
   private
 
-  def caller: (?Integer start_or_range, ?Integer length) -> ::Array[String]?
+  def self?.caller: (?Integer start_or_range, ?Integer length) -> ::Array[String]?
             | (?::Range[Integer] start_or_range) -> ::Array[String]?
             | () -> ::Array[String]
 
-  def caller_locations: (?Integer start_or_range, ?Integer length) -> ::Array[Thread::Backtrace::Location]?
+  def self?.caller_locations: (?Integer start_or_range, ?Integer length) -> ::Array[Thread::Backtrace::Location]?
                       | (?::Range[Integer] start_or_range) -> ::Array[Thread::Backtrace::Location]?
 
-  def catch: [T] (T tag) { (T tag) -> untyped } -> untyped
+  def self?.catch: [T] (T tag) { (T tag) -> untyped } -> untyped
            | () { (Object tag) -> untyped } -> untyped
 
   # In a perfect world this should be:
@@ -38,7 +38,7 @@ module Kernel : BasicObject
   def define_singleton_method: (Symbol | String symbol, ?Proc | Method | UnboundMethod method) -> Symbol
                              | (Symbol | String symbol) { () -> untyped } -> Symbol
 
-  def eval: (String arg0, ?Binding arg1, ?String filename, ?Integer lineno) -> untyped
+  def self?.eval: (String arg0, ?Binding arg1, ?String filename, ?Integer lineno) -> untyped
 
   # Returns `true` if `yield` would execute a block in the current context.
   #
@@ -54,7 +54,7 @@ module Kernel : BasicObject
   # try { "hello" }      #=> "hello"
   # try do "hello" end   #=> "hello"
   # ```
-  def block_given?: () -> bool
+  def self.block_given?: () -> bool
 
   # Returns the names of the current local variables.
   #
@@ -65,9 +65,9 @@ module Kernel : BasicObject
   # end
   # local_variables   #=> [:fred, :i]
   # ```
-  def local_variables: () -> ::Array[Symbol]
+  def self?.local_variables: () -> ::Array[Symbol]
 
-  def srand: (?Numeric number) -> Numeric
+  def self?.srand: (?Numeric number) -> Numeric
 
   # Creates a subprocess. If a block is specified, that block is run in the
   # subprocess, and the subprocess terminates with a status of zero.
@@ -86,7 +86,7 @@ module Kernel : BasicObject
   #
   # Note that fork(2) is not available on some platforms like Windows and
   # NetBSD 4. Therefore you should use spawn() instead of fork().
-  def fork: () -> Integer?
+  def self?.fork: () -> Integer?
           | () { () -> untyped } -> Integer?
 
   def initialize_copy: (self object) -> self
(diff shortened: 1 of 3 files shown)


pocke:
Well, this change doesn't affect the accessibility of `Kernel.puts` singleton method.

The accessibility is still `private` with this patch.

```bash
$ exe/rbs method --singleton Kernel puts
::Kernel.puts
  defined_in: ::Kernel
  implementation: ::Kernel
  accessibility: private
  types:
      (*untyped arg0) -> ::NilClass
```


I guess the cause is a mismatch of Ruby's module_function and RBS's `self?`.

In Ruby, module_function works as the following:

* Make the instance method private
* Create an alias of the instance method as singleton method, but it is public.

But in RBS, I guess `self?` syntax doesn't affect accessibility. [`Kernel` module has `private` member on the top of the module definition](https://github.com/ruby/rbs/blob/07e9da36dc3dadbdcef5046a5e04d27284b01dc7/core/kernel.rbs#L14), so both of instance and singleton methods are private.

---

So, I think we have two solutions to fix this problem.

First, we can define private instance methods and public singleton methods by hand.
For example:

```ruby
# kernel.rbs

module Kernel : BasicObject
  private
  def puts: (*untyped) -> nil
  public
  def self.puts: (*untyped) -> nil

  # ...
end
```

It works, but I think it is not efficient.


Second, we can change the semantics of `self?` to make it the same with Ruby's module_function.
I guess we can write Kernel's definition and other module_function efficiently with this change.
But it may introduce an incompatibility, and it means `self?` syntax ignores `public` and `private` syntax. Maybe it introduce confusion about accessibility.

```ruby
module M
  private
  def self?.foo: () -> untyped
end

# M.foo is public or private???
```

core/kernel.rbs
```diff
@@ -112,50 +113,50 @@ module Kernel : BasicObject
...
   # ```
-  def Array: (NilClass x) -> [ ]
+  def self?.Array: (NilClass x) -> [ ]
            | [T] (::Array[T] x) -> ::Array[T]
            | [T] (::Range[T] x) -> ::Array[T]
            | [K, V] (::Hash[K, V] x) -> ::Array[[K, V]]
            | [T] (T x) -> ::Array[T]
 
-  def Complex: (Numeric | String x, ?Numeric | String y, ?exception: bool exception) -> Complex
+  def self?.Complex: (Numeric | String x, ?Numeric | String y, ?exception: bool exception) -> Complex
```
soutaro, line 122: > __[Goodcheck]__ Prefer `boolish` over `bool` for method arguments and block return values ([view](https://sider.review/gh/repos/174796313/pulls/549#issue-5268995))
>
> <details>
>
> See the doc below:
> https://github.com/ruby/rbs/blob/78d04a2db0f1c4925d2b13c2939868edf9465d6c/docs/syntax.md#bool-or-boolish
>
> | Rule |
> | ---- |
> | `rbs.prefer_boolish` |
>
> You can [close this issue](https://sider.review/gh/repos/174796313/pulls/549#issue-5268995) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

:+1: `true` of `false` is required here.

core/kernel.rbs
```diff
@@ -112,50 +113,50 @@ module Kernel : BasicObject
...
 
-  def Hash: [K, V] (Object x) -> ::Hash[K, V]
+  def self?.Hash: [K, V] (Object x) -> ::Hash[K, V]
 
-  def Integer: (Numeric | String arg, ?exception: bool exception) -> Integer
+  def self?.Integer: (Numeric | String arg, ?exception: bool exception) -> Integer
              | (String arg, ?Integer base, ?exception: bool exception) -> Integer
 
-  def Rational: (Numeric | String | Object x, ?Numeric | String y, ?exception: bool exception) -> Rational
+  def self?.Rational: (Numeric | String | Object x, ?Numeric | String y, ?exception: bool exception) -> Rational
```
soutaro, line 131: > __[Goodcheck]__ Prefer `boolish` over `bool` for method arguments and block return values ([view](https://sider.review/gh/repos/174796313/pulls/549#issue-5268998))
>
> <details>
>
> See the doc below:
> https://github.com/ruby/rbs/blob/78d04a2db0f1c4925d2b13c2939868edf9465d6c/docs/syntax.md#bool-or-boolish
>
> | Rule |
> | ---- |
> | `rbs.prefer_boolish` |
>
> You can [close this issue](https://sider.review/gh/repos/174796313/pulls/549#issue-5268998) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

I closed this issue, because:
- When you strictly want `true | false`.

core/kernel.rbs
```diff
@@ -54,7 +54,7 @@ module Kernel : BasicObject
   # try { "hello" }      #=> "hello"
   # try do "hello" end   #=> "hello"
   # ```
-  def block_given?: () -> bool
+  def self?.block_given?
```
soutaro, line 57: Resolved the conflict caused by deletion of `iterator?` #569

PR #556, opened 2020-12-28
Migrate to test-unit to make test run on Ruby 3.1

soutaro:
Minitest doesn't install on Ruby 3.1 (RBS is testing with nightly-master too). So, we are moving to test-unit today. 💪


soutaro:
Minitest now allows installing on Ruby 3.1. https://github.com/seattlerb/minitest/pull/862
But, we had finished migration already...

PR #557, opened 2020-12-28
Object#=~ has been deprecated since ruby 2.6

nobu:


soutaro:
We need to skip the test because Ruby 3.1 removes the (deprecated) method finally.

nobu:
> It still exists but will be deleted. (right?)

Still a few gems used in bundled gems use it, so we couldn't delete it yet, and in 3.0.
By updating those gems or removing dependencies on them, it is going to be removed.

PR #568, opened 2021-01-03
Fix incorrect Module#class_eval type

pocke:
`Module#class_eval` doesn't take any arguments in Ruby,
but the RBS says "it takes one argument" mistakenly.
So this patch fixes the incorrect `Module#class_eval` type.


soutaro (approved):
🙏

PR #571, opened 2021-01-08
Improve signatures of IO#<<, IO#advise, IO#autoclose=, IO#autoclose?

ybiquitous:
This change does the following for some methods of the `IO` class:

- add missing documents
- add tests
- make some types more accurate

See the API doc: <https://ruby-doc.org/core-3.0.0/IO.html>

### DIFF
--- a/core/io.rbs
+++ b/core/io.rbs
@@ -107,14 +107,85 @@ class IO < Object
 
   include Enumerable[String]
 
-  def <<: (untyped arg0) -> self
+  # String Output---Writes *obj* to *ios*. *obj* will be converted to a string
+  # using `to_s`.
+  #
+  #     $stdout << "Hello " << "world!\n"
+  #
+  # *produces:*
+  #
+  #     Hello world!
+  #
+  def <<: (_ToS obj) -> self
 
-  def advise: (Symbol arg0, ?Integer offset, ?Integer len) -> NilClass
+  # Announce an intention to access data from the current file in a specific
+  # pattern. On platforms that do not support the *posix_fadvise(2)* system call,
+  # this method is a no-op.
+  #
+  # *advice* is one of the following symbols:
+  #
+  # :normal
+  # :   No advice to give; the default assumption for an open file.
+  # :sequential
+  # :   The data will be accessed sequentially with lower offsets read before
+  #     higher ones.
+  # :random
+  # :   The data will be accessed in random order.
+  # :willneed
+  # :   The data will be accessed in the near future.
+  # :dontneed
+  # :   The data will not be accessed in the near future.
+  # :noreuse
+  # :   The data will only be accessed once.
+  #
+  #
+  # The semantics of a piece of advice are platform-dependent. See *man 2
+  # posix_fadvise* for details.
+  #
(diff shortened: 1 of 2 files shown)


core/io.rbs
```diff
@@ -107,14 +107,85 @@ class IO < Object
...
+  #     IO.for_fd(f.fileno)
+  #     # ...
+  #     f.gets # may cause Errno::EBADF
+  #
+  #     f = open("/dev/null")
+  #     IO.for_fd(f.fileno).autoclose = false
+  #     # ...
+  #     f.gets # won't cause Errno::EBADF
+  #
+  def autoclose=: (bool) -> bool
```
ybiquitous, line 184: [note] Note that the argument type changes from `boolish` to `bool`. Because `io.autoclose = "yes"` returns `"yes"` for example.

We can confirm this behavior as follows:

```ruby
  def test_autoclose=
    IO.open(IO.sysopen(__FILE__)) do |io|
      assert_send_type "(String) -> bool",
                       io, :autoclose=, "yes"
    end
  end
```

```console
$ bundle exec ruby test/stdlib/IO_test.rb
...
Failure: test_autoclose=(IOInstanceTest):
  Call trace does not match with given method type: #<struct RBS::Test::CallTrace method_name=:autoclose=, method_call=#<RBS::Test::ArgumentsReturn:0x00007fa2aea9ac38 @arguments=["yes"], @exit_value="yes", @exit_type=:return>, block_calls=[], block_given=false>.
  <["[IOautoclose=] ReturnTypeError: expected `bool` but returns `\"yes\"`"]> was expected to be empty.
/Users/masafumi.koba/git/ybiquitous/rbs/test/stdlib/test_helper.rb:240:in `assert_send_type'
test/stdlib/IO_test.rb:129:in `block in test_autoclose='
     126:
     127:   def test_autoclose=
     128:     IO.open(IO.sysopen(__FILE__)) do |io|
  => 129:       assert_send_type "(String) -> bool",
     130:                        io, :autoclose=, "yes"
     131:     end
     132:   end
test/stdlib/IO_test.rb:128:in `open'
test/stdlib/IO_test.rb:128:in `test_autoclose='
...
```

ybiquitous: I believe the following Sider issue is closeable:
https://sider.review/gh/repos/174796313/pulls/571#issue-5289513

ybiquitous, line 184: > __[Goodcheck]__ Prefer `boolish` over `bool` for method arguments and block return values ([view](https://sider.review/gh/repos/174796313/pulls/571#issue-5289513))
>
> <details>
>
> See the doc below:
> https://github.com/ruby/rbs/blob/78d04a2db0f1c4925d2b13c2939868edf9465d6c/docs/syntax.md#bool-or-boolish
>
> | Rule |
> | ---- |
> | `rbs.prefer_boolish` |
>
> You can [close this issue](https://sider.review/gh/repos/174796313/pulls/571#issue-5289513) if no need to fix it. [Learn more](https://help.sider.review/getting-started/working-with-issues).
>
> </details>

See https://github.com/ruby/rbs/pull/571#discussion_r553676299

soutaro:
Thanks! 🎉

PR #575, opened 2021-01-15
Fix non-existent `stdlib/builtin`

ybiquitous:
The `stdlib/builtin` directory moved to the `core` directory on PR #405.
This change fixes `stdlib/builtin` in a Rake task code and documents.


soutaro (approved):
Nice catch! 👏

PR #576, opened 2021-01-16
Add signature of FileUtils

ybiquitous:
This change adds signatures for the `FileUtils` module.

Note: `FileUtils` has some sub-modules like `FileUtils::Verbose`, but this change is the first step and does not include such sub-modules.

See also:
- https://ruby-doc.org/stdlib-2.7.2/libdoc/fileutils/rdoc/FileUtils.html
- https://github.com/ruby/fileutils

### DIFF
--- a/stdlib/fileutils/0/fileutils.rbs
+++ b/stdlib/fileutils/0/fileutils.rbs
@@ -1,0 +1,569 @@
+# # fileutils.rb
+#
+# Copyright (c) 2000-2007 Minero Aoki
+#
+# This program is free software. You can distribute/modify this program under
+# the same terms of ruby.
+#
+# ## module FileUtils
+#
+# Namespace for several file utility methods for copying, moving, removing, etc.
+#
+# ### Module Functions
+#
+#     require 'fileutils'
+#
+#     FileUtils.cd(dir, **options)
+#     FileUtils.cd(dir, **options) {|dir| block }
+#     FileUtils.pwd()
+#     FileUtils.mkdir(dir, **options)
+#     FileUtils.mkdir(list, **options)
+#     FileUtils.mkdir_p(dir, **options)
+#     FileUtils.mkdir_p(list, **options)
+#     FileUtils.rmdir(dir, **options)
+#     FileUtils.rmdir(list, **options)
+#     FileUtils.ln(target, link, **options)
+#     FileUtils.ln(targets, dir, **options)
+#     FileUtils.ln_s(target, link, **options)
+#     FileUtils.ln_s(targets, dir, **options)
+#     FileUtils.ln_sf(target, link, **options)
+#     FileUtils.cp(src, dest, **options)
+#     FileUtils.cp(list, dir, **options)
+#     FileUtils.cp_r(src, dest, **options)
+#     FileUtils.cp_r(list, dir, **options)
+#     FileUtils.mv(src, dest, **options)
+#     FileUtils.mv(list, dir, **options)
+#     FileUtils.rm(list, **options)
+#     FileUtils.rm_r(list, **options)
+#     FileUtils.rm_rf(list, **options)
+#     FileUtils.install(src, dest, **options)
+#     FileUtils.chmod(mode, list, **options)
+#     FileUtils.chmod_R(mode, list, **options)
+#     FileUtils.chown(user, group, list, **options)
+#     FileUtils.chown_R(user, group, list, **options)
+#     FileUtils.touch(list, **options)
+#
+# Possible `options` are:
+#
+# `:force`
+# :   forced operation (rewrite files if exist, remove directories if not empty,
+#     etc.);
+# `:verbose`
+# :   print command to be run, in bash syntax, before performing it;
+# `:preserve`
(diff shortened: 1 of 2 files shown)


stdlib/fileutils/0/fileutils.rbs
```diff
@@ -0,0 +1,569 @@
...
+#
+# This module has all methods of FileUtils module, but never changes
+# files/directories.  This equates to passing the `:noop` and `:verbose` flags
+# to methods in FileUtils.
+module FileUtils
+  VERSION: String
+
+  type mode = Integer | String
+
+  type pathlist = String | Array[String]
```
pocke, line 102: I guess it also accepts `_ToPath` and `string`.
https://github.com/ruby/rbs/blob/e7504e60517f3a4484794e4b0fa3d4b0b0641ae5/core/builtin.rbs#L25-L27

https://github.com/ruby/rbs/blob/e7504e60517f3a4484794e4b0fa3d4b0b0641ae5/core/builtin.rbs#L64


Because most `File` methods accept them.

stdlib/fileutils/0/fileutils.rbs
```diff
@@ -0,0 +1,569 @@
...
+  #
+  #     FileUtils.cd('/')  # change directory
+  #
+  #     FileUtils.cd('/', verbose: true)   # change directory and report it
+  #
+  #     FileUtils.cd('/') do  # change directory
+  #       # ...               # do something
+  #     end                   # return to original directory
+  #
+  def self.cd: (String dir, ?verbose: boolish) -> void
```
pocke, line 117: Most (or all?) FileUtils methods are `module_function`. So let use `self?` instead of `self`.

pocke (commented):
Great improvement!

I wrote a few comments, but the other parts look good.

stdlib/fileutils/0/fileutils.rbs
ybiquitous: Thanks, it is a nice idea!

Also, it seems good to add the following type to `core/builtin.rbs`, what do you think?

```ruby
type path = string | _ToPath
```

Anyway, I will improve the signatures via new following types: 👍 

```
module FileUtils
  type path = string | _ToPath
  type pathlist = path | Array[path]
end
```

stdlib/fileutils/0/fileutils.rbs
ybiquitous: Thanks! I'll improve the signatures and tests! 💪

stdlib/fileutils/0/fileutils.rbs
ybiquitous: Fixed via 929ecc8

stdlib/fileutils/0/fileutils.rbs
ybiquitous: BTW, it might be nice if we could write it as follows... 🤔 

```diff
-alias self.chdir self.cd
-alias chdir cd
+alias self?.chdir self?.cd
```

ybiquitous: Fixed via ca1d180

ybiquitous:
@pocke Thank you so much for your advice! I've addressed your comments. 🙆‍♂️

stdlib/fileutils/0/fileutils.rbs
pocke: > Also, it seems good to add the following type to core/builtin.rbs, what do you think?



Personally I think the `path` type alias should not be defined to the top level. `path` alias is a domain-specific type, unlike `string` and so on. So I think it should be defined under a namespace, and I think `File::path` is appropriate.


And we need to reconsider the other type aliases and interfaces, such as `_ToPath`, `_Reader` and so on.

So I'd like to keep them in this pull request, and move them under a namespace by another pull request.

stdlib/fileutils/0/fileutils.rbs
pocke: > BTW, it might be nice if we could write it as follows... thinking



True. I heard the unexistence of `self?` alias is intentional by @soutaro because of rare.

But maybe we can add `self?` alias now. Previously I didn't know there are many singleton-instance aliases like that, so I thought `self?` is unnecessary. But we found demands for singleton-instance aliases.

stdlib/fileutils/0/fileutils.rbs
ybiquitous: That makes sense. 👍

stdlib/fileutils/0/fileutils.rbs
ybiquitous: > we found demands for singleton-instance aliases.

Yes, I think so. We could reduce duplications in a case like `FileUtils`.

soutaro: @ybiquitous Thanks! I don't think adding new syntax for aliasing module functions makes much sense for now. Let me know if you find this is a frequently seen pattern.

ybiquitous: OK, thanks for your comment 👍

soutaro (approved):
Thank you @ybiquitous ! 👏

PR #581, opened 2021-01-26
Renew test sample code

masahito1997:
We can get test sample code by executing of `generate:stdlib_test[class]` rake task and it is indicated in stdlib.md.

But, the code indicated in stdlib.md was different from the one generated by rake task.
So, I modified it.


soutaro (approved):
Thanks! 🎉

PR #587, opened 2021-02-03
Add signatures for URI::RFC2396_Parser

ybiquitous:
This change adds signatures for the `URI::RFC2396_Parser` class.

See also:
- https://docs.ruby-lang.org/en/3.0.0/URI/RFC2396_Parser.html
- https://github.com/ruby/uri/blob/v0.10.1/lib/uri/rfc2396_parser.rb


soutaro (approved):
🪂

ybiquitous:
Thanks for the merge! 😄

PR #588, opened 2021-02-07
Restrict  `refute_send_type`  to check method call doesn't match with type

pocke:
This pull request adds a more strict check with `refute_send_type`.

# Problem

I think `refute_send_type` assertion isn't enough to check the type mismatch. 

Currently, `refute_send_type` checks that the method call raises an error, and the method call is matched with the `method_type` that is specified as an argument of `refute_send_type`. But the RBS, which is exposed to the end-users, doesn't affect `refute_send_type`. So it checks nothing about the RBS, but it checks the implementation.

I think `refute_send_type` should check the RBS, so the current behavior is not enough as a test assertion.

# Solution

Check that the RBS declines the method call.

ref: https://github.com/ruby/rbs/pull/586#discussion_r571572265


core/array.rbs
```diff
@@ -1736,7 +1736,7 @@ class Array[unchecked out Elem] < Object
   # See also Enumerable#sort_by.
   #
   def sort: () -> ::Array[Elem]
-          | () { (Elem a, Elem b) -> ::Integer? } -> ::Array[Elem]
+          | () { (Elem a, Elem b) -> ::Integer } -> ::Array[Elem]
```
pocke, line 1739: `Array#sort`'s block doesn't accept nil as the returned value actually, so I removed nil.


It is a more strict type, but it can be problematic with `<=>` method.
For example, `Integer#<=>` may return a `nil` when the values are incomparable. So the returned type is `Integer?`.
https://github.com/ruby/rbs/blob/26baa08e184d1d27c18a34e6683493db42f5fe77/core/integer.rbs#L108-L115

So `Array#sort` will not become to accept the following code by this change.

```ruby
[1,2,3].sort { |a, b| a <=> b }
```


But personally I think `Integer` is better than `Integer?` as the returned value of `sort`'s block. `Array#sort` always raises when the block returns nil.
And we can improves `Integer#<=>` type, like `(Integer) => Integer | (Numeric) -> Integer?`. The improvement will resolve the `Array#sort`'s problem.

soutaro: This makes sense. 👍 

(I made it `Integer?` because of the return type of `<=>`.)

soutaro:
@pocke I'm not totally sure if this is the best way. But, I agree that type checking the calls with respect to RBS makes sense. Thanks!

PR #589, opened 2021-02-08
Add gem rbs collection manager

pocke:
This PR introduces `rbs collection` subcommand to manage 3rd party RBSs.

# Usage

See the added document, `collection.md`.

# Example

I tried introducing `rbs collection` to my repository, rbs_rails. https://github.com/pocke/rbs_rails/pull/187
It works well :smile: 

# Next Steps

This PR is the first step, so there are remaining works. I'll do them in other PRs.

## Support this feature by TypeProf and Steep

All rbs sub-command supports this feature, for example, `rbs validate` automatically loads RBSs from `rbs_collection.yaml` if the file exists. But TypeProf and Steep don't support it.

I'll do it soon :rocket: I guess it is an easy task.

## Write RBS dependencies

Currently `rbs collection` command resolves dependencies with `Gemfile.lock`. So it has a limitation that it can't resolve stdlib dependencies.
 
For example, `pathname` is a default gem, so we don't write `add_runtime_dependency 'pathname'` in order to use the library in a gem. Then a Gemfile.lock doesn't contain `pathname` even if the dependency actually depends on pathname library.

We need another approach to resolve stdlib dependencies.
We, soutaro, mame, and I, discussed the topic in closed meetings. We are considering introducing a meta file to describe stdlib dependencies for RBS repository, like `manifest.yaml`.

For example:

```yaml
# ruby/gem_rbs_collection/gems/activesupport/6.0/manifest.yaml
dependencies:
  - name: pathname
  - name: logger
    version: 1.4 # Can specify a version (optional)
  - name: mutex_m
  # ...
```

## Update individual gems

Currenly `rbs collection update` command only supports updating all gems. But I can easily imagine it is too loose. We'll want to update specific gems like we usually do `bundle update rails csv nokogiri` in Bundler.
I'll allow `rbs collection update` to take gem names as arguments, like `rbs collection update rails csv nokogiri`

## Without Gemfile.lock mode

Currently `rbs collection` command assumes Gemfile.lock existence. I guess it may need no Gemfile.lock mode for a project without Bundler or Gemfile.lock is not committed.

# Design document

I wrote a design document at the beginning of this PR. I think there are many differences since the doc. It looks obsolete.
https://pocke.kibe.la/shared/entries/3665a009-e229-4e56-a92b-c0743f7a49fe


lib/rbs/environment_loader.rb
```diff
@@ -47,6 +47,16 @@ def add(path: nil, library: nil, version: nil)
       end
     end
 
+    def add_collection(collection_config)
```
soutaro, line 50: 👍

PR #592, opened 2021-02-10
Add `#to_json` signatures to core classes

ybiquitous:
This change adds `#to_json` method signatures of the `json` gem to core classes such as `String`, `Hash`, etc.

Note: `bin/query-rdoc` and `bin/bin/annotate-with-rdoc` does not work for these new methods,
so I manually copy the methods' RDoc from the `json` gem source code.
For example, see <https://github.com/flori/json/blob/v2.5.1/ext/json/ext/generator/generator.c#L411-L418>

```console
$ bin/query-rdoc 'Hash#to_json'
...
Finding instance method Hash # to_json ...
Nothing to print; failed to query the document...

$ bin/annotate-with-rdoc stdlib/json/0/json.rbs
...
  Importing documentation for Array...
    Processing to_json...
  👺 No document found for Array#to_json
Writing stdlib/json/0/json.rbs...
```

### DIFF
--- a/stdlib/json/0/json.rbs
+++ b/stdlib/json/0/json.rbs
@@ -339,3 +339,67 @@ JSON::VERSION_BUILD: Integer
 JSON::VERSION_MAJOR: Integer
 
 JSON::VERSION_MINOR: Integer
+
+class Object
+  # Converts this object to a string (calling #to_s), converts
+  # it to a JSON string, and returns the result. This is a fallback, if no
+  # special method #to_json was defined for some object.
+  #
+  def to_json: (?JSON::State state) -> String
+end
+
+class NilClass
+  # Returns a JSON string for nil: 'null'.
+  #
+  def to_json: (?JSON::State state) -> String
+end
+
+class TrueClass
+  # Returns a JSON string for true: 'true'.
+  #
+  def to_json: (?JSON::State state) -> String
+end
+
+class FalseClass
+  # Returns a JSON string for false: 'false'.
+  #
+  def to_json: (?JSON::State state) -> String
+end
+
+class String
+  # This string should be encoded with UTF-8 A call to this method
+  # returns a JSON string encoded with UTF16 big endian characters as
+  # \u????.
+  #
+  def to_json: (?JSON::State state) -> String
+end
+
+class Integer
+  # Returns a JSON string representation for this Integer number.
+  #
+  def to_json: (?JSON::State state) -> String
+end
+
+class Float
+  # Returns a JSON string representation for this Float number.
+  #
+  def to_json: (?JSON::State state) -> String
+end
+
+class Hash[unchecked out K, unchecked out V]
+  # Returns a JSON string containing a JSON object, that is generated from
+  # this Hash instance.
+  # _state_ is a JSON::State object, that can also be used to configure the
+  # produced JSON string output further.
+  #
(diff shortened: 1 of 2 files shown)


test/stdlib/json/JSON_test.rb
```diff
@@ -238,4 +238,58 @@ def test_unparse
     assert_send_type "(ToJson) -> String", MyJSON.new, :unparse, ToJson.new
     assert_send_type "(ToJson, indent: String) -> String", MyJSON.new, :unparse, ToJson.new, { indent: "\t" }
   end
+
+  def test_to_json_with_object
+    assert_send_type "() -> String", Object.new, :to_json
+    assert_send_type "(JSON::State) -> String",  Object.new, :to_json, JSON::State.new
+    assert_send_type "(nil) -> String", Object.new, :to_json, nil
+  end
```
ybiquitous, line 246: [IMO] I prefer the multiline style to reduce cognitive load, but I follow the existing style in this file.

```diff
-assert_send_type "() -> String", Object.new, :to_json
+assert_send_type "() -> String",
+                 Object.new, :to_json
```

stdlib/json/0/json.rbs
```diff
@@ -339,3 +339,67 @@ JSON::VERSION_BUILD: Integer
...
 
 JSON::VERSION_MINOR: Integer
+
+class Object
+  # Converts this object to a string (calling #to_s), converts
+  # it to a JSON string, and returns the result. This is a fallback, if no
+  # special method #to_json was defined for some object.
+  #
+  def to_json: (?JSON::State state) -> String
+end
```
ybiquitous, line 349: [note] The same interface as `_ToJson`:

https://github.com/ruby/rbs/blob/f965869c4b4224a796ebc9f34cfc4b07bde82cfc/stdlib/json/0/json.rbs#L1-L3

ybiquitous:
Maybe we could remove these signatures of `to_json` in `sig/*.rbs` as well (I can do it if wanted).

```console
$ git grep to_json sig
sig/annotation.rbs:23:      def to_json: (*untyped) -> String
sig/comment.rbs:21:      def to_json: (*untyped) -> String
sig/location.rbs:44:    def to_json: (*untyped args) -> untyped
sig/method_types.rbs:12:    def to_json: (*untyped) -> String
sig/polyfill.rbs:2:  def to_json: (*untyped) -> String
sig/typename.rbs:37:    def to_json: (*untyped a) -> untyped
sig/types.rbs:35:      def to_json: (*untyped) -> String
sig/types.rbs:297:      def to_json: (*untyped) -> String
sig/types.rbs:333:      def to_json: (*untyped) -> String
sig/util.rbs:11:    def to_json: (*untyped) -> String
```

test/stdlib/json/JSON_test.rb
soutaro: I prefer the multiline style too, but don't have strong opinion to fix all of the existing test code.

soutaro (approved):
Thanks. 🎃

soutaro:
@ybiquitous I want to keep the existing `to_json` declarations in `sig/*.rbs` files, because they have custom implementations. (We can rewrite them using `JSON::State` type, but I don't think it makes much sense.)

ybiquitous:
@soutaro Make sense. Thanks for your explanation. 👌

PR #593, opened 2021-02-10
Improve <=> types

pocke:
* All `<=>` accepts `untyped`
* Do not optional for specific types

ref: https://github.com/ruby/rbs/pull/588#discussion_r571589397

### DIFF
--- a/core/file.rbs
+++ b/core/file.rbs
@@ -985,7 +985,8 @@ class File::Stat < Object
   include Comparable
 
   def initialize: (String file) -> Object
-  def <=>: (File::Stat other) -> Integer?
+  def <=>: (File::Stat other) -> Integer
+         | (untyped) -> nil
 
   def atime: () -> Time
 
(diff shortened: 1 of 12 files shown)


core/file.rbs
```diff
@@ -985,7 +985,8 @@ class File::Stat < Object
   include Comparable
 
   def initialize: (String file) -> Object
-  def <=>: (File::Stat other) -> Integer?
+  def <=>: (File::Stat other) -> Integer
```
pocke, line 988: https://github.com/ruby/ruby/blob/947d93b715436b13eefa39f87737bdad3c1f870a/file.c#L538-L553

core/integer.rbs
```diff
@@ -112,7 +112,8 @@ class Integer < Numeric
   #
   # `nil` is returned if the two values are incomparable.
   #
-  def <=>: (Numeric) -> Integer?
+  def <=>: (Integer | Rational) -> Integer
```
pocke, line 115: `int <=> float` may return a nil, for example: `42 <=> Float::NAN # => nil`

soutaro (approved):
🛸

PR #602, opened 2021-02-17
Make racc name customizable by RACC environment variable

nobu:
Executable file name may be transformed like "racc3.1".


soutaro (approved):
🙏

PR #605, opened 2021-02-19
Add signatures for `Gem` module in `rubygems` gem

ybiquitous:
This change adds signatures of the `Gem` module in the `rubygems` gem.

Note that this does not include signatures of `Gem::*` classes or modules like `Gem::Version`.
(because it would be a too big PR)

See also:
- https://github.com/rubygems/rubygems/blob/v3.2.11/lib/rubygems.rb
- https://ruby-doc.org/stdlib-3.0.0/libdoc/rubygems/rdoc/Gem.html


test/stdlib/rubygems/Gem_test.rb
```diff
@@ -0,0 +1,597 @@
...
+                      Gem, :load_plugins
+  end
+
+  def test_load_yaml
+    assert_send_type  "() -> bool?",
+                      Gem, :load_yaml
+  end
+
+  def test_loaded_specs
+    omit "due to Bundler::StubSpecification returns"
```
ybiquitous, line 310: [note] `Gem.loaded_specs` returns a `Hash` including `Bundler::StubSpecification`, but this test fails because `Bundler::StubSpecification` is **not** a `Gem::BasicSpecification`.

```console
$ bundle exec ruby -e 'pp Bundler::StubSpecification.ancestors'
[Bundler::StubSpecification,
 Bundler::RemoteSpecification,
 Comparable,
 Bundler::MatchPlatform,
 Bundler::GemHelpers,
 Object,
 Kernel,
 BasicObject]
```

Please tell me a better way if you know it. 🙏

.github/workflows/ruby.yml
```diff
@@ -27,10 +27,10 @@ jobs:
       run: |
         apt-get update
         apt-get install -y libdb-dev
-    - name: Install bundler
+    - name: Update rubygems & bundler
       run: |
         ruby -v
-        gem install bundler
+        gem update --system
```
ybiquitous, line 33: [note] The test for `Gem.plugindir` fails with Ruby 2.7 because Ruby 2.7 includes RubyGems 3.1 by default.

> Error: test_plugindir(GemSingletonTest): NoMethodError: undefined method `plugindir' for Gem:Module

https://github.com/ruby/rbs/pull/605/checks?check_run_id=1937717810#step:8:1248

`.plugindir` has been available since RubyGems 3.2 (see <https://github.com/rubygems/rubygems/pull/3163>)

So, I've updated this CI configuration to use the latest RubyGems.
This way may be correct, but it seems to me that this CI update should be done on another PR.

Can you provide me feedback, please?

soutaro: Installing the latest rubygems looks reasonable. 👍

stdlib/rubygems/0/rubygems.rbs
```diff
@@ -0,0 +1,711 @@
+# TODO: Is this defined as a built-in type?
+interface _HashLike[K, V]
+  def each_pair: () { ([ K, V ]) -> untyped } -> self
+end
```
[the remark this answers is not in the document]
soutaro: I think it's better to put the type inside `Gem` module, not exposing as a top level type.

test/stdlib/rubygems/Gem_test.rb
soutaro: Looks good to skip testing. 🎿

soutaro (approved):
Thank you! 💎

stdlib/rubygems/0/rubygems.rbs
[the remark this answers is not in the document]
ybiquitous: Thanks for the advice. I'll fix it with another PR. 👍

ybiquitous:
@soutaro Thank you so much for reviewing and merging it! I'll add more signatures for `rubygems` with another PR. 😸

PR #606, opened 2021-02-23
Allow trailing comma for Record and Tuple types

pocke:
This pull request allows trailing comma for Record and Tuple types.

# Problem

Currently RBS doesn't allow putting a trailing comma after the last item of Record and Tuple.

```rbs
# RBS

type t = { a: 42, } # error
type u = [42, ] # error
```

The use-case is a multi-line Record type. 

```rbs
# RBS

type t = {
  item1: 42,
  item2: 43,
  item3: 44, # This trailing comma introduces an error
}
```

# Solution

Allow trailing comma for them.

# Note

RBS allows a trailing comma for method parameters with or without this pull request.

```rbs
# RBS: OK
class C
  def foo: (untyped,) => untyped
end
```

RBS doesn't allow trailing comma for other places such as type parameters  even if applying this pull request. For example:

```rbs
# RBS: Error

class C[X,] < D[Y,]
  include M[Z,]
end
```

I don't have any use-case of trailing comma for them, so this pull request doesn't change anything for them.


soutaro (approved):
🇰🇲

soutaro:
I agree that allowing trailing comma is better. 👍 
(It's in fact not intentional to allow trailing comma in method type arguments...)

PR #610, opened 2021-02-25
Add signatures for Gem::Version

ybiquitous:
This change adds signatures for the `Gem::Version` class.

Follow-up of PR #605

See also:
- https://github.com/rubygems/rubygems/blob/v3.2.11/lib/rubygems/version.rb
- https://ruby-doc.org/stdlib-3.0.0/libdoc/rubygems/rdoc/Gem/Version.html

### DIFF
--- a/stdlib/rubygems/0/rubygems.rbs
+++ b/stdlib/rubygems/0/rubygems.rbs
@@ -1,8 +1,3 @@
-# TODO: Is this defined as a built-in type?
-interface _HashLike[K, V]
-  def each_pair: () { ([ K, V ]) -> untyped } -> self
-end
-
 # RubyGems is the Ruby standard for publishing and managing third party
 # libraries.
 #
@@ -99,6 +94,10 @@ end
 # -The RubyGems Team
 #
 module Gem
+  interface _HashLike[K, V]
+    def each_pair: () { ([ K, V ]) -> untyped } -> self
+  end
+
   DEFAULT_HOST: String
 
   GEM_DEP_FILES: Array[String]
--- a/stdlib/rubygems/0/version.rbs
+++ b/stdlib/rubygems/0/version.rbs
@@ -1,3 +1,228 @@
-class Gem::Version
-  # TODO: Add sinatures...
+module Gem
+  # The Version class processes string versions into comparable values. A version
+  # string should normally be a series of numbers separated by periods. Each part
+  # (digits separated by periods) is considered its own number, and these are used
+  # for sorting. So for instance, 3.10 sorts higher than 3.2 because ten is
+  # greater than two.
+  #
+  # If any part contains letters (currently only a-z are supported) then that
+  # version is considered prerelease. Versions with a prerelease part in the Nth
+  # part sort less than versions with N-1 parts. Prerelease parts are sorted
+  # alphabetically using the normal Ruby string sorting rules. If a prerelease
+  # part contains both letters and numbers, it will be broken into multiple parts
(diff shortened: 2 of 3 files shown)


stdlib/rubygems/0/rubygems.rbs
```diff
@@ -99,6 +94,10 @@ end
 # -The RubyGems Team
 #
 module Gem
+  interface _HashLike[K, V]
+    def each_pair: () { ([ K, V ]) -> untyped } -> self
+  end
+
```
ybiquitous, line 100: [note] According to this advice: https://github.com/ruby/rbs/pull/605#discussion_r581879772

stdlib/rubygems/0/version.rbs
```diff
@@ -1,3 +1,228 @@
...
+  #
+  # For the last example, single-digit versions are automatically extended with a
+  # zero to give a sensible result.
+  #
+  class Version
+    include Comparable
+
+    # True if the `version` string matches RubyGems' requirements.
+    #
+    def self.correct?: (_ToS version) -> bool
```
ybiquitous, line 164: [note] It seems that `nil` should not be received because of the deprecation below:

```console
$ ruby -e 'Gem::Version.correct?(nil)'
nil versions are discouraged and will be deprecated in Rubygems 4
```

stdlib/rubygems/0/version.rbs
```diff
@@ -1,3 +1,228 @@
...
+
+    # Dump only the raw version string, not the complete object. It's a string for
+    # backwards (RubyGems 1.3.5 and earlier) compatibility.
+    #
+    def marshal_dump: () -> Array[String]
+
+    # Load custom marshal format. It's a string for backwards (RubyGems 1.3.5 and
+    # earlier) compatibility.
+    #
+    def marshal_load: (Array[String] array) -> void
```
ybiquitous, line 211: [note] `#marshal_load` returns `nil` actually, but I think it should return `void` because `#marshal_load` calls internally `#initialize`:

https://github.com/rubygems/rubygems/blob/v3.2.11/lib/rubygems/version.rb#L273-L275

pocke:
I guess we can remove `Gem::Version` from polyfill.rbs. https://github.com/ruby/rbs/blob/448b67a1c458fe753e30355c6d3515f402aa87ec/sig/polyfill.rbs#L5-L18
The polyfill is for type checking for RBS itself. you can try using it with Steep and `./steep/Gemfile`.

ybiquitous:
@pocke Thanks for the comment. Should we add `signature` as below to remove the polyfill (not add `library "rubygems"`)?

```diff
 target :lib do
   signature "sig"
+  signature "stdlib/rubygems"
   check "lib"
   ignore "lib/rbs/parser.rb"
```

Because the published version of `rbs` is used in `steep/Gemfile`:

https://github.com/ruby/rbs/blob/448b67a1c458fe753e30355c6d3515f402aa87ec/steep/Gemfile.lock#L28-L29

stdlib/rubygems/0/version.rbs
soutaro: Ah! `_ToS` cannot reject `nil`...

soutaro (approved):
💎

stdlib/rubygems/0/version.rbs
ybiquitous: Oh, I missed it… 😓 

I want to improve this, but can we do it? (If we can, I will create a new PR)
It's not so bad as-is, though. 😅

PR #618, opened 2021-03-04
Fix RBS for Thread.new

ybiquitous:
This change fixes RBS for the `Thread.new` method (actually `Thread#initialize`).
Also, this adds a missing test for `Thread` RBS.

### DIFF
--- a/core/thread.rbs
+++ b/core/thread.rbs
@@ -287,7 +287,21 @@ class Thread < Object
 
   def group: () -> ThreadGroup?
 
-  def initialize: (*untyped args) -> Thread
+  # Creates a new thread executing the given block.
+  #
+  # Any `args` given to ::new will be passed to the block:
+  #
+  #     arr = []
+  #     a, b, c = 1, 2, 3
+  #     Thread.new(a,b,c) { |d,e,f| arr << d << e << f }.join
+  #     arr #=> [1, 2, 3]
+  #
+  # A ThreadError exception is raised if ::new is called without a block.
+  #
+  # If you're going to subclass Thread, be sure to call super in your `initialize`
+  # method, otherwise a ThreadError will be raised.
+  def initialize: () { () -> untyped } -> void
+                | (*untyped) { (*untyped) -> untyped } -> void
 
   # The calling thread will suspend execution and run this `thr` .
   #
--- a/core/thread.rbs
+++ b/core/thread.rbs
@@ -300,8 +300,7 @@ class Thread < Object
   #
   # If you're going to subclass Thread, be sure to call super in your `initialize`
   # method, otherwise a ThreadError will be raised.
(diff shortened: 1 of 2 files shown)


test/stdlib/Thread_test.rb
```diff
@@ -0,0 +1,14 @@
+require_relative "test_helper"
```
ybiquitous, line 1: [note] I've created this file via `bundle exec rake generate:stdlib_test'[Thread]'`.

core/thread.rbs
```diff
@@ -287,7 +287,21 @@ class Thread < Object
...
+  #     a, b, c = 1, 2, 3
+  #     Thread.new(a,b,c) { |d,e,f| arr << d << e << f }.join
+  #     arr #=> [1, 2, 3]
+  #
+  # A ThreadError exception is raised if ::new is called without a block.
+  #
+  # If you're going to subclass Thread, be sure to call super in your `initialize`
+  # method, otherwise a ThreadError will be raised.
+  def initialize: () { () -> untyped } -> void
+                | (*untyped) { (*untyped) -> untyped } -> void
```
ybiquitous, line 304: [note] We might have a better way that can avoid `*untyped`... 🤔 
(these both `*untyped*` have actually the same types)

ybiquitous:
The API doc for `Thread.new` has a pattern of `new(*args, &proc)`, but such a pattern does not actually exist in the spec <https://github.com/ruby/ruby/blob/v3_0_0/spec/ruby/core/thread/new_spec.rb>... 🤔 

image

https://ruby-doc.org/core-3.0.0/Thread.html#new-method

ybiquitous:
https://github.com/ruby/rbs/pull/618#issuecomment-790486980

This seems a bug of the doc:
https://github.com/ruby/ruby/blob/446d000e1148d4dd1dc5e0d3a745dbc205457632/thread.c#L1002

ybiquitous:
Sorry, I misunderstood... 😓 
The `new(*args, &proc)` pattern is also valid. See the test case below:

https://github.com/ruby/rbs/pull/618/files#diff-09ffa05c1d5b9da899c4c67670c5c8ff87235b0feecdee92774865f607029d89R14-R16

core/thread.rbs
soutaro: Seems like we don't need the first overload?

And we would write `void` for the type of return value of the block.

ybiquitous: Sure. I’ve fixed it with a6633f5.

ybiquitous:
I will fix up this to one commit when approved. 💪

soutaro (approved):
🎉

ybiquitous:
@soutaro Oh, thank you for merging it!

Will "squash and merge" be performed from now on, not "merge"?
(I have no opinion, just a question)

PR #622, opened 2021-03-07
Remove needless Gem::Version polyfill

ybiquitous:
The `Gem::Version` signature has been added with PR #610.
I think we now can remove the polyfill of `Gem::Version`.

Follow-up of <https://github.com/ruby/rbs/pull/610#issuecomment-785609744>


soutaro (approved):
💎

PR #626, opened 2021-03-10
Add signature for Gem::Requirement

ybiquitous:
This change adds the signature for the `Gem::Requirement` class.

See also:
- https://github.com/rubygems/rubygems/blob/v3.2.14/lib/rubygems/requirement.rb
- https://ruby-doc.org/stdlib-3.0.0/libdoc/rubygems/rdoc/Gem/Requirement.html

### DIFF
--- a/stdlib/rubygems/0/requirement.rbs
+++ b/stdlib/rubygems/0/requirement.rbs
@@ -1,3 +1,85 @@
-class Gem::Requirement
-  # TODO: Add sinatures...
+module Gem
+  # A Requirement is a set of one or more version restrictions. It supports a few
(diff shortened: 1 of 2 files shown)


soutaro:
@ybiquitous Could you review yourself that if the tests assume some gems are installed? (See https://github.com/ruby/rbs/pull/629)

ybiquitous:
@soutaro Oh, sorry. I couldn't notice the failure. 🙇 

I think this PR's test has no problems because it does not include other gems than `rubygems`.

soutaro (approved):
💎

PR #628, opened 2021-03-11
adding + and - methods to numeric

### LINKED ISSUE #609: Type `::Numeric` does not have method `+`
I'm getting the following error when using `steep`, for an operation defined like:

```ruby
def add: (Numeric a, Numeric b) -> Numeric
```

This seems to be because [+ is defined as a method in all subclasses of numeric](https://github.com/ruby/rbs/blob/master/core/integer.rbs#L69). However, I'd argue that such a signature is legitimate, don't you agree? Or am I missing smth?

HoneyryderChuck:
Fixes #609 

Allows signatures of the kind:

```ruby
def test: (Numeric a, Numeric b) -> Numeric
```

Where `a+b` happens.

### DIFF
--- a/core/numeric.rbs
+++ b/core/numeric.rbs
@@ -82,9 +82,19 @@ class Numeric
   #
   def %: (Numeric) -> Numeric
 
+  # Performs addition: the class of the resulting object depends on the class of
+  # `numeric`.
+  #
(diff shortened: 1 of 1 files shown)


soutaro (approved):
Thanks! 🙏

PR #629, opened 2021-03-12
Fix test for Ruby CI

soutaro:


stdlib/rubygems/0/rubygems.rbs
```diff
@@ -338,7 +338,7 @@ module Gem
   #
   def self.find_unresolved_default_spec: (String path) -> Specification?
 
-  def self.finish_resolve: (?RequestSet request_set) -> Array[Specification]
+  def self.finish_resolve: (?RequestSet request_set) -> void
```
soutaro, line 341: Seems like the return value is not used, so `void` would be better here.

stdlib/rubygems/0/rubygems.rbs
```diff
@@ -417,7 +417,7 @@ module Gem
   #
   def self.marshal_version: () -> String
 
-  def self.needs: () { (RequestSet) -> untyped } -> Array[Specification]
+  def self.needs: () { (RequestSet) -> void } -> void
```
soutaro, line 420: Seems like the return value is not used, so `void` would be better here too.

test/stdlib/rubygems/Gem_test.rb
```diff
@@ -89,7 +89,7 @@ def test_datadir
     assert_send_type  "(String) -> nil",
                       Gem, :datadir, ""
     assert_send_type  "(String) -> String",
-                      Gem, :datadir, "rake"
+                      Gem, :datadir, "test-unit"
```
soutaro, line 92: I'm not very sure but `datadir` for `rake` fails on Ruby CI.

test/stdlib/rubygems/Gem_test.rb
```diff
@@ -235,10 +235,10 @@ def test_find_unresolved_default_spec
   end
 
   def test_finish_resolve
-    assert_send_type  "() -> Array[Gem::Specification]",
+    assert_send_type  "() -> Array[untyped]",
```
soutaro, line 238: The return type of the method is not `Gem::Specification`.

test/stdlib/rubygems/Gem_test.rb
```diff
@@ -563,8 +563,17 @@ def test_ui
   def test_use_gemdeps
     assert_send_type  "() -> nil",
                       Gem, :use_gemdeps
-    assert_send_type  "(String) -> Array[Gem::Specification]",
-                      Gem, :use_gemdeps, "-"
+
+    Dir.mktmpdir do |dir|
```
soutaro, line 567: Make `Gemfile` explicitly.

Issue #613, opened 2021-02-27
How to define different signatures for each supported ruby versions?


kachick:
https://github.com/ruby/rbs/blob/11d99dbe7de1cd9da6487ad0a153424b6452ee15/docs/CONTRIBUTING.md#the-target-version

The section describes as below

>The Target Version
The standard library signatures targets Ruby 2.7 for now.
The library code targets Ruby 2.6, 2.7, and 3.0.

How about to add/modify signatures for `core/` in each ruby versions?
e.g.) Ruby 3.0 added `Hash#except`, but it looks nothing in `core/hash.rbs`. Does `rbs` have any version guard feature?

soutaro:
The doc is wrong... The target right now is `3.0`.

We don't have a plan to introduce something like version guard yet.

kachick:
I see! So I close this issue. And I might try some PRs around the doc and `hash.rbs` :)

PR #631, opened 2021-03-13
Add Hash#except signature

kachick:
ref: https://github.com/ruby/rbs/issues/613

I have tried to add missing `Hash#except` 😄 

🤔 
---

I don't know why existing `Hash` tests does not use `assert_send_type` as `Array` tests. So I have added both test cases in this PR.

### DIFF
--- a/test/stdlib/Hash_test.rb
+++ b/test/stdlib/Hash_test.rb
@@ -186,6 +186,7 @@ class HashTest < StdlibTest
   end
 
   def test_except
+    omit_if(!Hash.method_defined?(:except))
     { a: 100, b: 200, c: 300 }.except(:a)
     { a: 100, b: 200, c: 300 }.except(:b, :c, :d)
   end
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks!

The reason why it doesn't have `assert_send_type` tests is that the test code was written before `assert_send_type` is introduced. It's great to have `HashInstanceTest` with the new API! 👏

test/stdlib/Hash_test.rb
```diff
@@ -185,6 +185,11 @@ def test_eql?
     { a: 2 }.eql?({ a: 1 })
   end
 
+  def test_except
+    { a: 100, b: 200, c: 300 }.except(:a)
```
soutaro, line 189: Could you add a guard like:

```ruby
skip if Hash.method_defined?(:except)
```

soutaro (changes requested):
Please add guards to make CI pass.

test/stdlib/Hash_test.rb
kachick: https://github.com/ruby/rbs/pull/631/commits/ec6d8651784e5e9acc19aefbd8255a00c7ba0da4

test-unit has `omit_if` feature, does it fit for this case?
Or it is a noisy reporting? 🤔

soutaro: Ah, it looks better! 🙆‍♂️

soutaro (approved):
💮

kachick:
Thank you!

PR #632, opened 2021-03-13
Fix Hash#slice signature

kachick:
```sh
ruby -v -e 'class Foo < Hash; end; p Foo.new.slice.class'
```

outputs

```
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin20]
Hash
```


soutaro (approved):
Nice catch! 👏

kachick:
Thank you!

PR #633, opened 2021-03-13
Update target ruby versions

kachick:
ref: https://github.com/ruby/rbs/issues/613#issuecomment-797332210

I got an answer from the issue.

>The doc is wrong... The target right now is 3.0.
We don't have a plan to introduce something like version guard yet.

So I think something like https://github.com/ruby/rbs/pull/631 needs this change. 🤔 

Note
---

To grep repository with `2.7` founds other diffs, but I don't have confident they should be updated or not.
I hope this change is minimum and enough.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -29,7 +29,7 @@
 ## The Target Version
 
 * The standard library signatures targets Ruby 3.0 for now.
-* The library code targets Ruby 3.0.
+* The library code targets Ruby 2.7 and 3.0.
 
 ## Stdlib Worksheet
 
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -28,8 +28,8 @@
 
 ## The Target Version
 
-* The standard library signatures targets Ruby 2.7 for now.
-* The library code targets Ruby 2.6, 2.7, and 3.0.
+* The standard library signatures targets Ruby 3.0 for now.
+* The library code targets Ruby 3.0.
 
 ## Stdlib Worksheet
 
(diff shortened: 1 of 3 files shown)


docs/CONTRIBUTING.md
```diff
@@ -28,8 +28,8 @@
 
 ## The Target Version
 
-* The standard library signatures targets Ruby 2.7 for now.
-* The library code targets Ruby 2.6, 2.7, and 3.0.
+* The standard library signatures targets Ruby 3.0 for now.
+* The library code targets Ruby 3.0.
```
soutaro, line 32: The signature target is Ruby 3.0 now. But the library code targets are 2.7 and 3.0.
I'm sorry for the confusion. 🙇‍♂️

soutaro (changes requested):

docs/CONTRIBUTING.md
kachick: Thank you for letting me know! 😄 
So reverted some changes in https://github.com/ruby/rbs/pull/633/commits/bdc86b12ce04b5f29c80376c7b6caec3bf08a029 🚮 

By the way, should we update gemspec too? https://github.com/ruby/rbs/blob/3bf688260bb4562335fdd8af0cd33e230f09db76/rbs.gemspec#L37

soutaro: Leave it as is. (I'm not sure if rejecting all of the `2.6` users makes sense...)

soutaro (approved):
💎

docs/CONTRIBUTING.md
kachick: Sure!

kachick:
Thank you!

Issue #641, opened 2021-03-14
`RBS_TEST_TARGET` looks not working


kachick:
On current [HEAD](https://github.com/ruby/rbs/tree/b37d41d6dc8fe7aff040eb211c9b2eb38f7eaff5),  https://github.com/ruby/rbs/blob/b37d41d6dc8fe7aff040eb211c9b2eb38f7eaff5/docs/sigs.md#testing-signatures describes as below

```
When you finish writing signature, you may want to test the signature. rbs provides a feature to test your signature.
$ RBS_TEST_TARGET='Foo::*' bundle exec ruby -r rbs/test/setup test/foo_test.rb
```

Then I try following command to run specific tests. Using ruby is `ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin20]`.

```
$ RBS_TEST_TARGET='Array' bundle exec ruby -r rbs/test/setup test/stdlib/Array_test.rb
```

It outputs

```
Loaded suite test/stdlib/Array_test
Started
...................................................................................................
Finished in 0.206635 seconds.
-------------------------------------------------------------------------------------------------------------------------------
99 tests, 592 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------------------------------------------------------
479.11 tests/s, 2864.96 assertions/s
D, [2021-03-14T19:31:46.740999 #79429] DEBUG -- : No type checker was installed!
```

But I have confused, following command reports same test reports.

```
RBS_TEST_TARGET='THIS_CLASS_DOES_NOT_EXIST' bundle exec ruby -r rbs/test/setup test/stdlib/Array_test.rb
```

It outputs

```
Loaded suite test/stdlib/Array_test
Started
...................................................................................................
Finished in 0.217572 seconds.
-------------------------------------------------------------------------------------------------------------------------------
99 tests, 592 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
-------------------------------------------------------------------------------------------------------------------------------
455.02 tests/s, 2720.94 assertions/s
D, [2021-03-14T19:34:38.881770 #79492] DEBUG -- : No type checker was installed!
```

Do I misunderstand the document? 🤔

soutaro:
[...]

PR #642, opened 2021-03-14
[Doc] Add how to test specific stdlib with the path

kachick:
I am using this way since faced https://github.com/ruby/rbs/issues/641.
The rake tasks looks defined at https://github.com/ruby/rbs/blob/4a911a9a8e3de24c1f08c697d7f6f54345c23ea7/Rakefile#L59-L64

It works enough and does not show warnings. So useful to me.
How about to add it to this document? Or it should be an unofficial feature from some reasons?


soutaro (approved):
👍

kachick:
Thank you!

PR #646, opened 2021-03-15
Fix Range#end signature

kachick:


soutaro:
This is intentional. `begin` and `end` may return `nil` for begin-less/end-less ranges. My intuition is it's better to make type parameters optional types for that case.

kachick:
>it's better to make type parameters optional types for that case.

Thanks for your consideration! I will track it 😄

PR #648, opened 2021-03-15
Fix Array#concat signature on subclasses

kachick:
```console
$ ruby -v -e 'p Class.new(Array).new.concat.class'
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin20]
#<Class:0x00007ff93307abf0>
```

```console
$ ruby -v -e 'p Class.new(Array).new.concat.class'
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin20]
#<Class:0x00007faa0010d2e8>
```

### DIFF
--- a/test/stdlib/Array_test.rb
+++ b/test/stdlib/Array_test.rb
@@ -217,6 +217,8 @@ class ArrayInstanceTest < Test::Unit::TestCase
   def test_concat
     assert_send_type "(Array[Integer], Array[Integer]) -> Array[Integer]",
                      [1,2,3], :concat, [4,5,6], [7,8,9]
+    assert_send_type "(Array[Integer], Array[Integer]) -> self",
+                     Class.new(Array).new, :concat, [4,5,6], [7,8,9]
   end
 
   def test_count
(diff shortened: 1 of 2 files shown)


test/stdlib/Array_test.rb
```diff
@@ -217,6 +217,8 @@ def test_compact
   def test_concat
     assert_send_type "(Array[Integer], Array[Integer]) -> Array[Integer]",
                      [1,2,3], :concat, [4,5,6], [7,8,9]
+    assert_send_type "(Array[Integer], Array[Integer]) -> self",
+                     Class.new(Array).new, :concat, [4,5,6], [7,8,9]
```
kachick, line 221: I don’t know why even in without changing signature, this added test passed.  But adding it makes better, I think.
Please tell me a way if `rbs` have something to ensure it.  🙏

soutaro: Good point. The test passes because `Class.new(Array).new.is_a?(Array)` holds. The runtime type checker can be improved...

soutaro (approved):
👍

test/stdlib/Array_test.rb
kachick: 😮 Thank you for letting me know the why!

kachick:
Thank you!

PR #649, opened 2021-03-15
Fix Proc#curry signature

kachick:


core/proc.rbs
```diff
@@ -262,7 +262,7 @@ class Proc < Object
 
   def []: (*untyped arg0) -> untyped
 
-  def curry: (?Integer arity) -> Proc
+  def curry: (?_ToInt arity) -> Proc
```
kachick, line 265:
```suggestion
  def curry: (?int arity) -> Proc
```

`int` looks work too, both is same? Then which one should be preferred? 🤔

soutaro: It's equivalent because `_ToInt` is `int`.

kachick: Sure!

kachick:
Thank you!

PR #658, opened 2021-03-23
Update documentation for overloading methods

Adrian-Hirt:
From https://github.com/ruby/gem_rbs_collection/issues/16:

I did not find the `...` triple dot syntax for overloading methods, e.g. when defining a method multiple times.

I added the `DuplicatedMethodDefinitionError` to the docs/sigs.md, such that users that encounter that error have some pointers on what to do.

I also added the `...` syntax to the `_method-types_` in docs/syntax.md, but I'm not sure if that is 100% the correct place.
In some places, the triple-dot (ellipsis) is used as a placeholder for "other may follow", which can be a bit confusing, and as such, I replaced it by `etc.`. 
However, at two places, I was not entirely sure if the `...` is a mere placeholder, or functions as the overloading operator:

1) In the *Members* section, for the `_method-name_`:

```ruby
_method-name_ ::= ...
                | /`[^`]+`/
```

2) In the *Declarations* section, for the `_global-name_`:

```ruby
_global-name_ ::= /$[a-zA-Z]\w+/ | ...
```

I'd appreciate it if someone could give some feedback, what function the `...` has at that place, I can then gladly add another commit to make it more clear.

Regards, and thank you all for your work

Adrian

### DIFF
--- a/docs/sigs.md
+++ b/docs/sigs.md
@@ -70,6 +70,23 @@ ERROR -- : [Kaigi::Conference#speakers] UnexpectedBlockError: unexpected block i
 The error means there is a type error on overloaded methods.
 The `rbs` test framework tries to the best error message for overloaded methods too, but it reports the `UnresolvedOverloadingError` when it fails.
 
+### DuplicatedMethodDefinitionError
+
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks!

PR #659, opened 2021-03-24
Correct IO#{close_on_exec=, autoclose=, sync=} signatures

kachick:


test/stdlib/IO_test.rb
```diff
@@ -180,4 +184,34 @@ def test_write
       end
     end
   end
+
+  def test_close_on_exec
+    IO.open(IO.sysopen(__FILE__)) do |io|
+      assert_send_type '() -> bool',
+                       io, :close_on_exec?
+      assert_send_type '(::Integer) -> untyped',
```
kachick, line 192:
```suggestion
      assert_send_type '(::Integer) -> ::Integer',
```

When I write above, it fails.
I have issued it in https://bugs.ruby-lang.org/issues/17745 and https://github.com/ruby/ruby/pull/4321.

soutaro (approved):
Thanks!

kachick:
Thank you!

PR #664, opened 2021-04-12
Update type checking

soutaro:


lib/rbs/ast/annotation.rb
```diff
@@ -19,8 +19,8 @@ def hash
         self.class.hash ^ string.hash
       end
 
-      def to_json(*a)
-        { string: string, location: location }.to_json(*a)
+      def to_json(state = _ = nil)
```
akimd, line 22: Sorry, I don't know this idiom.  What does this `_` brings?  TIA!

PR #672, opened 2021-04-26
Expand how to set `pathname`

linhpham:
It's not obvious what pathname should be set to include 
some classes in stdlib.


soutaro:
@linhpham Okay. Could you mention to `set` library too?

linhpham:
Done. Please review again @soutaro . Thanks.

soutaro (approved):
👏

PR #674, opened 2021-05-03
Suppress unused variable warning

pocke:
parser.y displays the following warning:

```bash
$ exe/rbs parse sig/ancestor_builder.rbs
parser.y:611: warning: assigned but unused variable - children
```

This patch suppresses the warning.


soutaro (approved):
🔪

PR #676, opened 2021-05-12
Using rb instead of rbs in README

guillaumebriday:
## Before
Screen Shot 2021-05-12 at 09 57 08

## After
Screen Shot 2021-05-12 at 09 57 21

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ It also allows declaring constants and global variables.
 The following is a small example of RBS for a chat app.
 
 <!-- run-start:a.rbs:bundle exec rbs -I a.rbs validate -->
-```rbs
+```rb
 module ChatApp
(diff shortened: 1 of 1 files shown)


soutaro:
@guillaumebriday Thank you for opening this PR. I don't think changing it to `rb` is a good idea, because RBS is another language with different syntax from Ruby. https://github.com/ruby/rbs/issues/682 seems the correct way for this.

guillaumebriday:
It's seem right!

But that could be a temporary solution and we will change back to `.rbs` when it will be ready on GH?

What do you think?

soutaro:
Hmm..., Okay. Let me merge this!

(It is only one place in the repo, and it is much fewer than I thought... 😄)

Issue #657, opened 2021-03-23
RBS allows including/prepending/extending with a class constant instead of module constant


hjwylde:
It seems that RBS allows including/prepending/extending a class constant. This is a surprise to me as normally it's only possible to include modules into classes/modules in ruby, and I would have expected the same in RBS.

```
class Foo
  def foo: () -> untyped
end

class Qux
  include Foo
end
```

`rbs methods Qux | grep foo` -> `foo (public)`

I would expect classes/modules to not be allowed to include/prepend/extend other classes, and for a sensible error message to be shown. At the moment `steep check` exits with an exception, but it is not a helpful error message.

PR #681, opened 2021-05-28
Do not allow including/prepending/extending a class constant

Yuuki77:
Fix https://github.com/ruby/rbs/issues/657
now it should throw an error 
```ruby
class Foo
  def foo: () -> untyped
end

class Qux
  include Foo
end
```

### DIFF
--- a/lib/rbs/definition_builder/ancestor_builder.rb
+++ b/lib/rbs/definition_builder/ancestor_builder.rb
@@ -321,7 +321,7 @@ module RBS
 
             case
             when member.name.class? && included_modules
-              InvalidMixinClassError.check!(mixin_name: "include", env: env, member: member)
+              MixinClassError.check!(env: env, member: member)
               NoMixinFoundError.check!(member.name, env: env, member: member)
               included_modules << ancestor
             when member.name.interface? && included_interfaces
@@ -331,7 +331,7 @@ module RBS
 
           when AST::Members::Prepend
             if prepended_modules
-              InvalidMixinClassError.check!(mixin_name: "perpend", env: env, member: member)
+              MixinClassError.check!(env: env, member: member)
               NoMixinFoundError.check!(member.name, env: env, member: member)
 
               module_name = member.name
@@ -347,7 +347,7 @@ module RBS
 
             case
             when member.name.class? && extended_modules
-              InvalidMixinClassError.check!(mixin_name: "extend", env: env, member: member)
+              MixinClassError.check!(env: env, member: member)
               NoMixinFoundError.check!(member.name, env: env, member: member)
               extended_modules << ancestor
             when member.name.interface? && extended_interfaces
--- a/lib/rbs/definition_builder/ancestor_builder.rb
+++ b/lib/rbs/definition_builder/ancestor_builder.rb
@@ -237,6 +237,7 @@ module RBS
         end
 
         mixin_ancestors(entry,
+                        type_name,
                         included_modules: ancestors.included_modules,
                         included_interfaces: ancestors.included_interfaces,
                         prepended_modules: ancestors.prepended_modules,
@@ -284,6 +285,7 @@ module RBS
         end
 
         mixin_ancestors(entry,
+                        type_name,
                         included_modules: nil,
                         included_interfaces: nil,
                         prepended_modules: nil,
@@ -301,6 +303,7 @@ module RBS
 
             OneAncestors.interface(type_name: type_name, params: params).tap do |ancestors|
               mixin_ancestors0(entry.decl,
+                               type_name,
                                align_params: nil,
                                included_modules: nil,
                                included_interfaces: ancestors.included_interfaces,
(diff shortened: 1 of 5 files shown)


lib/rbs/definition_builder/ancestor_builder.rb
```diff
@@ -321,6 +321,7 @@ def mixin_ancestors0(decl, align_params:, included_modules:, included_interfaces
 
             case
             when member.name.class? && included_modules
+              InvalidMixinClassError.check!(mixin_name: "include", env: env, member: member)
```
soutaro, line 324: The `mixin_name` argument seems redundant.
It would be better to detect the _type_ of the mixin from `member`.

```
case member
when AST::Members::Prepend
  :prepend
when AST::Members::Include
  :include
when AST::Members::Extend
  :extend
end
```

lib/rbs/errors.rb
```diff
@@ -350,4 +350,29 @@ def location
...
+
+  class InvalidMixinClassError < DefinitionError
+    attr_reader :type_name
+    attr_reader :member
+
+    def initialize(mixin_name:, member:)
+      @type_name = member.name
+      @member = member
+
+      super "#{Location.to_string location}: Not allow to #{mixin_name} #{type_name} class"
```
soutaro, line 362: Could you change the message?

Something like 

```
Cannot #{mixin_operation} a class `#{type_name}`
```

seems better to me.

lib/rbs/errors.rb
```diff
@@ -350,4 +350,29 @@ def location
       original.location
     end
   end
+
+  class InvalidMixinClassError < DefinitionError
```
soutaro, line 354: Could you call it `MixinClassError`?

soutaro: Please add an RBS declaration in `errors.rbs` like:

```
class MixinClassError < DefinitionError
  type member = AST::Members::Include | AST::Members::Prepend | AST::Members::Extend

  attr_reader type_name: TypeName
  attr_reader member: member

  def initialize: (type_name: TypeName, member: member) -> void

  def location: () -> Location?

  def self.check!: (TypeName, env: Environment, member: member) -> void
end
```

lib/rbs/errors.rb
```diff
@@ -350,4 +350,29 @@ def location
...
+
+      super "#{Location.to_string location}: Not allow to #{mixin_name} #{type_name} class"
+    end
+
+    def location
+      member.location
+    end
+
+    def self.check!(mixin_name:, env:, member:)
+      if member.name.class? && env.class_decls.key?(member.name)
```
soutaro, line 370: The `if` condition can be simplified to only `member.name.class?`.

soutaro (changes requested):
@Yuuki77 Good first step! I'd like to ask to fix the following:

* [x] Change error classname
* [x] Change the error message
* [x] Fix some of the implementation
* [x] Add RBS declaration of the new error class

lib/rbs/errors.rb
```diff
@@ -350,4 +350,42 @@ def location
       original.location
     end
   end
+
+  class MixinClassError < DefinitionError
+    attr_reader :type_name
+    attr_reader :member
+
+    def initialize(member:)
+      @type_name = member.name
```
soutaro, line 359: Should add a parameter `typename` which contains the name of the class/module definition.

lib/rbs/errors.rb
```diff
@@ -350,4 +350,42 @@ def location
...
+      @member = member
+
+      super "Cannot #{mixin_name} a class #{member.name} in the definition of #{type_name}"
```
soutaro, line 362:
```suggestion
      super "Cannot #{mixin_name} a class `#{member.name}` in the definition of `#{type_name}`"
```

lib/rbs/errors.rb
```diff
@@ -350,4 +350,42 @@ def location
...
+    end
+
+    def location
+      member.location
+    end
+
+    def self.check!(env:, member:)
```
soutaro, line 369: Add `type_name:` keyword argument.

lib/rbs/errors.rb
```diff
@@ -350,4 +350,42 @@ def location
...
+
+  class MixinClassError < DefinitionError
+    attr_reader :type_name
+    attr_reader :member
+
+    def initialize(member:)
+      @type_name = member.name
+      @member = member
+
+      super "Cannot #{mixin_name} a class `#{member.name}` in the definition of `#{type_name}`"
```
soutaro, line 362:
```suggestion
      super "#{Location.to_string member.location}: Cannot #{mixin_name} a class `#{member.name}` in the definition of `#{type_name}`"
```

lib/rbs/errors.rb
```diff
@@ -350,4 +350,42 @@ def location
...
+    end
+
+    def location
+      member.location
+    end
+
+    def self.check!(env:, member:)
+      if member.name.class?
```
soutaro, line 370: Drop the `member.name.class?` test.

lib/rbs/errors.rb
Yuuki77: now error is like this
```
:in `check!': test.rbs:6:2...6:13: Cannot include a class `::Foo` in the definition of `::Qux` (RBS::MixinClassError)
```
this is the sample file
```ruby
class Foo
  def foo: () -> untyped
end

class Qux
  include Foo
end
```

Yuuki77:
Thank you for the reviews 🙏 
now I fixed what you comment.

soutaro (approved):
👏

PR #683, opened 2021-06-16
Handle Mutex#sleep change planned in Ruby 3.1

jeremyevans:
Mutex#sleep in Ruby 3.1 will return nil if not interrupted. This
change is needed so that ruby/ruby PR 4256 can be merged, in order
to fix Ruby bug 16608.

### DIFF
--- a/test/stdlib/Mutex_m_test.rb
+++ b/test/stdlib/Mutex_m_test.rb
@@ -47,10 +47,9 @@ class Mutex_mInstanceTest < Test::Unit::TestCase
   def test_sleep
     mu = mu()
     mu.lock
-    return_type = "Integer#{'?' if RUBY_VERSION >= '3.1'}"
-    assert_send_type "(Integer) -> #{return_type}",
+    assert_send_type "(Integer) -> Integer?",
                      mu, :sleep, 0
-    assert_send_type "(Float) -> #{return_type}",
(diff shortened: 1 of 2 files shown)


jeremyevans:
CI failure does not look related to the change, as it occurs while installing dependencies, before the tests are run: `E: Unable to locate package libdb-dev`

test/stdlib/Mutex_m_test.rb
```diff
@@ -47,9 +47,10 @@ def test_mu_unlock
   def test_sleep
     mu = mu()
     mu.lock
-    assert_send_type "(Integer) -> Integer",
+    return_type = "Integer#{'?' if RUBY_VERSION >= '3.1'}"
+    assert_send_type "(Integer) -> #{return_type}",
```
soutaro, line 51: I think simply changing to `Integer?` for all versions is good.

soutaro:
Thank you @jeremyevans!
 I think we should also change the definition in `mutex_m.rbs` to return `nil`.

```
  def sleep: (?Numeric timeout) -> Integer?
```

jeremyevans:
@soutaro Thank you for your review. I've pushed commits to address both of those issues.

soutaro (approved):
Thank you! 🎉

PR #684, opened 2021-06-18
Revise Marshal signature

soutaro:


core/marshal.rbs
```diff
@@ -136,8 +136,8 @@ module Marshal
   #     ThreadGroup, Continuation
   # *   objects which define singleton methods
   #
-  def self.dump: (Object arg0, ?IO arg1, ?Integer arg2) -> Object
-               | (Object arg0, ?Integer arg1) -> Object
+  def self.dump: (untyped obj, untyped port, ?Integer limit) -> untyped
```
soutaro, line 139: The type of `port` can be other than `IO`, while it's not very clear which type is good.

PR #689, opened 2021-06-24
s/a RBS/an RBS/

pocke:
I'm not professional of English but I think "an RBS" is easier to pronounce than "a RBS"

### DIFF
--- a/docs/stdlib.md
+++ b/docs/stdlib.md
@@ -57,7 +57,7 @@ But it will help you to have an overview of the signatures you are trying to wri
 
 ### What to do with existing RBS files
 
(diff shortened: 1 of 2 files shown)


docs/stdlib.md
```diff
@@ -57,7 +57,7 @@ But it will help you to have an overview of the signatures you are trying to wri
 
 ### What to do with existing RBS files
 
-Generating prototypes will override everything, so the problem is if there is a RBS files already.
+Generating prototypes will override everything, so the problem is if there is an RBS files already.
```
adam12, line 60:
```suggestion
Generating prototypes will overwrite existing RBS files, which might be undesirable.
```

PR #691, opened 2021-06-26
Do not check parser.rb diff on ruby-master

pocke:
To fix build failure. For example: https://github.com/ruby/rbs/pull/689/checks?check_run_id=2921271817
Racc can have unreleased change on Ruby's master branch, so skip
`confirm_parser` rake task on ruby master.


soutaro (approved):
👍

PR #692, opened 2021-06-26
test racc error

soutaro:


soutaro:
We cannot reinstall racc 1.5.2 from rubygems.org because:

* They have exactly the same version with the HEAD version included in master-nightly.
* It is _default_ gem so that it cannot be uninstalled.

💣

PR #693, opened 2021-06-27
Add optparse signature

pocke:
This pull request adds OptionParser RBSs.

Note that I kept `untyped` for internal APIs and private methods. Typing them will be hard work but only a few benefits.

`optparse` library has files that is needed to require explicitly. For example, `optparse/ac` file.
But this pull request doesn't add types for them.

### DIFF
--- a/sig/polyfill.rbs
+++ b/sig/polyfill.rbs
@@ -11,24 +11,6 @@ module Gem
   end
 end
 
-class OptionParser
-  def initialize: () ?{ (instance) -> void } -> void
-
-  def on: [T] (*String) { (T) -> void } -> void
-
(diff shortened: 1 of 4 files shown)


sig/polyfill.rbs
```diff
@@ -11,24 +11,6 @@ module Gem
   end
 end
 
-class OptionParser
```
pocke, line 14: OptionParser polyfill for internal is not necessary :put_litter_in_its_place: 

The new type definition of `OptionParser#on` introduces a problem. I guess it is a Steep's bug. see https://github.com/soutaro/steep/issues/389

soutaro (approved):
🐙

PR #694, opened 2021-06-28
Add docs to teach RBS method syntax

alts:
I put this doc together to teach my team how to write method signatures. It's assumes the audience knows nothing about RBS, but is familiar with the standard library, and uses that as a starting point for teaching RBS syntax/grammar. When I was learning RBS, I mostly learned through trial-and-error, and I hope that this guide can help people become proficient in signature writing faster.


soutaro (approved):
:clap:

PR #697, opened 2021-06-30
[resolv] Add RBS file for Resolv

HoneyryderChuck:
Added RBS signatures for the `resolv` package, + some tests.


soutaro (changes requested):
Please fix the name resolution errors like `dns_name`, `Sender`, etc... Write the type names in qualified style or declare classes/modules in nested style.

I also found socket classes are undefined. I think writing socket RBS is too much for this, so just adding empty classes in a file like `resolv/0/sockets.rbs` is good.

HoneyryderChuck:
> I also found socket classes are undefined.

@soutaro I noticed it while reviewing the errors, and went ahead with it [here](https://github.com/ruby/rbs/pull/699). Let's  deal with it first, then I'll revisit this.

HoneyryderChuck:
@soutaro it's ready for review.

> Write the type names in qualified style or declare classes/modules in nested style.

I didn't understand this comment. Can you elaborate?

soutaro:
@HoneyryderChuck Could you remove the trailing whit spaces?

HoneyryderChuck:
@soutaro done.

soutaro (approved):
👏

PR #698, opened 2021-07-01
changing IO.select timeout into a numeric

HoneyryderChuck:
`IO.select` actually can receive floats, and I've tried it with rationals as well.


soutaro (approved):
Thanks!

PR #702, opened 2021-07-05
Fix race condition for multi-thread support

### LINKED ISSUE #695: RuntimeError because interface_cache not thread_safe
I'm having a non-deterministic error popping up sometimes in CI, around [this part of the rbs code](https://github.com/ruby/rbs/blob/master/lib/rbs/definition_builder.rb#L786-L793), when running "parallel mode" `minitest`. 

I believe that two threads are contending on that code block, where one sets `cc` to `false` then yielding, whereas another thread enters this part before the yield from the previous thread finishes.

I believe that extending `interface_cache` with `Mutex_m` and wrapping that block with `synchronize{}` could mitigate this issue.

HoneyryderChuck:
Inside DefinitionBuilder, access to caches happens recursively, where
set operations mark the field in the cache with false. This breaks
sometimes when used in a threaded context, such as a minitest parallel
run.

This patch fixes it by making cache accesses mutex-aware, while also
allowing recursive sets on the same cache.

Fixes #695

### DIFF
--- a/lib/rbs/definition_builder.rb
+++ b/lib/rbs/definition_builder.rb
@@ -774,26 +774,10 @@ module RBS
     end
 
     def try_cache(type_name, cache:, key: type_name)
-      # @type var cc: Hash[untyped, Definition | false | nil]
+      # @type var cc: Hash[untyped, Definition | nil]
       cc = _ = cache
-      cached = cc[key]
 
-      case cached
-      when Definition
-        cached
-      when false
-        raise
-      when nil
-        cc[key] = false
-        begin
-          cc[key] = yield
-        rescue => ex
-          cc.delete(key)
-          raise ex
-        end
-      else
-        raise
-      end
+      cc[key] ||= yield
     end
 
     def expand_alias(type_name)


soutaro:
I'm not very open to merge this PR. Leaving possible technical improvements.

lib/rbs/definition_builder.rb
```diff
@@ -774,25 +780,13 @@ def merge_method(type_name, methods, name, method, sub, implemented_in: :keep, k
...
-      when nil
-        cc[key] = false
-        begin
-          cc[key] = yield
-        rescue => ex
-          cc.delete(key)
-          raise ex
-        end
+      if cc.locked?
+        cc[key] ||= yield
```
soutaro, line 787: This is a good improvement. 👏  The recursion check is moved away to `AncestorBuilder`, so we don't need assigning `false`.

lib/rbs/definition_builder.rb
```diff
@@ -774,25 +780,13 @@ def merge_method(type_name, methods, name, method, sub, implemented_in: :keep, k
...
-        raise
-      when nil
-        cc[key] = false
-        begin
-          cc[key] = yield
-        rescue => ex
-          cc.delete(key)
-          raise ex
-        end
+      if cc.locked?
```
soutaro, line 786: Using a [Monitor](https://docs.ruby-lang.org/en/3.0.0/Monitor.html) would make the code simple.

```
@monitor.synchronize do 
  cc[key] ||= yield
end
```

HoneyryderChuck:
@soutaro removed the lock strategies. I found out during the testing that the issue is not thread-safety, but rather the order in which the modules are saved and cached, i.e. a class signature might be building, where used modules might make a reference to it, before it ever gets cached. It's probably a tad inneficient, although I don't think it's an issue, assuming the result is idempotent.

soutaro:
> the issue is not thread-safety

Oh no. It sounds reasonable.

> assuming the result is idempotent

Correct.

Dose this patch solve your problem?

HoneyryderChuck:
It does. 👍

soutaro (approved):
🧵

soutaro:
@HoneyryderChuck Merged. Thank you for finding and fixing the issue! 🎉

PR #705, opened 2021-07-05
fixed signature of caller

HoneyryderChuck:
`Kernel.caller` signatures were overriding the case of no arguments, leading to the steep errors when performing `ex.set_backtrace(caller)`

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -13,8 +13,8 @@
 module Kernel : BasicObject
   private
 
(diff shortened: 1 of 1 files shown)


pocke:
How about removing `?` of `?::Range[Integer] start_or_range)`? `Kernel.caller` actually allows calling with a Range but your change disallows it.

HoneyryderChuck:
@pocke well spotted! fixed.

pocke (approved):
LGTM :+1: Thanks!

PR #706, opened 2021-07-05
Fix method_missing signature with non kwargs

HoneyryderChuck:
`method_missing` is also allowed to receive kwargs.


core/basic_object.rbs
```diff
@@ -216,6 +216,7 @@ class BasicObject
   #     r.mm      #=> 2000
   #
   def method_missing: (Symbol, *untyped) -> untyped
+                    | (Symbol, *untyped, **untyped) -> untyped
```
soutaro, line 219: Good point! Could you merge the two overloads?

soutaro (approved):
Good point!

PR #707, opened 2021-07-09
Avoid redundant overloading of Array#shift

mame:
It returns an array of elements only when the length argument is
specified.


pocke (approved):
:clap:

soutaro (approved):
👏

PR #711, opened 2021-07-11
Enumerable#flat_map

soutaro:
Drops object version from `Enumerable#flat_map`, because it usually expects the block value is an `Array`

```ruby
[1,2,3].flat_map {|x| [x] }    # OK because the block returns an Array.
[1,2,3].flat_map {|x| x+1 }    # Type error with this type signature update, because the non-array block type is dropped.
```

### DIFF
--- a/core/enumerable.rbs
+++ b/core/enumerable.rbs
@@ -48,7 +48,7 @@ module Enumerable[unchecked out Elem]: _Each[Elem]
   def collect: [U] () { (Elem arg0) -> U } -> ::Array[U]
              | () -> ::Enumerator[Elem, ::Array[untyped]]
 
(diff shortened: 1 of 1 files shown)


soutaro:
The implementation of `#flat_map` implies exactly current method type. 🤔

hjwylde:
What's the reason for making this change? Using a non-array as the return value in the block is valid.

```
irb(main):001:0> [1, 2, 3].flat_map { _1 + 1 }
=> [2, 3, 4]
```

soutaro:
@hjwylde Yes. That's valid and the current type is correct. I have to fix Steep to type check the code, instead of changing the type of the method.

Issue #687, opened 2021-06-22
RBS allows recursive type aliases


stoyanovskiy239:
#### Steps to reproduce
In the terminal:
```
echo "type x = x" >> test.rbs
rbs -I . validate
```

#### Expected
An exception of sorts, with some reasonable error message.

#### Actual
```
Validating alias: `::x`...
```
is present in the output, `validate` execution seems to finish successfully.

#### Environment
```
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin19]
rbs 1.2.1
macOS Catalina 10.15.7
```

soutaro:
Hi @stoyanovskiy239,

Thank you for reporting this issue. Yes. This is clearly a bug. We will implement validation for recursive definitions...

PR #719, opened 2021-07-17
Add Recursiive type alias defnition validation

raosush:
# Description
- Implemented validation for recursive type alias
- Added a `RecursiveTypeAliasError` class
- Added signatures for `RecursiveTypeAliasError`
- Added a test to validate the funcitoning of `RecursiveTypeAliasError`

# Example
image

Fixes: #687

Signed-off-by: Sushanth Sathesh Rao <rao.sush004@gmail.com>

### DIFF
--- a/lib/rbs/errors.rb
+++ b/lib/rbs/errors.rb
@@ -386,4 +386,20 @@ module RBS
       end
     end
   end
+
+  class RecursiveTypeAliasError < LoadingError
+    attr_reader :alias_names
+    attr_reader :location
+
+    def initialize(alias_names:, location:)
+      @alias_names = alias_names
+      @location = location
+
+      super "#{Location.to_string location}: Recursive type alias definition found for: #{name}"
+    end
+
+    def name
+      @alias_names.map(&:name).join(', ')
+    end
+  end
 end
--- a/lib/rbs/type_alias_dependency.rb
+++ b/lib/rbs/type_alias_dependency.rb
@@ -1,0 +1,88 @@
+module RBS
+  class TypeAliasDependency
+    attr_reader :env
+
+    # Direct dependencies corresponds to a directed graph
+    # with vertices as types and directions based on assignment of types
+    attr_reader :direct_dependencies
+    # A hash which stores the transitive closure
+    # of the directed graph
+    attr_reader :dependencies
+
+    def initialize(env:)
+      @env = env
+    end
+
+    # Check if an alias type definition is circular & prohibited
+    def circular_definition?(alias_name)
+      # Construct transitive closure, if not constructed already
+      transitive_closure() unless @dependencies
+
+      # Check for recursive type alias
+      @dependencies[alias_name][alias_name]
+    end
+
+    def build_dependencies
+      return if @direct_dependencies
+
+      # Initialize hash(a directed graph)
+      @direct_dependencies = {}
+      # Initialize dependencies as an empty hash
+      @dependencies = {}
+      # Iterate over alias declarations inserted into environment
+      env.alias_decls.each do |name, entry|
+        # Construct a directed graph by recursively extracting type aliases
+        @direct_dependencies[name] = direct_dependency(entry.decl.type)
+        # Initialize dependencies with an empty hash
+        @dependencies[name] = {}
+      end
+    end
+
+    def transitive_closure
+      # Construct a graph of direct dependencies
+      build_dependencies()
+      # Construct transitive closure by using DFS(recursive technique)
+      @direct_dependencies.each_key do |name|
+        dependency(name, name)
+      end
+    end
+
+    private
+
+    # Constructs directed graph recursively
+    def direct_dependency(type, result = Set[])
+      case type
+      when RBS::Types::Union, RBS::Types::Intersection, RBS::Types::Optional
+        # Iterate over nested types & extract type aliases recursively
+        type.each_type do |nested_type|
+          direct_dependency(nested_type, result)
+        end
+      when RBS::Types::Alias
+        # Append type name if the type is an alias
+        result << type.name
+      end
+
+      result
+    end
+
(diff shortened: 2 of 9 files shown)


raosush:
# Suggestion
* The current code raises `LoadingError` if & only if the assigned type is an alias & is the same as the one it is being assigned to
* The implementation could be changed to raise an exception whenever the type being assigned to an alias is again an alias, because I do not seem to find any use case of aliasing an aliased type

raosush:
- [ ] Implement recursive type alias validation for `RBS::Types::Union` and `RBS::Types::Intersection`

soutaro:
Thank you @raosush, but I don't think this is the best way to implement the validation. We can discuss this later.

test/validator_test.rb
```diff
@@ -59,4 +59,46 @@ class Foo
       end
     end
   end
+
+  def test_validate_recursive_type_alias
```
soutaro, line 63: Please include assertions to let us know which `type` declaration is prohibited/allowed.

lib/rbs/definition_builder.rb
```diff
@@ -786,6 +786,59 @@ def expand_alias(type_name)
...
+        if cache[name].nil?
+          # nop
+        else
+          cache[name].each(&block)
+        end
+      end
+      # Build a hash table of type names & strongly connected aliases index
+      scc_hash = {}
+      # Apply TSort technique & fetch strongly connected components
+      strongly_connected_components = TSort.strongly_connected_components(each_node, each_child)
```
soutaro, line 810: I don't think building strongly connected components is necessary.
We can build the transitive closure of dependencies from type name. And test if the left hand side is included in the closure. I believe the definition with transitive closure is more straightforward.

1. Define _dependent type names_ for given type `T` (`DD(T)`) same as `extract_type_aliases`.
2. Calculate transitive closure of _dependent type names_ (`D(T)`) of one type.
3. Test if the type alias definition `type t = S` includes a circular reference by `D(S) ∋ t`.

On the other hand, I'm not very sure if which one is faster.

soutaro:
We can define a class to calculate dependencies between types like `TypeAliasDependency` that:

```rbs
class TypeAliasDependency
  attr_reader env: Environment

  attr_reader direct_dependencies: Hash[Types::t, Set[TypeName]]    # Cache calculated dependencies for performance
  attr_reader dependencies: Hash[Types::t, Set[TypeName]]
  
  def initialize: (env: Environment) -> void

  # Returns true if type alias declaration of `alias_name` is circular and prohibited.
  #
  def circular_definition?: (TypeName alias_name) -> bool

  private

  def direct_dependency: (Types::t) -> Set[TypeName]       # Utility methods to calculate dependencies
  def dependency: (Types::t) -> Set[TypeName]
end
```

This is an example to illustrate my idea, and you would need to revise the detail of the API.

test/validator_test.rb
soutaro: All of the examples are prohibited ones. 👍

lib/rbs/errors.rb
```diff
@@ -386,4 +386,59 @@ def mixin_name
...
+      @location = location
+
+      super "#{Location.to_string location}: Recursive type alias definition found for: #{name}"
+    end
+
+    def name
+      @defs.map(&:name).join(', ')
+    end
+
+    def self.check!(entry:, scc:, scc_hash:)
```
soutaro, line 405: Looks better to implement the test in `Validator` class for this case.

sig/errors.rbs
```diff
@@ -196,4 +196,19 @@ module RBS
 
     def mixin_name: () -> String
   end
+
+  class RecursiveTypeAliasError < LoadingError
+    attr_reader defs: Array[TypeName]
```
soutaro, line 201: Rename to `alias_names`.

soutaro (changes requested):
@raosush Posted comments!

lib/rbs/type_alias_dependency.rb
```diff
@@ -0,0 +1,105 @@
...
+        end
+      end
+
+      scc_hash
+    end
+
+    private
+
+    # Constructs directed graph recursively
+    def direct_dependency(type, result = Set[])
```
soutaro, line 73: I'm not sure if this definition of `#direct_dependency` allows type definitions like:

```rbs
type a = Array[a]
type b = [b, b]
type c = { foo: c }
type d = ^(d) -> d
```

`#direct_dependency` will test if the type is a union, optional, or intersection and then returns the `#direct_dependency`s of the direct child types.

lib/rbs/type_alias_dependency.rb
```diff
@@ -74,6 +76,8 @@ def direct_dependency(type, result = Set[])
       # Append type name if the type is an alias
       result << type.name if type.is_a?(RBS::Types::Alias)
 
+      return result if SAFELIST.include?(type.class)
```
soutaro, line 79: Please write a case analysis using `case-when` here, instead of having `SAFELIST`.

lib/rbs/type_alias_dependency.rb
```diff
@@ -0,0 +1,109 @@
...
+      # Initialize hash(a directed graph)
+      @direct_dependencies = {}
+      # Initialize dependencies as an empty hash
+      @dependencies = {}
+      # Iterate over alias declarations inserted into environment
+      env.alias_decls.each do |name, entry|
+        # Construct a directed graph by recursively extracting type aliases
+        @direct_dependencies[name] = direct_dependency(entry.decl.type)
+        # Initialize dependencies with an empty hash
+        @dependencies[name] = {}
```
soutaro, line 39:
```suggestion
        @dependencies[name] = Set[]
```

soutaro (approved):
Thanks! 👍

PR #723, opened 2021-07-22
Fix parser to accept alias names starting with `_`

### LINKED ISSUE #722: Regression in 1.3.1 with parameters that start with `_` no longer valid
After updating from RBS 1.2 to 1.3.1, parameters that start with an underscore are no longer valid. Example from [ruby/gem_rbs_collection](https://github.com/ruby/gem_rbs_collection):

https://github.com/ruby/gem_rbs_collection/blob/main/gems/actionpack/6.0/actionpack-generated.rbs#L5750-L5753
```rb
    # Aliasing these off because AD::Http::Cache::Response defines them.
    alias _cache_control cache_control

    alias _cache_control= cache_control=
```

yields this error when running `steep_check`:

```sh
#<RBS::Parser::SyntaxError: parse error on value: #<RBS::Parser::LocatedValue:0x00007f8bd14dd5f0 @location=#<RBS::Location:5240 @buffer=/Users/john/Documents/GitHub/repo/gem_rbs/gems/actionpack/6.0/actionpack-generated.rbs, @pos=210321...210335, source='_cache_control', start_line=5751, start_column=10>, @value="_cache_control"> (tUNDERSCOREIDENT)>
```

It looks like https://github.com/ruby/rbs/pull/721 was an attempt to address this bug but didn't totally fix it.

Also filed https://github.com/ruby/gem_rbs_collection/issues/44 in case it's an issue with the generated RBS files in the repo.

pocke:
Fix #722

I've confirmed other `tINTERFACEIDENT` usage, for `var_name_opt`, `simple_name`, and`interface_name` are valid.
`var_name_opt` has been fixed by https://github.com/ruby/rbs/pull/721.
`simple_name` is used by `qualified_name`, and it doesn't need to accept underscored name. For example, `Foo::_bar` is invalid always.
`inteface_name` is just `tINTERFACEIDENT`, so it doesn't need to accept `tUNDERSCOREIDENT`.

Now `exe/rbs parse ../gem_rbs_collection/gems/**/*.rbs` prints no error :tada:


test/rbs/signature_parsing_test.rb
```diff
@@ -1761,11 +1761,27 @@ def hello: (String _name) -> void
...
     end
   end
+
+  def test_underscore_qualified_name
+    assert_raises RBS::Parser::SyntaxError do
+      Parser.parse_signature(<<-RBS)
+type x = Foo::_bar
+      RBS
+    end
+  end
```
pocke, line 1786: This PR doesn't affect qualified_name, but I added the test for qualifed_name to clear that it doesn't accept tUNDERSCOREIDENT.

soutaro (approved):
Thanks! 🙇‍♂️

PR #728, opened 2021-08-03
Correct various signatures

hjwylde:
This just corrects a number of incorrect signatures that I found. It also updates a few to have types that I feel are more useful for users and type checkers (e.g., `Hash#replace` and the `TrueClass`/`FalseClass` updates).

I've tried to add tests to the existing test classes where possible, however I'm not sure I understand the difference in test cases when `assert_send_type` is used vs. regular method calls (such as in `Float_test.rb`), so I just used the style of the given file.

Let me know if this looks okay, or if I missed something and need to make some changes to it.

### DIFF
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -628,8 +628,8 @@ class Array[unchecked out Elem] < Object
   #     a #=>  ["", "b", "c!", "d!"]
   #
   # collect! is monomorphic because of RBS limitation.
-  def collect!: [U] () { (Elem item) -> U } -> ::Array[U]
-              | () -> ::Enumerator[Elem, ::Array[untyped]]
+  def collect!: () { (Elem item) -> Elem } -> self
+              | () -> ::Enumerator[Elem, self]
 
   # When invoked with a block, yields all combinations of length `n` of elements
   # from the array and then returns the array itself.
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -577,7 +577,7 @@ class Array[unchecked out Elem] < Object
   # actually picked up at each iteration.
   #
   def bsearch: () -> ::Enumerator[Elem, Elem?]
-             | () { (Elem) -> bool } -> Elem?
+             | () { (Elem) -> boolish } -> Elem?
              | () { (Elem) -> ::Integer } -> Elem?
 
   # By using binary search, finds an index of a value from this array which meets
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -576,7 +576,8 @@ class Array[unchecked out Elem] < Object
   # true/false, or always return a number.  It is undefined which value is
   # actually picked up at each iteration.
   #
-  def bsearch: () { (Elem) -> (true | false) } -> Elem?
+  def bsearch: () -> ::Enumerator[Elem, Elem?]
+             | () { (Elem) -> bool } -> Elem?
              | () { (Elem) -> ::Integer } -> Elem?
 
   # By using binary search, finds an index of a value from this array which meets
@@ -627,8 +628,8 @@ class Array[unchecked out Elem] < Object
   #     a #=>  ["", "b", "c!", "d!"]
   #
   # collect! is monomorphic because of RBS limitation.
-  def collect!: () { (Elem item) -> Elem } -> self
-              | () -> ::Enumerator[Elem, self]
+  def collect!: [U] () { (Elem item) -> U } -> ::Array[U]
+              | () -> ::Enumerator[Elem, ::Array[untyped]]
 
   # When invoked with a block, yields all combinations of length `n` of elements
   # from the array and then returns the array itself.
@@ -855,7 +856,7 @@ class Array[unchecked out Elem] < Object
   #     0 -- 1 -- 2 --
   #
   def each_index: () { (::Integer index) -> void } -> self
-                | () -> ::Enumerator[Elem, self]
+                | () -> ::Enumerator[::Integer, self]
 
   # Returns `true` if `self` contains no elements.
   #
@@ -1585,7 +1586,7 @@ class Array[unchecked out Elem] < Object
   #     a.sample(4, random: Random.new(1))  #=> [6, 10, 9, 2]
   #
   def sample: (?random: _Rand rng) -> Elem?
-            | (?int n, ?random: _Rand rng) -> ::Array[Elem]
+            | (int n, ?random: _Rand rng) -> ::Array[Elem]
 
   # Returns a new array containing all elements of `ary` for which the given
   # `block` returns a true value.
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -577,7 +577,7 @@ class Array[unchecked out Elem] < Object
   # actually picked up at each iteration.
   #
   def bsearch: () -> ::Enumerator[Elem, Elem?]
-             | () { (Elem) -> boolish } -> Elem?
+             | () { (Elem) -> (true | false) } -> Elem?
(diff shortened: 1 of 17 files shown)


core/hash.rbs
```diff
@@ -776,7 +776,7 @@ class Hash[unchecked out K, unchecked out V] < Object
   #     h = { "a" => 100, "b" => 200 }
   #     h.replace({ "c" => 300, "d" => 400 })   #=> {"c"=>300, "d"=>400}
   #
-  def replace: (Hash[K, V]) -> self
+  def replace: [A, B] (Hash[A, B]) -> Hash[A, B]
```
soutaro, line 779:
```suggestion
  def replace: [A, B] (Hash[A, B]) -> Hash[A | K, B | V]
```

core/hash.rbs
```diff
@@ -935,8 +935,8 @@ class Hash[unchecked out K, unchecked out V] < Object
   #
   # If no block is given, an enumerator is returned instead.
   #
-  def transform_values!: () -> Enumerator[K, Hash[K, untyped]]
-                       | () { (V) -> V } -> Hash[K, V]
+  def transform_values!: () -> Enumerator[V, Hash[K, untyped]]
+                       | [A] () { (V) -> A } -> Hash[K, A]
```
soutaro, line 939: Please keep the monomorphic version.

core/object.rbs
```diff
@@ -205,8 +205,8 @@ class Object < BasicObject
   #     enum.first(4) # => [1, 1, 1, 2]
   #     enum.size # => 42
   #
-  def enum_for: (Symbol method, *untyped args) ?{ (*untyped args) -> Integer } -> Enumerator[untyped, untyped]
-              | (*untyped args) ?{ (*untyped args) -> Integer } -> Enumerator[untyped, untyped]
+  def enum_for: (Symbol method, *untyped args) ?{ (*untyped args) -> Integer } -> Enumerator[untyped, self]
```
soutaro, line 208: I don't think we can assume `Enumerator[untyped, self]` here, because the `enum_for` can be used for any methods.

core/range.rbs
```diff
@@ -101,7 +101,8 @@ class Range[out Elem] < Object
   # ```
   def begin: () -> Elem      # Begin-less ranges have type of Range[Integer?]
 
-  def bsearch: [U] () { (Elem) -> boolish } -> U?
+  def bsearch: () -> ::Enumerator[Elem, Elem?]
+             | () { (Elem) -> boolish } -> Elem?
```
soutaro, line 105: Maybe the same type with `Array#bsearch` ?

soutaro (changes requested):
@hjwylde Thank you for opening this PR! I post some comments to improve.

core/array.rbs
```diff
@@ -627,8 +628,8 @@ class Array[unchecked out Elem] < Object
   #     a #=>  ["", "b", "c!", "d!"]
   #
   # collect! is monomorphic because of RBS limitation.
-  def collect!: () { (Elem item) -> Elem } -> self
-              | () -> ::Enumerator[Elem, self]
+  def collect!: [U] () { (Elem item) -> U } -> ::Array[U]
```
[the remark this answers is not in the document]
hjwylde: Hm, in such cases, is it really better to just return `self`?

I understand that it makes the signature describe the method behaviour better, however I think using generics here will help the signature to better describe the type behaviour. To me this is more what type signatures are for, as we can refer to the documentation or implementation if we are interested in learning the side-effects of the method.

core/hash.rbs
hjwylde: Could you elaborate on the reason for having `| K`/`| V` please?

AFAIK `replace` clears the hash first, so there shouldn't be any `K`s or `V`s left. I could see a reason for having `| K`/`| V` for `{a: 0, b: 1}.replace({})`, however here I think it's better to just use `clear` which will retain the receiver's type information.

core/hash.rbs
hjwylde: Thanks. I'll just wait until the discussion on `collect!` is resolved as it's related.

core/object.rbs
hjwylde: Good point, I'll revert this.

hjwylde: I was just thinking on this, I realised the main reason I added this was because `to_enum` is aliased here, so the use-case where it's useful is for when someone calls `some_obj.to_enum` with no arguments - this defaults to `each` and returning `self`.

Would the following signature be okay by you? Here `self` is the return type when no method symbol is provided, and in the other situation we use `untyped`.

```
  def enum_for: (*untyped args) ?{ (*untyped args) -> Integer } -> Enumerator[untyped, self]
              | (Symbol method, *untyped args) ?{ (*untyped args) -> Integer } -> Enumerator[untyped, untyped]
```

soutaro: I agree that most (all?) `#each` implementation returns `self` and assuming `self` is reasonable.

Minor fix: if we omit the `method` argument, we cannot pass any other argument?

```
def enum_for: () ?{ () -> Integer } -> Enumerator[untyped, self]
            | ~~~
```

core/hash.rbs
soutaro: My bad. I was confusing with `#merge`.
Will continue at `#collect!` for in-place updates.

core/array.rbs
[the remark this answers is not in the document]
soutaro: The following is the case to detect a type error in my mind.

```rb
a = [1,2,3]
a.map! {|x| x.to_s }
a[0] + 1
```

On the other hand, the following runs without any error == Steep detects a false error.

```rb
a = [1,2,3]
b = a.map! {|x| x.to_s }
b[0] + "string"
```

I'm not sure if we can find a solution to make the both work, and we may have to choose one. My intuition is _using `#map!` is usually done without assignment because it is in-place_.

Any thoughts from @mame regarding to TypeProf?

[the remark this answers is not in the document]
mame: Currently, TypeProf performs weak-update of the element type of a container type when its method accepts a new type at a contravariance occurrence, i.e.,

```
ary = [1, 2, 3]
ary.map!(&:to_s)
p ary #=> Array[Integer | String]
```
https://mame.github.io/typeprof-playground/#rb=ary+%3D+%5B1%2C+2%2C+3%5D%0Aary.map%21%28%26%3Ato_s%29%0Ap+ary+%23%3D%3E+Array%5BInteger+%7C+String%5D&rbs=

Changing the RBS declaration as proposed will break this behavior of TypeProf.

That being said, I have no strong opinion about the change. I'm unsure how much is used a call to `map!` that changes the element type. If there are a lot of such cases, prohibiting them could be very conservative and annoying. @hjwylde do you have any statistics?
BTW, I guess the return type of `map!` is not important. IMO it could be even `void`.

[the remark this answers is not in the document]
hjwylde: I'm afraid I don't have any statistics. I too don't feel strongly about the return type of `map!`, so I will revert this one, and make the appropriate changes to the other points in this review - thank you.

> TypeProf performs weak-update of the element type of a container type when its method accepts a new type

This is interesting, perhaps we can adopt something similar too. Initially I had disregarded the idea as I thought it would lead to issues with `Array#include?` etc., but it seems these types of methods take an `untyped` object and it would be okay.

core/object.rbs
hjwylde: Seems you are correct, done.

soutaro (approved):
Thank you so much! 🎉

PR #730, opened 2021-08-05
Generate RBS from JSON Schema

raosush:
# Description
- This PR contains changes required to support generation of RBS from JSON schema
- Added `rbs-json-schema` CLI base command for usage
- Created `json-schema` folder with `cli.rb` class & `generator.rb` class which generates RBS

# Example

1. 
image

2.
image
image
image


raosush:
- [ ] Add signatures for new classes
- [ ] Add tests for new feature

soutaro (changes requested):
Thank you for the contribution! I post some comments about minor improvements.

The major possible improvement is to refactor `#fetch_type` to accept a `Hash[String, untyped]`, an object representing a type. There are two steps currently, but I feels we can do in one method and the code would be more straightforward.

```json
{ "type": "string" }              // => ::String
{ "type": "string", "enum": ["foo", "bar"] }    // => "foo" | "bar"
{ "$ref": "location.json" }       // => location
{ "type": "array", "item": [{ "type": "integer" }, { "type": "string" }] } // => [Integer, String]

// etc...
```

lib/rbs/json_schema/generator.rb
```diff
@@ -0,0 +1,344 @@
...
+            raise ValidationError.new(message: "Invalid JSON content!\n#{e.full_message}")
+          end
+
+          unless (schema.is_a?(Hash) && (schema.key?("$schema") || schema.key?("$id")))
+            raise ValidationError.new(message: "Invalid JSON Schema: #{schema}")
+          end
+
+          # Parse & generate declarations from `definitions` attribute of the schema
+          # SCHEMA_DEFINITIONS[0] => "definitions"
+          schema.dig(SCHEMA_DEFINITIONS[0])&.each do |name, definition|
```
soutaro, line 68: I don't think using `SCHEMA_DEFINITIONS` constant is good.

* I would simply write a string literal `"definitions"` here
* If you don't want to do that, define constants for each string piece for readability. `DEFINITIONS = "definitions"`

lib/rbs/json_schema/generator.rb
```diff
@@ -0,0 +1,344 @@
...
+
+        if type.nil? && !(schema_keys = schema.keys).empty? && !(schema_keys - SCHEMA_DEFINITIONS).empty?
+          RBS.logger.warn "#{key} => Unsupported attributes: #{schema_keys.join(", ")}"
+        end
+
+        type
+      end
+
+      # Returns corresponding `RBS::Types`
+      def fetch_type(type, name: nil, absolute: nil, prefix: true)
```
soutaro, line 245: It's not clear to me why we need this method. Calling the constructor methods directly would be sufficient?

lib/rbs/json_schema/generator.rb
```diff
@@ -0,0 +1,344 @@
...
+          rescue JSON::ParserError => e
+            # Print error message for invalid JSON content
+            raise ValidationError.new(message: "Invalid JSON content!\n#{e.full_message}")
+          end
+
+          unless (schema.is_a?(Hash) && (schema.key?("$schema") || schema.key?("$id")))
+            raise ValidationError.new(message: "Invalid JSON Schema: #{schema}")
+          end
+
+          # Parse & generate declarations from `definitions` attribute of the schema
```
soutaro, line 66: The problem is that there is no standard for the `definitions` key.

* https://json-schema.org/understanding-json-schema/structuring.html#ref
* https://json-schema.org/understanding-json-schema/structuring.html#id19

`$defs` would be more standard key...

lib/rbs/json_schema/generator.rb
```diff
@@ -0,0 +1,453 @@
...
+  module JSONSchema
+    class Generator
+      attr_reader :options
+      attr_reader :stdout
+      attr_reader :stderr
+      attr_reader :decls
+      attr_reader :generated_schemas
+
+      # SCHEMA DEFINITIONS required for generating RBS from JSON schema
+      SCHEMA_DEFINITIONS = %w(definitions properties type enum const allOf oneOf items $ref $schema title description).freeze
```
soutaro, line 15: I don't think declarating this constant makes much sense. Writing literals directly is better.

lib/rbs/json_schema/generator.rb
```diff
@@ -0,0 +1,453 @@
...
+
+      # Alias declarations for easier referencing
+      Alias = RBS::AST::Declarations::Alias
+      TypeName = RBS::TypeName
```
soutaro, line 19: No need to declare `TypeName` and `Namespace`.

lib/rbs/json_schema/generator.rb
```diff
@@ -0,0 +1,453 @@
...
+
+      # Parse JSON schema & return the `RBS::Types` to be assigned
+      def translate_type(uri, schema)
+        # Default initial value
+        type = nil
+
+        # Check for various schema attributes presence
+        case
+        # SCHEMA_DEFINITIONS[3] => enum => A union of literals
+        when schema.key?(SCHEMA_DEFINITIONS[3])
```
soutaro, line 104: The implementation for `enum` type would be something like the following, without `SCHEMA_DEFINITIONS` constant and `fetch_type` method:

```rb
        when values = schema[:enum]
          unless values.is_a?(Array)
            raise ValidationError.new(......)
          end

          types = values.map do |value|
            case value
            when String, Integer
              Types::Literal.new(literal: value, location: nil)
            when nil
              Types::Bases::Nil.new(location: nil)
            else
              raise ......
            end
          end

          type = Types::Union.new(types: types, location: nil)
```

I believe this is short and understandable enough.

lib/rbs/json_schema/cli.rb
```diff
@@ -0,0 +1,39 @@
...
+        opts.on("--[no-]stringify-keys", "Generate record types with string keys") do |bool|
+          @options[:stringify_keys] = bool
+        end
+      end
+
+      def run(args)
+        OptionParser.new do |opts|
+          setup_initial_options(opts)
+
+          opts.on("-I DIR", "Directory containing a collection of JSON schema from which RBS is to be generated") do |dir|
```
soutaro, line 26: I think passing file paths or directory paths will work.

```rb
generator = JSONSchema::Generator.new(stringify_keys: true, stdout: stdout, stderr: stderr)

ARGV.each do |arg|
  path = Pathname(arg)

  case
  when path.file?
    generator.generate(URI.parse(path.to_s))
  when path.directory?
    path.children.each do |child|
      generator.generate(URI.parse(child.to_s))
    end
  end
end

generator.write_output
```

lib/rbs/json_schema/generator.rb
```diff
@@ -0,0 +1,453 @@
...
+        when schema.key?(SCHEMA_DEFINITIONS[8])
+          begin
+            # Parse URI of `$ref`
+            ref_uri = URI.parse(schema[SCHEMA_DEFINITIONS[8]])
+          rescue URI::InvalidURIError => _
+            raise ValidationError.new(message: "Invalid URI encountered in: #{SCHEMA_DEFINITIONS[8]} = #{schema[SCHEMA_DEFINITIONS[8]]}")
+          end
+
+          resolved_uri = resolve_uri(uri, ref_uri) # Resolve `$ref` URI with respect to current URI
+          if resolved_uri.path == uri.path
```
soutaro, line 222: I don't think we should skip generating definitions for other files, but save the declarations included each JSON files in Hash.

```
attr_reader path_decls: Hash[String,AST::Declarations::Module]
```

Then we can group the declarations of each JSON file in the output.

```ruby
# generate_rbs
@path_decls[uri.path].members << decl if !decl.type.nil?
```

soutaro (changes requested):
The overall structure of finding types looks good.
I would like to request relatively minor fixes around the `#translate_type` method.

soutaro (approved):
👍

Issue #729, opened 2021-08-04
Incompatibility of constant name resolution in ruby and rbs


ksss:
```
$ cat sample.rb
class A
  module B
  end
end

class C < A
  include B # A::B
end

$ ruby sample.rb

$ rbs prototype rb sample.rb > sample.rbs

$ cat sample.rbs
class A
  module B
  end
end

class C < A
  include B
end

$ rbs -I . validate
...(snip)...
Validating class/module definition: `::A`...
Validating class/module definition: `::A::B`...
Validating class/module definition: `::C`...
/Users/ksss/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/rbs-1.3.3/lib/rbs/errors.rb:178:in `check!': t.rbs:7:2...7:11: Could not find mixin: B (RBS::NoMixinFoundError)
```

Is this intentional?

pocke:
Yes, the constant name resolution is intentional. see https://github.com/ruby/rbs/pull/307

> Type name resolution became more compatible with Ruby's constant resolution. It resolves given relative type names based on class/module nesting structure. Note that the type name resolution does not refer inheritance and includes.


And `rbs prototype rb` can't generate the expected RBS, such as `include A::B`, due to the design. It just parses Ruby code but doesn't evaluate, so `rbs prototype rb` doesn't know the module name's ancestors.


By the way, we can improve two points about the issue.

First, we can write a documentation about module resolution. Probably the docs don't mention this incompatibility. A doc should mention it.


Second, `rbs prototype runtime` has the same problem but it could be fixed unlike `prototype rb`. `prototype runtime` evaluates Ruby code, so it can get the module full name.

```bash
$ rbs prototype runtime -R sample.rb A 'A::*' 'C' 'C::*'
class A
  module B
  end
end

class C < A
  include B  # ← It should be `include A::B`, but isn't.
end
```

ksss:
Thank you for your kind answer.
I understood the incompatibility of name resolution in ruby and rbs.

PR #731, opened 2021-08-05
Generate included modules with complete name

raosush:
# Description
- As @pocke sir mentioned in his [comment](https://github.com/ruby/rbs/issues/729#issuecomment-893261119), it is possible to generate included modules with complete name
- This PR contains a minor change to enable generating RBS with included modules being written out with their namespace

# Example
image


soutaro (approved):
Okay. My intuition is to generate qualified name only when necessary, but this fix is good enough that it absolutely fixes problems. 👍

PR #732, opened 2021-08-09
Allow Array#[] to receive Range of optional Integer

pocke:
`Array#[]` and `Array#[]=` accept a Range of optional Integer as the argument in Ruby, but the RBS doesn't allow it unexpectedly.

By the way, `String#[]` already accepts it.


soutaro (approved):
👍

PR #733, opened 2021-08-09
fixing hash yielding of field and value

HoneyryderChuck:
This generates steep errors on hash each yields, as field and value are
spread, and not yielded as a tuple.

Also fixing Array() when called with an enumerable, for which it can
also generate an array.

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -113,6 +113,7 @@ module Kernel : BasicObject
   def self?.Array: (NilClass x) -> [ ]
            | [T] (::Array[T] x) -> ::Array[T]
            | [T] (::Range[T] x) -> ::Array[T]
+           | [T] (_Each[T] x) -> ::Array[T]
            | [K, V] (::Hash[K, V] x) -> ::Array[[K, V]]
            | [T] (T x) -> ::Array[T]
 


core/kernel.rbs
```diff
@@ -113,6 +113,7 @@ module Kernel : BasicObject
   def self?.Array: (NilClass x) -> [ ]
            | [T] (::Array[T] x) -> ::Array[T]
            | [T] (::Range[T] x) -> ::Array[T]
+           | [T] (::Enumerable[T] x) -> ::Array[T]
```
soutaro, line 116: Good point. Looks like `_Each[T]` can replace anything else.

```
def self?.Array: (nil) -> [ ]
               | [T] (_Each[T]) -> ::Array[T]
```

core/hash.rbs
```diff
@@ -391,7 +391,7 @@ class Hash[unchecked out K, unchecked out V] < Object
   #     a is 100
   #     b is 200
   #
-  def each: () { ([ K, V ] arg0) -> untyped } -> self
+  def each: () { (K, V) -> untyped } -> self
```
soutaro, line 394: Passing a tuple is intentional and I believe it's consistent with the implementation.

```
irb(main):001:0> RUBY_VERSION
=> "3.0.0"
irb(main):002:0> {a: 1}.each.to_a
=> [[:a, 1]]
```

HoneyryderChuck: Then maybe it's a problem of `steep` that it doesn't acknonwledge these spreads, is that it? Probably related to https://github.com/ruby/rbs/issues/734#issuecomment-898276685

soutaro (approved):
Thanks.

PR #741, opened 2021-08-14
[scanf] Add RBS file for "scanf"

nvh0412:
Added RBS signatures for the "scanf" package

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,7 @@ gem "rspec"
 gem "racc"
 gem "rubocop"
 gem "rubocop-rubycw"
+gem "scanf"
 gem "json"
 gem "json-schema"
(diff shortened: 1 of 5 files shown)


soutaro:
Hi @nvh0412!

Thank you for opening this PR. The change looks good, but I don't think `scanf` should be a part of this gem. It's not a default gem nor bundled gem since 2.6. Would you mind opening another PR on https://github.com/ruby/gem_rbs_collection ?

nvh0412:
Hi @soutaro 

if so, should we update the list in this spreadsheet https://docs.google.com/spreadsheets/d/199rRB93I16H0k4TGZS3EGojns2R0W1oCsN8UPJzOpyU/edit#gid=1383307992?

PR #743, opened 2021-08-16
[openssl] Add RBS file for "openssl"

HoneyryderChuck:
Added RBS signatures for the "openssl" package

Also adds the following:

* Improve `IO.select` signature for more accurate return type without timeout;
* Improve `SecureRandom.random_number`  signature for more accurate return type based on argument;
* Added `exception: false` option support for `IO#read_nonblock/write_nonblock`;

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -63,6 +63,10 @@ task :validate => :parser do
       lib << "timeout"
     end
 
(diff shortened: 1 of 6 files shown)


soutaro (approved):
👍

PR #745, opened 2021-08-17
fixing Kernel.String sig to accept objects implementing #to_str

### LINKED ISSUE #735: Kernel.String wrong sig
I got an error from Steep while passing a variable typed as `string` to `String(v)` method. Although the error from Steep is probably related to [the problem I tried addressing here](https://github.com/soutaro/steep/pull/396), I think the signature is still wrong, as `Kernel.String` should receive objects implementing `to_str` (`_ToStr` in rbs).

HoneyryderChuck:
Fixes #735

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -127,7 +127,7 @@ module Kernel : BasicObject
 
   def self?.Rational: (Numeric | String | Object x, ?Numeric | String y, ?exception: bool exception) -> Rational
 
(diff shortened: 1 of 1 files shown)


soutaro (approved):
It assumes `#to_s` method.

soutaro:
My bad. `#String` assumes `#to_s`, but `type string = String | _ToStr`...

PR #748, opened 2021-08-17
fixing signature of Kernel.system, to take options and env into account

HoneyryderChuck:
also, added `Kernel.spawn`, which has been left out for some reason.


soutaro:
I was thinking of making the type of `#spawn` precise, but now I don't think it's possible.

HoneyryderChuck:
what suggestions do you have, beyond what's implemented?

soutaro (approved):
I'm good to merge. (Seems like I was confusing with other PR...)

PR #750, opened 2021-08-18
fixing _Reader interface implementations to what's expected

HoneyryderChuck:
in all #read family implementations, length is anything coerced to
integer, and the same applies to buffer (anything coercing to string).

The APIs also allow nil in some cases (i.e. `io.read nil, nil`).

One particular thing which I couldn't narrow down in the interface is
the following (quoting the docs):

> When this method is called at end of file, it returns nil or "",
> depending on length: read, read(nil), and read(0) return "",
> read(positive_integer) returns nil.

I suspect that this should also be defined in the API, as length will
rarely be 0 or nil in most practical cases.

### DIFF
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -31,7 +31,7 @@ interface _Each[out A]
 end
 
 interface _Reader
-  def read: (?int length, ?string outbuf) -> String?
+  def read: (?int? length, ?string outbuf) -> String?
 end
 
 interface _ReaderPartial
--- a/core/file.rbs
+++ b/core/file.rbs
@@ -1042,7 +1042,7 @@ class File::Stat < Object
 
   def rdev_minor: () -> Integer
 
-  def read: (?Integer length, ?String outbuf) -> String
+  def read: (?int? length, ?string outbuf) -> String?
 
   def readable?: () -> bool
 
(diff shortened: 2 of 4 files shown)


core/builtin.rbs
```diff
@@ -31,11 +31,11 @@ interface _Each[out A]
 end
 
 interface _Reader
-  def read: (?int length, ?string outbuf) -> String?
+  def read: (?int? length, ?string? outbuf) -> String?
```
soutaro, line 34:
```suggestion
  def read: (?int? length, ?String outbuf) -> String?
```

I don't think making the type of the second parameter `string?` makes much sense, even though no runtime error raises. Is there a good reason to pass `nil` or non-string object with `#to_str` method?

soutaro (changes requested):
🙏

core/builtin.rbs
HoneyryderChuck: I'm thinking more about adoption/backwards-compatibility, given that there may be code somewhere passing some other object may or may not be nil, which type-checkers will complain about, and which just works. I agree that maybe we shouldn't allow nil from a type-checking perspective though. What do you think should be the best approach? `?string`?

Regarding the `string`, I'd leave it there. I've seen many buffer implementations on top of string objects, that make me believe users will pass smth other than strings there.

soutaro: Ok. It makes sense. `?string` seems the best type here to me.

HoneyryderChuck: done

soutaro (approved):
👍

PR #753, opened 2021-08-18
Fix global_variables.rbs

pocke:
This PR contains two fixes to global_variables.rbs, which is introduced in #749.

* Fix `$-i` type, it stores String or a nil
* Fix `$0` and `$PROGRAM_NAME` document.
  * I brought the doc from https://github.com/ruby/ruby/blob/95aff214687a5e12c3eb57d056665741e734c188/doc/globals.rdoc


ksss:
Thanks ✨

PR #755, opened 2021-08-19
Add Ractor RBS

pocke:
This PR adds RBS for Ractor.

### DIFF
--- a/test/stdlib/Ractor_test.rb
+++ b/test/stdlib/Ractor_test.rb
@@ -1,0 +1,169 @@
+require_relative "test_helper"
+
+class RactorSingletonTest < Test::Unit::TestCase
+  include TypeAssertions
+
+  testing "singleton(::Ractor)"
+
+  def test_count
+    assert_send_type "() -> Integer",
+                     Ractor, :count
+  end
+
+  def test_current
+    assert_send_type "() -> Ractor",
+                     Ractor, :current
+  end
+
+  def test_main
+    assert_send_type "() -> Ractor",
+                     Ractor, :main
+  end
+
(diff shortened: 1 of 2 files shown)


test/stdlib/Ractor_test.rb
```diff
@@ -0,0 +1,173 @@
...
+    assert_send_type "(String, copy: true) -> String",
+                     Ractor, :make_shareable, 'foo', copy: true
+    assert_send_type "(String, copy: false) -> String",
+                     Ractor, :make_shareable, 'foo', copy: false
+    assert_send_type "(String, copy: nil) -> String",
+                     Ractor, :make_shareable, 'foo', copy: nil
+  end
+
+  def test_new
+    # TODO: it raises an error because the proc is not isolated
```
pocke, line 40: I guess we need to make `assert_send_type` Ractor safe to avoid this error, and I guess it is not easy. So I skipped the test case.

test/stdlib/Ractor_test.rb
```diff
@@ -0,0 +1,173 @@
...
+  end
+
+  def test_receive_if
+    Ractor.current.send 42
+    assert_send_type "() { (Integer) -> bool } -> Integer",
+                     Ractor, :receive_if do |n| n == 42 end
+  end
+
+  def test_select
+    r1 = Ractor.new { loop { Ractor.yield 42 } }
```
pocke, line 64: This test code leaks Ractor processes in some places. The leaks don't have a large impact because this test code is short-lived code.

soutaro (approved):
👍

PR #563, opened 2021-01-02
Add signatures for io/wait

vbrazo:
Based on [wait.c](https://github.com/ruby/io-wait/blob/master/ext/io/wait/wait.c).

PR #756, opened 2021-08-20
[io/wait] Add RBS for "io/wait"

HoneyryderChuck:
Picking up the work left in https://github.com/ruby/rbs/pull/563.

The most controversial part is the mismatch between what documentation
says about return types and what actually gets returned (doc says bool,
but IO or nil are actually returned).

### DIFF
--- a/core/io/wait.rbs
+++ b/core/io/wait.rbs
@@ -1,0 +1,33 @@
+class IO
+  # Returns number of bytes that can be read without blocking. Returns zero if no
+  # information available.
+  #
+  def nread: () -> Integer
+
+  # Returns `true` if input available without blocking, or `false`.
+  #
+  def ready?: () -> boolish
+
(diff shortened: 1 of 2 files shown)


soutaro:
> doc says bool, but IO or nil are actually returned

Seems like RDoc is out of date. [JA version of the doc](https://docs.ruby-lang.org/ja/latest/method/IO/i/wait.html) says it returns `self | bool | nil`.

soutaro:
I'm not sure if we can implement a stable test case for the methods...

HoneyryderChuck:
I "bent" the rules a bit to make the tests pass in all rubies. Seems that the behaviour changed 🤷 .

Would it be possible to "pass the ball" to ruby core to fix the return value of these methods, and then rely on rbs test cases to keep it from drifting again? I'm not sure if we can do better here.

soutaro (approved):
Will take over this. Thanks! 👍

PR #761, opened 2021-08-23
fixing Set and CGI signatures

HoneyryderChuck:
* Set: set the same type of generic element as for arrays;
* CGI: added missing modules (Cookie, Util, Escape...)

### DIFF
--- a/stdlib/cgi/0/core.rbs
+++ b/stdlib/cgi/0/core.rbs
@@ -272,6 +272,7 @@ class CGI
   include CGI::Util
 
   extend CGI::Util
+
   # <!--
   #   rdoc-file=lib/cgi/core.rb
   #   - CGI.new(tag_maker) { block }
@@ -352,6 +353,7 @@ class CGI
   #
   def initialize: (?String tag_maker) ?{ (String name, String value) -> void } -> void
                 | (Hash[Symbol, untyped] options_hash) ?{ (String name, String value) -> void } -> void
+
   # <!-- rdoc-file=lib/cgi/core.rb -->
   # Return the accept character set for this CGI instance.
   #
@@ -392,6 +394,7 @@ class CGI
   #
   def accept_charset: () -> String
 
+  # <!-- rdoc-file=lib/cgi/core.rb -->
   # This method is an alias for #http_header, when HTML5 tag maker is inactive.
   #
   # NOTE: use #http_header to create HTTP header blocks, this alias is only
@@ -658,6 +661,7 @@ class CGI
 
   VERSION: String
 
+  # <!-- rdoc-file=lib/cgi/cookie.rb -->
   # Class representing an HTTP cookie.
   #
   # In addition to its specific fields and methods, a Cookie instance is a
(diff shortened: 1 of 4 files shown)


HoneyryderChuck:
@soutaro I see there's already an open PR for `Tempfile`. Is there a solution thought out for delegate classes? Both these and Struct seem to be unaddressed when it comes to dynamic attributes.

stdlib/json/0/json.rbs
```diff
@@ -560,7 +560,7 @@ class Regexp
   def to_json: (?JSON::State state) -> String
 end
 
-class Set[A]
+class Set[unchecked out A]
```
soutaro, line 563: Could you update the type of `json_create?`. It has been broken...

```
def self.json_create: [A] (Hash[String, String | Array[A]] object) -> Set[A]
```

And, the same for other parameterized classes. 🙏

stdlib/set/0/set.rbs
```diff
@@ -40,7 +40,8 @@
 #
 #     - Akinori MUSHA <knu@iDaemons.org> (current maintainer)
 #
-class Set[A]
+class Set[unchecked out A]
```
soutaro, line 43: 👍

stdlib/openssl/0/openssl.rbs
```diff
@@ -1441,9 +1441,9 @@ module OpenSSL
   end
 
   class HMAC
-    def self.digest: (String algo, String key, String data) -> String
+    def self.digest: (String | Digest algo, String key, String data) -> String
```
soutaro, line 1444: 👍

stdlib/cgi/0/core.rbs
```diff
@@ -560,36 +490,523 @@ class CGI
...
   def stdinput: () -> ::IO
 
   # Synonym for $stdout.
   #
   def stdoutput: () -> ::IO
-end
 
-# String for carriage return
-CGI::CR: String
+  CR: String
```
soutaro, line 587: I want these constant declarations have comments... Seems like `annotate-with-rdoc` is still broken.

soutaro (changes requested):
@HoneyryderChuck Could you update the type of `json_create`s?

stdlib/cgi/0/core.rbs
soutaro: I hope https://github.com/ruby/rbs/pull/768 prints the docs.

stdlib/json/0/json.rbs
HoneyryderChuck: done

stdlib/cgi/0/core.rbs
HoneyryderChuck: done, it did 👍

HoneyryderChuck:
> And we are waiting for #767 get merged, right?

I'd say #767 is also dependent on #765 . Nevertheless, I'd be ok with merging it, as it's an acceptable workaround (and I just checked that `sorbet` did the same).

However, it's a hack, as `File` is not the superclass of `Tempfile` (`Delegator` is), and this will include `File` class methods in `Tempfile`, which is wrong.

This type of more "dynamic-ish" typing will become more important to figure out as the stdlib signature set completes, as I think it's relevant for a few of them such as `delegate`, `struct` and `ostruct`  (and `method_missing` delegations, perhaps). I've done a "draft" in #765 on how it could look like internally, but I think that the top-level API should look different and have specific RBS grammar.

HoneyryderChuck:
@soutaro I think this is ready to be merged.

On the proposal above ☝️ , do you have some thoughts?

HoneyryderChuck:
@soutaro ping ☝️

HoneyryderChuck:
@soutaro just rebased the latest changes, can you give it another try?

soutaro (approved):

soutaro:
@HoneyryderChuck Done! Fixed some issues myself and pushed to your repo. Thanks! 🙏

PR #765, opened 2021-08-24
(delegate) Sigs for delegate + proposal for delegator defs in RBS 

HoneyryderChuck:
(tentative) added support for inheriting from delegator classes

    By defining such RBS classes such as:

```ruby
    class Tempfile < Delegator[File]
    ...
```

    with this patch, while building the instance definition, the class
    assigned to the Delegator definition is then merged with the main
    class definitions, which achieves the purpose of inheriting method
    definitions from the delegated class.

    This is a solution based on a stdlib definition however, so it might not
    be the best way to move forward.


stdlib/delegate/0/core.rbs
```diff
@@ -0,0 +1,131 @@
...
+class SimpleDelegator[A] < Delegator[A]
+  @delegate_sd_obj: A
+
+  def __getobj__: () -> A
+
+  def __setobj__: (A obj) -> A
+end
+
+module Kernel
+  def self?.DelegateClass: [A] (Class superclass) ?{ () -> singleton(Delegator) } -> singleton(Delegator)
```
HoneyryderChuck, line 130: Ideally this should be:

```ruby
def self?.DelegateClass: [A] (Class superclass) ?{ () -> singleton(Delegator[A]) } -> singleton(Delegator[A])
```

but the parser can't recognize `singleton(Delegator[A])`.

@soutaro add this along to the other caveats in the description.

HoneyryderChuck:
@soutaro given the recent additions in 2.0 around bounded generics, shall we revisit this? I think we still need special syntax for this, i.e. smth like:

```rbs
class Delegator[T]
  @delegate_sd_obj: T

  # some rbs syntax to say "inherit API from T?
...

PR #774, opened 2021-09-05
Fix steep

soutaro:
* Update the configuration
* Fix type errors
* Skip type checking unsupported code

### DIFF
--- a/Steepfile
+++ b/Steepfile
@@ -1,3 +1,5 @@
+D = Steep::Diagnostic
+
 target :lib do
   signature "sig"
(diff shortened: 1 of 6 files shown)


soutaro:
Now, it shows only one diagnostic:

```
lib/rbs/cli.rb:124:11: [error] Unsupported splat node occurrence
│ Diagnostic ID: Ruby::UnsupportedSyntax
│
└       when *COMMANDS
             ~~~~~~~~~
```

I don't think there are reasonable and supported rewriting for this...

PR #778, opened 2021-09-05
Fix bundler related test for test-bundled-gems

pocke:
This PR fixes bundler related tests fro test-bundled-gems.

The RBS's tests fail on ruby/ruby's CI. https://github.com/ruby/ruby/pull/4809/checks?check_run_id=3516979076

The following three tests fail

* runtime test
  * rbs collection does `require 'bundler'` always, so `if defined?(Bundler)` is not appropriate solution to detect availability of `bunder/setup`.
* cli test for collection
  * `Bundler::LockfileParser` needs `Gemfile` in the current directory to detect `Bundler.root`, but the CLI test doesn't put `Gemfile`.
  * The problem is concealed if the test runs on Bundler environment because it sets `BUNDLE_GEMFILE` env var.
* test with strong_json
  * The test failed if strong_json gem is not available.
  * The problem is concealed on the Bundler environment because strong_json is installed with `bundle install` as a dependency of goodcheck gem.
  * I replaced strong_json with `rbs-amber`.

### DIFF
--- a/test/rbs/cli_test.rb
+++ b/test/rbs/cli_test.rb
@@ -333,6 +333,7 @@ singleton(::BasicObject)
 
           path: #{dir.join('gem_rbs_collection')}
         YAML
+        dir.join('Gemfile').write('')
         dir.join('Gemfile.lock').write(<<~LOCK)
           GEM
             remote: https://rubygems.org/
@@ -403,6 +404,7 @@ singleton(::BasicObject)
 
           path: #{dir.join('gem_rbs_collection')}
         YAML
+        dir.join('Gemfile').write('')
         dir.join('Gemfile.lock').write(<<~LOCK)
           GEM
             remote: https://rubygems.org/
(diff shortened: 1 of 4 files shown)


test/rbs/test/runtime_test_test.rb
```diff
@@ -68,10 +68,10 @@ def move(x: 0, y: 0)
           "RBS_TEST_OPT" => "-I./foo.rbs"
         }
         ruby = ENV['RUBY'] || RbConfig.ruby
-        command_line = if defined?(Bundler)
-                         [ruby, "-rbundler/setup", "-rrbs/test/setup", "sample.rb"]
-                       else
+        command_line = if ENV['RUBY']
```
soutaro, line 71: Seems like we can omit the `ruby` var, but it's up to you. (If you have another fix...)

test/rbs/cli_test.rb
```diff
@@ -333,6 +333,7 @@ def test_collection_install
 
           path: #{dir.join('gem_rbs_collection')}
         YAML
+        dir.join('Gemfile').write('')
```
soutaro, line 336: 🤹

test/rbs/test/runtime_test_test.rb
pocke: Thanks, I removed the `ruby` var. :put_litter_in_its_place:

PR #782, opened 2021-09-08
Add signature `Enumerator::Lazy#force`

ksss:
Ref: https://docs.ruby-lang.org/en/3.0.0/Enumerator/Lazy.html#method-i-force


soutaro (approved):
👍

PR #783, opened 2021-09-09
Add stdlib sigs for io/console and io/console/size

ksss:
I've written rbs for the `io/console` and `io/console/size` library.

ref: https://github.com/ruby/io-console

### DIFF
--- a/stdlib/io-console/0/io-console.rbs
+++ b/stdlib/io-console/0/io-console.rbs
@@ -11,7 +11,7 @@ class IO
   # result will be returned instead of the console IO itself.
   def self.console: () -> File?
                   | (:close) -> nil
-                  | (Symbol sym, *untyped args) -> top
+                  | (Symbol sym, *untyped args) -> untyped
 
   # returns console window size
   #
--- a/stdlib/io-console/0/io-console.rbs
+++ b/stdlib/io-console/0/io-console.rbs
@@ -1,8 +1,8 @@
 class IO
   class ConsoleMode
-    def echo=: ((true | false)) -> (true | false)
-    def raw: (?min: int, ?time: int, ?intr: (true | false)) -> self
-    def raw!: (?min: int, ?time: int, ?intr: (true | false)) -> self
+    def echo=: (bool) -> bool
+    def raw: (?min: int, ?time: int, ?intr: bool) -> self
+    def raw!: (?min: int, ?time: int, ?intr: bool) -> self
   end
 
   # Returns an File instance opened console.
@@ -56,10 +56,10 @@ class IO
 
   # Enables/disables echo back. On some platforms, all combinations of this flags
   # and raw/cooked mode may not be valid.
-  def echo=: ((true | false) flag) -> (true | false)
+  def echo=: (bool flag) -> bool
 
   # Returns `true` if echo back is enabled.
-  def echo?: () -> (true | false)
+  def echo?: () -> bool
 
   def erase_line: (0 | 1 | 2 | nil) -> self
   def erase_screen: (0 | 1 | 2 | 3 | nil) -> self
@@ -67,7 +67,7 @@ class IO
   # Reads and returns a character in raw mode.
   #
   # See IO#raw for details on the parameters.
-  def getch: (?min: int, ?time: int, ?intr: (true | false)) -> String
+  def getch: (?min: int, ?time: int, ?intr: bool) -> String
 
   # Reads and returns a line without echo back.
   # Prints +prompt+ unless it is +nil+.
@@ -115,14 +115,14 @@ class IO
   # special characters.
   #
   # Refer to the manual page of termios for further details.
-  def raw: [T] (?min: int, ?time: int, ?intr: (true | false)) { (self) -> T } -> T
+  def raw: [T] (?min: int, ?time: int, ?intr: bool) { (self) -> T } -> T
 
   # Enables raw mode, and returns `io`.
   #
   # If the terminal mode needs to be back, use `io.raw { ... }`.
   #
   # See IO#raw for details on the parameters.
-  def raw!: (?min: int, ?time: int, ?intr: (true | false)) -> self
+  def raw!: (?min: int, ?time: int, ?intr: bool) -> self
 
   def scroll_backward: (int) -> self
   def scroll_forward: (int) -> self
--- a/stdlib/io-console/0/io-console.rbs
+++ b/stdlib/io-console/0/io-console.rbs
@@ -1,0 +1,137 @@
+class IO
+  class ConsoleMode
+    def echo=: ((true | false)) -> (true | false)
+    def raw: (?min: int, ?time: int, ?intr: (true | false)) -> self
+    def raw!: (?min: int, ?time: int, ?intr: (true | false)) -> self
+  end
+
+  # Returns an File instance opened console.
+  #
+  # If `sym` is given, it will be sent to the opened console with `args` and the
+  # result will be returned instead of the console IO itself.
+  def self.console: () -> File?
+                  | (:close) -> nil
+                  | (Symbol sym, *untyped args) -> top
+
+  # returns console window size
+  #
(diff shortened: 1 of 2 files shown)


ksss:
Oh, CI say `Errno::ENOTTY`...

ksss:
I used pty 🍔

stdlib/io-console/0/io-console.rbs
```diff
@@ -0,0 +1,137 @@
+class IO
+  class ConsoleMode
+    def echo=: ((true | false)) -> (true | false)
```
pocke, line 3: How about using `bool` instead of `true | false`?
https://github.com/ruby/rbs/blob/edfa19dc7517b0d87d44a144ffcdbc40e4883fde/docs/syntax.md#bool-or-boolish

stdlib/io-console/0/io-console.rbs
```diff
@@ -0,0 +1,137 @@
...
+    def raw!: (?min: int, ?time: int, ?intr: (true | false)) -> self
+  end
+
+  # Returns an File instance opened console.
+  #
+  # If `sym` is given, it will be sent to the opened console with `args` and the
+  # result will be returned instead of the console IO itself.
+  def self.console: () -> File?
+                  | (:close) -> nil
+                  | (Symbol sym, *untyped args) -> top
```
pocke, line 14:
```suggestion
                  | (Symbol sym, *untyped args) -> untyped
```

We use `untyped` instead of `top` for types not determine.

stdlib/io-console/0/io-console.rbs
```diff
@@ -0,0 +1,137 @@
...
+
+  # Sets the console mode to `mode`.
+  def console_mode=: (IO::ConsoleMode mode) -> IO::ConsoleMode
+
+  # Yields `self` within cooked mode.
+  #
+  #     STDIN.cooked(&:gets)
+  #
+  # will read and return a line with echo back and line editing.
+  def cooked: () { (self) -> void } -> void
```
pocke, line 42:
```suggestion
  def cooked: [T] () { (self) -> T } -> T
```

Because `IO#cooked` returns the block's returned value.

```ruby
STDIN.cooked {42} # => 42
```

stdlib/io-console/0/io-console.rbs
```diff
@@ -0,0 +1,137 @@
...
+
+  # Flushes input and output buffers in kernel.
+  def ioflush: () -> self
+
+  # Yields `self` with disabling echo back.
+  #
+  #     STDIN.noecho(&:gets)
+  #
+  # will read and return a line without echo back.
+  def noecho: () { (self) -> void } -> void
```
pocke, line 95:
```suggestion
  def noecho: [T] () { (self) -> T } -> T
```

stdlib/io-console/0/io-console.rbs
```diff
@@ -0,0 +1,137 @@
...
+  # received when a read operation is performed. (default: 1)
+  #
+  # The parameter `time` specifies the timeout in *seconds* with a precision of
+  # 1/10 of a second. (default: 0)
+  #
+  # If the parameter `intr` is `true`, enables break, interrupt, quit, and suspend
+  # special characters.
+  #
+  # Refer to the manual page of termios for further details.
+  def raw: (?min: int, ?time: int, ?intr: (true | false)) { (self) -> void } -> void
```
pocke, line 118:
```suggestion
  def raw: [T] (?min: int, ?time: int, ?intr: (true | false)) { (self) -> T } -> T
```

test/stdlib/IO_console_test.rb
```diff
@@ -0,0 +1,95 @@
...
+class IOConsoleTest < Test::Unit::TestCase
+  include TypeAssertions
+
+  library 'io-console'
+  testing "::IO"
+
+  def helper
+    m, s = PTY.open
+  rescue RuntimeError
+    puts "no pty"
```
pocke, line 39:
```suggestion
    omit "no pty"
```

How about using `omit`? It clarifies the test is omitted.

stdlib/io-console/0/io-console.rbs
ksss: Sider of CI say "Prefer boolish over bool for method arguments and block return values"

https://sider.review/gh/repos/174796313/pulls/783#issue-6033330

But here is not intended to boolish.

pocke: Ah, thanks for telling the Sider issue.

I think we can ignore the Sier issue and can use `bool` instead. Because here is not intended to `boolish` as you said.
We don't have to follow Sider issues when the issue has a justification that matches the intention. For instance, in this case, the Sider issue says "When you strictly want true | false." as a justification and we want strictly `true | false`.

So let leave the Sider issue and tell us the issue looks a false positive.


By the way, I think we should consider improving something to avoid the confusion. For example, we can update the justification and CONTRIBUTING.md. I'll try it :muscle:

stdlib/io-console/0/io-console.rbs
ksss: I want to show that all types are possible. It's not that I haven't decided.

This is how I interpreted it.
untyped: No type determined yet.
top: any type is possible.

https://github.com/ruby/rbs/blob/edfa19dc7517b0d87d44a144ffcdbc40e4883fde/docs/syntax.md#void-boolish-or-top
> `top` is anything else.

`IO.console(sym, *args)` can call any method like `Object#send`.

```rb
IO.console(:fileno)
#=> 9

IO.console(:path)
#=> "/dev/tty"
```

Should I still use `untyped` in this case?

stdlib/io-console/0/io-console.rbs
ksss: I didn't know how to write here. Thank you ✨

test/stdlib/IO_console_test.rb
ksss: Sounds great!

I brought this code from https://github.com/ruby/io-console/blob/046e645407d150afc40b6fc67118200b8d932a72/test/io/console/test_io_console.rb#L410-L419 , but had to omit it because I had not checked how it would behave. 

Now that I've confirmed that `omit` works in this context, I'll use `omit`.

stdlib/io-console/0/io-console.rbs
ksss: OK Thanks!
I'll use `bool` instead.

I'll help if there's anything I can do!

ksss:
I resolved all conversation without https://github.com/ruby/rbs/pull/783#discussion_r706766113 .

stdlib/io-console/0/io-console.rbs
pocke: Thanks for your explanation. I understand what you mean. 

I think it should be `top` In principle. If it is `top`, the user needs to cast the returned value explicitly. It is a bit verbose but safer than `untyped`.

```ruby
# path is `top`
path = IO.console(:path)
"foo" + path # type error
```


However `Object#send`'s returned value is `untyped` in RBS. So I'm not sure which is appropriate for this case.

https://github.com/ruby/rbs/blob/f54d88e30f59da75f796a6fdf13e4cd99e9da8a5/core/object.rbs#L649


I have a meeting with @soutaro tomorrow, so I'll ask him about it. Please wait a while :pray:

soutaro: Hi @ksss,

I think it's better to go with `untyped` here. It's almost equivalent with `send`, and thus it should make sense to assume the Ruby developers want the same here.

I agree that there are a discussion about which is better `untyped` or `top` for cases like this. I'm currently assuming that forcing `case-when` every `send` method calls (or anything we have `untyped` today) looks overkill for most of the Ruby developers.

ksss: Thank you @pocke and @soutaro .
I will think about it that way from now on.

ksss:
Resolved all conversation.

pocke (approved):
Thanks!

PR #784, opened 2021-09-09
[Net::HTTP] Update `Net::HTTPRequest` signature

raosush:
# Description
* Update signature of of `Net::HTTPRequest` to accept `URI` as a parameter
* Update `.gitignore` file to ignore configuration files such as `.rvmrc, .ruby-version`, etc irrespective of sub-directory, since `steep` supports the aforementioned Ruby environments


pocke (approved):
Thanks!

PR #788, opened 2021-09-12
Replace RBS::Parser

soutaro:


soutaro:
Wow. The CI jobs become up to 2 times faster. 🚅

PR #790, opened 2021-09-15
Ignore `vendor/bundle/` directories from Git

ybiquitous:
When setting `BUNDLE_PATH: "vendor/bundle"` in the `.bundle/config` file,
I'd like to ignore `vendor/bundle/` directories from Git.


soutaro (approved):
🙈

PR #792, opened 2021-09-15
Allow tuple `basename` for `Tempfile.{new,create}` methods

ybiquitous:
This change allows the `basename` parameter receives a 2-strings tuple for the `Tempfile.new` and `Tempfile.create` methods.

The `Tempfile.new` doc says:

> The `basename` parameter is used to determine the name of the temporary file.
> You can either pass a String or an Array with 2 String elements. In the former
> form, the temporary file's base name will begin with the given string. In the
> latter form, the temporary file's base name will begin with the array's first
> element, and end with the second element. For example:

### DIFF
--- a/stdlib/tempfile/0/tempfile.rbs
+++ b/stdlib/tempfile/0/tempfile.rbs
@@ -95,8 +95,8 @@ class Tempfile < File
   #        # ... do something with f ...
   #     end
   #
-  def self.create: (?String basename, ?String? tmpdir, ?mode: Integer, **untyped) -> File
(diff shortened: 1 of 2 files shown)


pocke (approved):
LGTM :star2:

pocke:
By the way, I've confirmed the `tmpdir` signature doesn't have the same problem :heavy_check_mark: 
https://github.com/ruby/rbs/blob/a402b008bc989fc74aa75fcf65dcef1fed06e42f/stdlib/tmpdir/0/tmpdir.rbs#L51-L52

ybiquitous:
@pocke Thanks for the review and merge! 😄

PR #804, opened 2021-09-22
Make Parser::KEYWORDS a hash

soutaro:


test/test_helper.rb
```diff
@@ -37,11 +37,11 @@ def skip_minitest?
     ENV.key?("NO_MINITEST")
   end
 
-  def parse_type(string, variables: Set.new)
+  def parse_type(string, variables: [])
```
soutaro, line 40: Fix incorrect default argument too...

PR #806, opened 2021-09-28
Fix type error with Enumerable

ksss:
steep checks if the module satisfies the method declaration in the interface.
However, in csv.rbs, ModuleSelfTypeError is raised on steep.

```
Module self type constraint in type `::CSV` doesn't satisfy: `::CSV <: ::_Each[untyped]`(RBS::ModuleSelfTypeError) [162, 3]
Module self type constraint in type `::CSV::Row` doesn't satisfy: `::CSV::Row <: ::_Each[untyped]`(RBS::ModuleSelfTypeError) [411, 3]
Module self type constraint in type `::CSV::Table` doesn't satisfy: `::CSV::Table[Elem] <: ::_Each[untyped]`(RBS::ModuleSelfTypeError) [583, 3]
```

It is possible to declare the untyped part of the type in more detail, but in this PR we concentrate on making sure that no errors occur.

### DIFF
--- a/stdlib/csv/0/csv.rbs
+++ b/stdlib/csv/0/csv.rbs
@@ -360,6 +360,49 @@ class CSV < Object
   # output non-ASCII compatible data.
   #
   def self.generate: (?String str, **untyped options) { (CSV csv) -> void } -> String
+
+  # :call-seq:
+  #   csv.each -> enumerator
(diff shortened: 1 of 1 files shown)


soutaro (approved):
👏

PR #807, opened 2021-09-30
Remove leftover documentation about `super`

DmitryPogrebnoy:
Removed leftover documentation of `super` from `syntax.md` at [L292](https://github.com/ruby/rbs/blob/master/docs/syntax.md)


pocke (approved):
Thanks!

PR #808, opened 2021-10-10
Add manifest.yaml for collection to specify stdlib dependencies

pocke:
This PR adds manifest.yaml feature to rbs collection.

# Problem

`rbs collection` resolves the dependencies with Gemfile.lock. It means it couldn't find dependencies that do not appear in Gemfile.lock.

That's a problem in the standard libraries. Even if a gem depends on a standard library, the gem doesn't need a dependency entry of the stdlib in the gemspec.

# Solution

Add a feature to describe standard libraries dependencies. It is `manifest.yaml`.

## What is manifest.yaml

It is a file to describe a library's RBSs.

Currently the file only contains one section, `dependencies`. `rbs collection` installs and loads the `dependencies` also.

## For RBS users

If you uses RBS for your application, you can remove standard libraries from your `rbs_collection.yaml` if the stdlib is available on `manifest.yaml`.

## For library RBS authors

If you write RBSs for libraries, you may need to write `manifest.yaml`.

If the library you'd like to write RBS has stdlib dependencies, you should write the library to `manifest.yaml`.
If the library doesn't have stdlib dependencies, you don't need to write `manifest.yaml`.

You can find examples from this PR or [ruby/gem_rbs_collection](https://github.com/ruby/gem_rbs_collection) repo.

* https://github.com/ruby/gem_rbs_collection/blob/88e86e0b67262f9ab6244a356e81dd9ca8c55b37/gems/activesupport/6.0/manifest.yaml
* https://github.com/ruby/rbs/blob/b1ad16ea6b145ff9f006903f5192104ab84d87dc/stdlib/bigdecimal-math/0/manifest.yaml

Note that you need to put `manifest.yaml` under `sig/` directory if you want to publish your gem with `sig/` directory.

## For RBS tool developers, such as Steep and TypeProf

No actions are required. `rbs collection` automatically loads `manifest.yaml`, so these tools don't need changes.
But if you have any problems, feel free to tell me :raised_hands: 

# Reference

See also my talk at RubyKaigi Takeout 2021.
https://rubykaigi.org/2021-takeout/presentations/p_ck_.html
I told them about the problem and a solution idea.

# Next Action

I'll write `manifest.yaml` for existing gems in ruby/gem_rbs_collection.

### DIFF
--- a/docs/collection.md
+++ b/docs/collection.md
@@ -7,7 +7,6 @@
 * `git(1)`
 * `Gemfile.lock`
 
-
 ## Usage
 
 ### Setup
@@ -65,9 +64,13 @@ Finally the third party RBSs are available! `rbs` commands, such as `rbs validat
 
 ## Configuration
 
+### `rbs_collection.yaml`
+
 Configure `rbs collection` with editing `rbs_collection.yaml`.
 
 ```yaml
+# rbs_collection.yaml
+
 # Download sources.
 # You can add own collection git repository.
 sources:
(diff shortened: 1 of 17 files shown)


soutaro (approved):
👍

PR #809, opened 2021-10-20
Fix return type of Set.[] and Set#flatten

ksss:
The following types are currently a mismatch.

t.rbs
```rb
class FooSet < Set[Integer]
  def self.gen: (*Integer) -> FooSet
end
```

t.rb
```rb
class FooSet < Set
  def self.gen(*arg)
    FooSet[*arg]
  end
end
```

steep check
```
t.rb:2:11:[error] Cannot allow method body have type `::Set[::Integer]` because declared as type `::FooSet`
│   ::Set[::Integer] <: ::FooSet
│     ::Object <: ::FooSet
│       ::BasicObject <: ::FooSet
│
│ Diagnostic ID: Ruby::MethodBodyTypeMismatch
│
└   def self.gen(*arg)
```

This PR fix this issue.

`Set.[]` and `Set#flatten` return instance of inherited class.
I adapted return type to this behavior.

```rb
class FooSet < Set
end

FooSet[1, 2, 3]
#=> #<FooSet: {1, 2, 3}>

FooSet[1, 2, FooSet[3]].flatten
#=> #<FooSet: {1, 2, 3}>
```

### DIFF
--- a/stdlib/set/0/set.rbs
+++ b/stdlib/set/0/set.rbs
@@ -61,7 +61,7 @@ class Set[A]
   #     Set[1, 2, 1]                # => #<Set: {1, 2}>
   #     Set[1, 'c', :s]             # => #<Set: {1, "c", :s}>
   #
-  def self.[]: [X] (*X) -> Set[X]
+  def self.[]: [X] (*X) -> instance
 
   # Returns a new set containing elements common to the set and the given
   # enumerable object.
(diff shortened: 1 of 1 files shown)


soutaro:
@ksss Unfortunately, I don't think your proposal makes more sense than the current definition. 

* The new type of `Set.[]` will make the return type `Set[untyped]`, but we want `Set[X]`.
* The new type of `Set#flatten` will make the return type of method call where `self` is `Set[Array[Integer]]` itself (== `Set[Array[Integer]]`). I don't think this makes more sense.

It is clear that `self` and `instance` types are not as useful as I want, mainly because it cannot be generic.

ksss:
Oh, I was wrong in my understanding.
I'll be handled by redefinition.

Thanks!

PR #812, opened 2021-10-22
Add ENV signature

ksss:
`ENV` is an instance of `Object` class.
But, RBS does not have syntax for extended object.
Therefore, I introduce the temporary defined class `ENVClass`.

### interface _ENV

```
interface _ENV
end
::ENV: _ENV
```

😁 No need to introduce a new class.
😥 interface does not include Enumerable.

### singleton methods

```
class ENV
  def self.[]: (String) -> String?
end
```

😁 No need to introduce a new class.
😥 It's not correct and does not include Enumerable.

### Special implementation

Special implementation in external tool (steep) for `ENV`.

😁 No need to introduce a new class.
😥 Eventually, we'll need a type definition.

### Add syntax for extended object to RBS

(sample)
```
def ENV.[]: (String) -> String?
```

😁 Can be expressed correctly.
😥 Just for the ENV?

### DIFF
--- a/core/env.rbs
+++ b/core/env.rbs
@@ -1,0 +1,881 @@
+# ENV is a hash-like accessor for environment variables.
+#
+# ### Interaction with the Operating System
+#
+# The ENV object interacts with the operating system's environment variables:
+#
+# *   When you get the value for a name in ENV, the value is retrieved from
+#     among the current environment variables.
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Hmm. Thank you for reporting a really interesting problem. I don't know what to do for `ENV`...

Merging this PR for now. 🙏

PR #813, opened 2021-10-23
Special consts should't allow clone(freeze: false)

ksss:
In some special classes, we cannot effectively pass `false` for freeze option, even though it calls `Object#clone`.

https://github.com/ruby/ruby/blob/185c5738211e16f289aa7448823f678348597bb5/test/ruby/test_object.rb#L81-L84
https://github.com/ruby/ruby/blob/185c5738211e16f289aa7448823f678348597bb5/object.c#L354

ref: https://github.com/ruby/rbs/pull/811

### DIFF
--- a/core/false_class.rbs
+++ b/core/false_class.rbs
@@ -37,4 +37,6 @@ class FalseClass
   def |: (nil) -> false
        | (false) -> false
        | (untyped obj) -> true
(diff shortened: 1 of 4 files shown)


soutaro (approved):
👏

PR #814, opened 2021-10-26
Check the argument type with the generator type

ksss:
The following typo argument errors cannot be detected if `untyped`.

```rb
# typo
[1,2,3].count(:odd?)

# expect
[1,2,3].count(&:odd?)
```

I defined the argument type as generator type to solve this problem.

### DIFF
--- a/core/enumerable.rbs
+++ b/core/enumerable.rbs
@@ -63,7 +63,7 @@ module Enumerable[unchecked out Elem]: _Each[Elem]
   # ary.count{ |x| x%2==0 } #=> 3
   # ```
   def count: () -> Integer
-           | (?untyped) -> Integer
+           | (Elem) -> Integer
            | () { (Elem) -> boolish } -> Integer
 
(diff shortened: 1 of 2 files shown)


soutaro (changes requested):
Accepting any argument in `#count` is intentional because of it's runtime behavior. Could you revert it to `untyped`?

core/enumerable.rbs
```diff
@@ -63,7 +63,7 @@ module Enumerable[unchecked out Elem]: _Each[Elem]
   # ary.count{ |x| x%2==0 } #=> 3
   # ```
   def count: () -> Integer
-           | (?untyped) -> Integer
+           | (Elem) -> Integer
```
soutaro, line 66: 👍 for changing to a required argument.

```suggestion
           | (untyped) -> Integer
```

ksss:
You're right.
I guess I was blinded by practicality.
In this case, I should write this if I want to avoid typos.

```rb
[1,2,3].count { |i| i.odd? }
```

PR #815, opened 2021-10-29
Remove a call to undefined function "pp" from C extension

mame:
It is just for debugging, and it does not build on MinGW
```
linking shared-object rbs_extension.so
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: parserstate.o:parserstate.c:(.text+0x2e2): undefined reference to `pp'
collect2.exe: error: ld returned 1 exit status
```


soutaro (approved):
🗑️

PR #816, opened 2021-10-29
Prevent two GCC warnings

mame:
The following warning can be prevented by using `PRINTF_ARGS` macro that
is provided by ruby header files.

```
../../../../ext/rbs_extension/parser.c:90:3: warning: function ‘raise_syntax_error’ might be a candidate for ‘gnu_printf’ format attribute [-Wsuggest-attribute=format]
```

The following one is preveted by adding `rbs_abort()` to an impossible
case.

```
../../../../ext/rbs_extension/parser.c:1499:10: warning: ‘k’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```


soutaro (approved):
👍

PR #819, opened 2021-11-04
Respect logger level for test/setup logger

Adam-Stomski:
Even though we have set :error for logger level we still receive `Setting up hooks for ...` logs.


soutaro (approved):
Thanks! 👍

soutaro:
Ah... The test is broken. I'll fix it myself.

PR #820, opened 2021-11-05
Avoid a mixture of Array#filter and Array#select

kymmt90:
Although `Array#select` is an alias of `Array#filter`, it would be clearer to use only `filter` in the examples for readers.


pocke (approved):
Good improvement!

PR #825, opened 2021-11-13
Remove ruby/signature.rb

pocke:
This patch removes `lib/ruby/signature.rb`.

This file has been deprecated since over a year ago, so I guess anyone doesn't use it.


soutaro (approved):
👻

PR #828, opened 2021-11-17
Use "untyped" as an exceptation of return value of IO#ready?

mame:
In Ruby 2.7, the method returns true|false.
In Ruby 3.0, it returns nil|IO.
Now in Ruby 3.1.0-dev, it returns true|false again.

The CI of ruby/ruby repository now fails due to the type test of the
method, so I'd like to disable the test tentatively.


pocke (approved):
:+1:

PR #830, opened 2021-11-20
Support Complex and BigDecimal for Date#{+,-}

ksss:
Date#{+,-} allows to pass an object with numeric argument and `to_r` method.

https://github.com/ruby/ruby/blob/6d540c1b9844a5832846618b53ce35d12d64deac/ext/date/date_core.c#L5785-L5786

For example, `Complex` and `BigDecimal` can be passed.

```rb
Date.today + Complex(1, 0)
#=> Date

Date.today + BigDecimal("1")
#=> Date
```

To achieve this behavior, the `_ToR` interface was introduced.

And `Integer`, `Rational`, and even `Float` are removed because they satisfy the condition of `(Numeric & _ToR)`.

### DIFF
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -6,6 +6,10 @@ interface _ToInt
   def to_int: -> Integer
 end
 
+interface _ToR
+  def to_r: () -> Rational
+end
+
 interface _ToS
   def to_s: -> String
 end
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Interesting! 👏 👏

PR #835, opened 2021-12-01
Strict include/delete element types

soutaro:
RBS has provided a method type of `include?`/`delete` with `untyped`.

```rbs
class Array[Elem]
  def include?: (untyped) -> bool   # Allows passing object of unrelated types.
end
```

This is more compatible with Ruby runtime behavior.

```rb
a = [1, 2, 3]      # RBS assumes the `a` has type of `Array[Integer]`
a.include?("1")    # Ruby evaluates this line without an error, but returns false.
```

I have found that limiting the type of argument here makes more sense to find possible bugs, and want to change the type.

[This](https://github.com/ruby/rbs/pull/834/files#diff-af649a561a0fc1f8255ec6690d893ecf223fb2b409449790da1bb33fc48cba60L436) is an example of the type error: `Parser::KEYWORDS` is `Hash[String, bot]`, and passing a Symbol to `include?` always returns `false`, which is not what I want to do here.

### DIFF
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -699,7 +699,7 @@ class Array[unchecked out Elem] < Object
   #     ary.count {|x| x%2 == 0}   #=> 3
   #
   def count: () -> ::Integer
-           | (untyped obj) -> ::Integer
+           | (Elem obj) -> ::Integer
            | () { (Elem) -> boolish } -> ::Integer
 
   # Calls the given block for each element `n` times or forever if `nil` is given.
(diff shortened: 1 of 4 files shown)


ksss:
Is this a same problem?
https://github.com/ruby/rbs/pull/814

soutaro:
@ksss Yes, I changed my opinion... And I have to fix the type of `#count` too.

ksss:
Thank you for your consideration of my opinion.

Issue #845, opened 2021-12-15
Missing URI::for and incorrect URI::parse return type


wagenet:
Currently `URI::parse` returns a `URI::Generic`, but it actually would return subclasses which might have additional methods like `URI::HTTP`.

PR #846, opened 2021-12-17
Fix URI signatures

ksss:
Fix https://github.com/ruby/rbs/issues/845

I fixed the definitions of `URI.parse` and `URI.for`.
And I also added the missing subclasses.

CRuby allows users to add classes later using `URI.register_scheme`, but users can override `URI.parse` and `URI.for`.


soutaro (approved):
👏

PR #849, opened 2021-12-17
Polish URI::Generic parameter signatures

ksss:
Many parameters of `URI::Generic` are nilable.

For example, the relative path scheme is nil.
```
URI.parse('./').scheme
#=> nil
```

I loosened the type restrictions for some methods.

### DIFF
--- a/stdlib/uri/0/generic.rbs
+++ b/stdlib/uri/0/generic.rbs
@@ -193,14 +193,14 @@ module URI
     #
     # Creates a new URI::Generic instance from ``generic'' components without check.
     #
-    def initialize: (String scheme, String userinfo, String host, Integer port, String? registry, String path, String? opaque, String query, String fragment, ?untyped parser, ?boolish arg_check) -> URI::Generic
+    def initialize: (String? scheme, String? userinfo, String? host, (String | Integer)? port, nil registry, String? path, String? opaque, String? query, String? fragment, ?untyped parser, ?boolish arg_check) -> void
 
     #
     # Returns the scheme component of the URI.
     #
     #   URI("http://foo/bar/baz").scheme #=> "http"
     #
-    attr_reader scheme: String
+    attr_reader scheme: String?
 
     # Returns the host component of the URI.
     #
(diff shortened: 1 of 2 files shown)


ksss:
CI has failed.
Indeed, the following is true.

```rb
URI.parse("").send(:userinfo=, 'user:pass')
#=> ["user", "pass"]
```

But I think the following result should be written as a type

```rb
URI.parse("").userinfo = 'user:pass'
#=> "user:pass"
```

How should I test this case?

soutaro:
@ksss Wow! 😮 I think the return type of `userinfo=` method is a tuple, and the value of `x.userinfo = y` is a behavior of the Ruby language.

ksss:
@soutaro Thank you for reviewing.

As you said, the Ruby language specification seems to return the argument itself for methods ending with `=`, regardless of the implementation of the method.

I checked with steep, and this behavior is implemented.

(Maybe here?)
https://github.com/soutaro/steep/blob/89d42e06f5a2c77c55756c97a40e1a6c4b8fb3a4/lib/steep/type_construction.rb#L2739-L2744

However, there seems to be a little bit of confusion since it returns a different type than the documentation.

```rb
# Run with the type of `::URI::Generic#userinfo=` before this modification.

# uri: ::URI::Generic
uri = URI::Generic.new(nil, nil, nil, nil, nil, nil, nil, nil, nil)

# ::URI::Generic#userinfo= ~> (::String | nil)
# Sets userinfo, argument is string like 'name:pass'.
# ・(::String? userinfo) -> ::Array[::String | nil]?
userinfo = uri.userinfo = 'user:pass'

# userinfo: (::String | nil)
userinfo
```

I suggest that the returned type should be the argument type.

And how about the following modification to test this behavior?

```diff
diff --git a/test/stdlib/test_helper.rb b/test/stdlib/test_helper.rb
index cfeae24c..aac42bd7 100644
--- a/test/stdlib/test_helper.rb
+++ b/test/stdlib/test_helper.rb
@@ -245,6 +245,10 @@ module TypeAssertions

     begin
       result = spy.wrapped_object.__send__(method, *args, &block)
+      case method.to_s
+      when "[]=", /\w=\Z/
+        result = args.last
+      end
     rescue Exception => exn
       exception = exn
     end
```

soutaro:
@ksss 

> However, there seems to be a little bit of confusion since it returns a different type than the documentation.

It makes sense... I think what we should fix is the implementation of `uri` library.
How about skipping the test of the method for now?

soutaro:
@ksss I merged this PR to include the change to RBS 2.0.0, which will be bundled to Ruby 3.1.0! 🙇‍♂️ 🎉

ksss:
@soutaro Thank you and GJ 🎄

PR #858, opened 2021-12-24
more precise signature for clock_gettime

HoneyryderChuck:
the expected return type is dependent on the unit passed as argument.


soutaro (approved):
👏

PR #864, opened 2021-12-31
URI sig improvements

HoneyryderChuck:
Improving method defs which can receive a generic uri as arg.

Added sigs for new `URI::HTTP` methods added in v0.11.

### DIFF
--- a/stdlib/uri/0/generic.rbs
+++ b/stdlib/uri/0/generic.rbs
@@ -880,7 +880,7 @@ module URI
     #     uri.merge!("/main.rbx?page=1")
     #     uri.to_s  # => "http://my.example.com/main.rbx?page=1"
     #
-    def merge!: (string oth) -> String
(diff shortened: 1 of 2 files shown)


HoneyryderChuck:
How can I fix the CI issue? it seems that I'd have to copy all prev version sigs, how can I just let them merge, or is the right fix to copy files?

HoneyryderChuck:
ping @soutaro

soutaro:
Hi @HoneyryderChuck, 

Could you move to the updated RBS files to `uri/0`? We don't support providing RBS files for multiple versions, since URI is a default gem, gems which is always available even under bundler.

soutaro (changes requested):
🙏

HoneyryderChuck:
@soutaro done

soutaro (approved):
Thanks! 🎉

PR #866, opened 2021-12-31
several openssl fixes

HoneyryderChuck:
fixing pkey and bn initializers, as well as some digest-related APIs.


soutaro (approved):
Thanks! 🙏

PR #874, opened 2022-01-22
Update Random signatures

snaka:
There are a few things I'm not sure about the changes, so it would be helpful if you could check the comments.

### DIFF
--- a/core/random.rbs
+++ b/core/random.rbs
@@ -8,18 +8,71 @@
 #
 # The class method Random.rand provides the base functionality of Kernel.rand
 # along with better handling of floating point values. These are both interfaces
-# to Random::DEFAULT, the Ruby system PRNG.
+# to the Ruby system PRNG.
 #
-# Random.new will create a new PRNG with a state independent of Random::DEFAULT,
-# allowing multiple generators with different seed values or sequence positions
-# to exist simultaneously. Random objects can be marshaled, allowing sequences
-# to be saved and resumed.
+# Random.new will create a new PRNG with a state independent of the Ruby system
+# PRNG, allowing multiple generators with different seed values or sequence
+# positions to exist simultaneously. Random objects can be marshaled, allowing
+# sequences to be saved and resumed.
 #
 # PRNGs are currently implemented as a modified Mersenne Twister with a period
-# of 2**19937-1.
-#
+# of 2**19937-1.  As this algorithm is *not* for cryptographical use, you must
+# use SecureRandom for security purpose, instead of this PRNG.
 class Random < Object
-  include Random::Formatter
+  # Returns a random binary string. The argument `size` specifies the length of
+  # the returned string.
+  #
+  def self.bytes: (Integer size) -> String
+
+  # Returns an arbitrary seed value. This is used by Random.new when no seed value
+  # is specified as an argument.
+  #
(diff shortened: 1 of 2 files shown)


core/random.rbs
```diff
@@ -104,14 +31,16 @@ class Random < Object
...
-  # Alias of Random::DEFAULT.rand.
-  #
   def self.rand: () -> Float
-               | (Integer | ::Range[Integer] max) -> Integer
-               | (Float | ::Range[Float] max) -> Float
+               | (Integer max) -> Integer
+               | (Float max) -> Float
+               | (::Range[Integer] range) -> Integer?
+               | (::Range[Float] range) -> Float?
+
```
snaka, line 39: The original had a definition for `Numeric`, but I am not sure if it is needed.

core/random.rbs
```diff
@@ -129,141 +58,53 @@ class Random < Object
   #     [ rand, rand ]           # => [0.1915194503788923, 0.6221087710398319]
   #
   def self.srand: (?Integer number) -> Integer
-end
-
-# The default Pseudorandom number generator.  Used by class methods of Random.
-#
-#
-Random::DEFAULT: Random
```
snaka, line 137: I ran `rbs prototype runtime --merge Random` and the `Random::DEFAULT` constant was removed from the definition file.

Do I need to specify any options when I run `rbs prototype`?

core/random.rbs
```diff
@@ -129,141 +58,53 @@ class Random < Object
...
-  #     SecureRandom.uuid #=> "62936e70-1815-439b-bf89-8492855a7e6b"
-  #
-  # The version 4 UUID is purely random (except the version). It doesn't contain
-  # meaningful information such as MAC addresses, timestamps, etc.
-  #
-  # The result contains 122 random bits (15.25 random bytes).
-  #
-  # See RFC 4122 for details of UUID.
-  #
-  def uuid: () -> String
```
snaka, line 268: When I run `rbs prototype runtime --merge Random`, the `Random::Formatter` module is removed from the definition file.

Do I need to specify any options when I run `rbs prototype`?

snaka: Removing `Random::Formatter` seems to be causing the test to fail...

core/random.rbs
ksss: In what cases does it return `nil` in `Random.rand`?
The last check seems to have eliminated the possibility of returning nil or false.

https://github.com/ruby/ruby/blob/342e7a094a70d6f90b96262c88177dae32976c85/random.c#L1388-L1389

There is no such description in ruby/spec either.

https://github.com/ruby/spec/blob/a8afb240adcb413994d4de4d33f3f725105811da/core/random/rand_spec.rb

snaka:
Oh, I found that the pull request already existed.
https://github.com/ruby/rbs/pull/470

This pull request will be taken down.
(I'll put the URL of the above PR in the row of the `Random` class in Spreadsheet)

core/random.rbs
snaka: Thank you for your comment and for the helpful information.

As for the part you asked about, I referred to the following document.

https://docs.ruby-lang.org/ja/3.0/method/Random/i/rand.html

I have read the above and confirmed that there is a difference between RDoc and the description.
However, I was not able to follow the C implementation.

So, I wrote the RBS according to the above document for the time being.

Issue #1, opened 2019-04-18
Improve standard library signature coverage


soutaro:
We provide the signatures of a very limited number of standard library classes. We will really appreciate if you write the signature of other classes.

(Note that we are currently working for standard library signatures.)

connorshea:
Sorbet already has type signatures for a lot of the stdlib, have you considered writing something to transform their files into ruby-signature-compatible files to automate this process?

https://github.com/sorbet/sorbet/tree/master/rbi/core

soutaro:
Yes, it should be one of the best ways to improve the coverage. I haven't tried yet.

connorshea:
I know this ended up partially happening in #23, do we have any way of tracking progress for stdlib coverage? Essentially code coverage for type signatures. :)

soutaro:
I have a guide for contributions at https://github.com/ruby/ruby-signature/blob/master/docs/CONTRIBUTING.md . Could you check it? Tell me if you have any questions or feedback.

connorshea:
For a lot of the libraries in the spreadsheet, we should be able to get the signatures from Sorbet e.g. for [`csv`](https://github.com/sorbet/sorbet/blob/master/rbi/stdlib/csv.rbi) or [`benchmark`](https://github.com/sorbet/sorbet/blob/master/rbi/stdlib/benchmark.rbi).

soutaro:
Yes, it should be a good option to start having an initial version. However, I believe we need to review and revise them ourselves anyway.

gaurav2699:
Hey, is this issue still open? I would like to work on this issue.

Harshit2929:
Is this issue still open?

snaka:
Hi.
It seems that @gaurav2699 's question hasn't been answered until now, so I'd like to make sure.

Is it correct that this issue is still active?

Also, have you already implemented the idea of importing the sorbet type definition file?

I'm hesitant to start a contribution because I personally can't read the status of this issue.

Thank you for your time.

snaka:
I' ve read the code and the PR, and it seems to be still open, so I'll try to help as much as I can.
Thank you very much.

> Also, have you already implemented the idea of importing the sorbet type definition file?

It seems to have been addressed in the following PR 👍 

https://github.com/ruby/rbs/pull/24

snaka:
I thought it would be better to change the title of this issue to "core and standard library".

Reference : https://github.com/ruby/rbs/issues/876

PR #879, opened 2022-01-26
Add File::Stat test

snaka:
There were a few things that I found annoying, so it would be helpful if you could check them out just in case.

(I also wondered about the following, though it is a minor point)

I ran `bin/annotate-with-rdoc core/file.rbs`, but no annotations were added.
I don't know whether this is the correct behavior, or there is something wrong with my environment or operation.


core/file.rbs
```diff
@@ -1042,8 +1042,6 @@ class File::Stat < Object
 
   def rdev_minor: () -> Integer
 
-  def read: (?int? length, ?string outbuf) -> String?
-
```
snaka, line 1046: `File::Stat#read` does not seem to exist.

I removed the definition of this method as it seemed unneeded.

It seems to have been inherited from the Sorbet RBI file from which this file originated.

https://github.com/sorbet/sorbet/blob/a0c88fe9ce2ce713a1e0b386460704284d50c568/rbi/core/file.rbi#L1822-L1829

core/file.rbs
```diff
@@ -984,7 +984,7 @@ File::Constants::WRONLY: Integer
 class File::Stat < Object
   include Comparable
 
-  def initialize: (String file) -> Object
```
snaka, line 987: As far as I can see from the existing RBS file, the return type of the `initialize` method is `void` in some places.

I believe that the return value of a constructor can only be an instance of the class itself.

Therefore, the return value of the constructor should be `self`.

What are the cases in which it is appropriate to use `void`?

test/stdlib/File_Stat_test.rb
```diff
@@ -0,0 +1,235 @@
...
+  def test_writable_real?
+    assert_send_type "() -> bool",
+                      File::Stat.new(__FILE__), :writable_real?
+  end
+
+  def test_zero?
+    assert_send_type "() -> bool",
+                      File::Stat.new(__FILE__), :zero?
+  end
+end
```
snaka, line 235: I ran `bin/annotate-with-rdoc core/file.rbs`, but no annotations were added.

I don't know whether this is the correct behavior, or whether there is something wrong with my environment or my operation.

core/file.rbs
snaka: Thanks for your comment.
Now I know the meaning of `new`, `allocate` and `initialize` in Ruby.

soutaro (approved):
Thank you! 👏

snaka:
Linked
https://github.com/ruby/rbs/issues/1

PR #880, opened 2022-01-28
Fix FileTest signature and add tests

snaka:
The following method cannot receive IO instance as an argument, so I fixed it.

- `executable?`
- `executable_real?`
- `readable?`
- `readable_real?`
- `writable?`
- `writable_real?`
- `symlink?`

Also, I added tests.

### DIFF
--- a/test/stdlib/FileTest_test.rb
+++ b/test/stdlib/FileTest_test.rb
@@ -1,0 +1,188 @@
+require_relative "test_helper"
+
+class FileTestSingletonTest < Test::Unit::TestCase
+  include TypeAssertions
+
+  testing "singleton(::FileTest)"
+
+  def test_blockdev?
+    assert_send_type  "(::String file_name) -> bool",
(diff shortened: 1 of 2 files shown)


soutaro (approved):
🙏

test/stdlib/FileTest_test.rb
```diff
@@ -0,0 +1,188 @@
...
+  end
+
+  def test_exist?
+    assert_send_type  "(::String file_name) -> bool",
+                      FileTest, :exist?, __FILE__
+    assert_send_type  "(::IO file_name) -> bool",
+                      FileTest, :exist?, io_open
+  end
+
+  def test_exists?
```
soutaro, line 53: `exists?` was [removed in 3.0](https://github.com/ruby/ruby/commit/bf97415c02b11a8949f715431aca9eeb6311add2) (?). Deleting this test case would pass the test.

soutaro:
Closing this PR since #888 includes the change.

PR #885, opened 2022-01-31
Annotate nkf

soutaro:


soutaro:
Annotate stdlib files with ruby `3.1.0`.

PR #891, opened 2022-02-02
[Schema] Fix incorrect URI reference in `typeParam.json`

raosush:
# Description
The current schema definition of `typeParam.json` is an invalid definition. It results in a validation error.
image
The issue is a missing `/` at the beginning of the `fragment` of `URI` specified for `$ref`.
I have added the missing `/` to the respective `URI`s.


soutaro (approved):
🙇‍♂️

PR #905, opened 2022-02-19
Remove CharScanner due to no longer used

pocke:
Remove CharScanner. It is no longer used because the parser is rewritten in C.


soutaro (approved):
🔪

PR #914, opened 2022-03-04
Accept Numeric by Integer#{upto,downto}

pocke:
`Integer#{upto,downto}` accept Float, Rational, and so on.
This PR allows them.

By the way, the type definitions will also accept Complex as the argument but actually it raises an error on Ruby level.
I think it is acceptable for now. They accept the same argument as `Integer#<` ([ref](https://github.com/ruby/ruby/blob/fb928f0ea0e35e06ecc96f592702ff21d28fbdd4/numeric.c#L5603)) and currently `Integer#<` accepts Numeric on RBS level.

### DIFF
--- a/core/integer.rbs
+++ b/core/integer.rbs
@@ -857,8 +857,8 @@ class Integer < Numeric
   #
   # With no block given, returns an Enumerator.
   #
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👏

PR #921, opened 2022-03-10
Add manifest.yaml for rbs gem

pocke:
Fix ruby/typeprof#111

Previously loading rbs gem's RBSs raises an error, which displays "no
Logger found", without specifying the all dependencies in the
rbs_collection.yaml.
Because I forgot adding manifest.yaml for this gem.

This patch adds manifest.yaml for RBS gem to fix this problem.

----

By the way, the dependencies list is duplicated with Steepfile. I should remove the duplication by introducing `rbs collection` for this repository, but I remained the duplication to keep this PR simple. I'll do it soon.

### DIFF
--- a/sig/manifest.yaml
+++ b/sig/manifest.yaml
@@ -1,0 +1,8 @@
+dependencies:
+  - name: logger
+  - name: set
+  - name: pathname
+  - name: json
+  - name: optparse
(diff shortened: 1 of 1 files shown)


soutaro (approved):
👍

PR #923, opened 2022-03-16
Add visibility modifier to method definition json

dleavitt:
It's currently omitted when the ast is serialized to json


soutaro (approved):
Thanks!

PR #927, opened 2022-03-17
[rbs collection] Fix error message to tell gem name when source is not found

takahashim:
In the current `rbs collection` command, if the rbs of gem listed in `rbs_collection.yaml` was not found, the error message did not tell which gem was missing.

This PR modifies the error message to put the name of the missing gem in the error message.

### DIFF
--- a/lib/rbs/collection/installer.rb
+++ b/lib/rbs/collection/installer.rb
@@ -20,7 +20,10 @@ module RBS
 
       private def source_for(config_entry)
         @source_for ||= {}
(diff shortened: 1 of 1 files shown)


pocke:
Thanks for this and many PRs on ruby/gem_rbs_collection repo! I'll lock at the PRs after a while. I'm busy this week so it will be delayed. I'll review PRs next weekend at least. Please wait a moment :pray:

pocke (approved):
Thanks!

PR #931, opened 2022-03-21
Core signature updates

hjwylde:
Hi!

These are just some small signature updates that I came across the past couple of months.

* `Array#[]` marked as nilable. I'm not sure if it was intentionally non-nil before, but given `#fetch` is documented as the "non nil" version that throws an error, I think it makes sense to have `#[]` nilable and if a user doesn't want warnings then they can use `#fetch`.
* Missing superclass in `Enumerator::Chain`, couple of tiny missing signatures in `Object` and `Module`.
* Made some signatures more accurate (`true`/`false` instead of `bool`) for some core methods such as `nil?`. Making these more accurate helps with IDE support - we use the return type in some cases to offer fixes that simplify expressions, so this just helps a lot here (e.g., in the IDE we can offer to replace `!nil` with `true` now).

Hope they look okay - let me know if I should change anything in them, or drop any commits if you don't like the changes.

### DIFF
--- a/core/enumerator.rbs
+++ b/core/enumerator.rbs
@@ -515,7 +515,7 @@ end
 #
 # This type of objects can be created by Enumerable#chain and Enumerator#+.
 #
-class Enumerator::Chain[out Elem] < Object
+class Enumerator::Chain[out Elem] < Enumerator[Elem, void]
   include Enumerable[Elem]
 
   # <!--
--- a/core/false_class.rbs
+++ b/core/false_class.rbs
@@ -27,7 +27,7 @@ class FalseClass
   # statements.
   #
   def ===: (false) -> true
-         | (untyped obj) -> bool
+         | (untyped obj) -> false
 
   # <!--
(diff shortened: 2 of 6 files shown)


core/array.rbs
```diff
@@ -850,7 +850,7 @@ class Array[unchecked out Elem] < Object
   #
   # Array#slice is an alias for Array#[].
   #
-  def []: (int index) -> Elem
+  def []: (int index) -> Elem?
```
soutaro, line 853: This is intentional, following the type of array in TypeScript, which I regret for years. Could you revert?

I'm open to make it strict, but don't think we can do it right now because it will break many typed Ruby code. (Maybe next major release?)

soutaro (changes requested):
Thanks! One change is requested on the type of `Array#[]`.

core/array.rbs
hjwylde: Sure, reverted :). Thanks for the info on why it's intentional.

PR #932, opened 2022-03-22
Resolve method calls that return `self`

ksss:
## Background

I call `rbs prototype` on a number of files defined in my Rails application to keep the type definitions as fresh as possible without having to write them by hand.

## Issues to be resolved

It is a common pattern to call `kernel#freeze` when defining constants.

However, all calls to the `Kernel#freeze` method are currently untyped.

```
cat t.rb
class Foo
  CONST_1 = 'bar'
  CONST_2 = 'baz'.freeze
end

$ rbs prototype rb t.rb
class Foo
  CONST_1: ::String

  CONST_2: untyped
end
```

## Expect

If I can output the type before calling the `#freeze` method, I don't have to modify many constants directly.

```rbs
class Foo
  CONST_1: ::String

  CONST_2: ::String
end
```

## How to resolve

In the case of a method name that returns self in an object instance, the receiver is determined to be an instance that inherits from the object class, and the type of the receiver is returned.

## Side effect

The same conversion applies to method calls with optional method arguments.

## Out of scope

Array or Hash object.

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -597,6 +597,14 @@ module RBS
           BuiltinNames::Array.instance_type(default)
         when :HASH
           BuiltinNames::Hash.instance_type(default, default)
+        when :CALL
+          receiver, method_name, * = node.children
+          case method_name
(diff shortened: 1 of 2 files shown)


soutaro (approved):
❄️

Issue #924, opened 2022-03-17
Prime's signature is broken


ksss:
In `steep check` show `ModuleSelfTypeError` in prime.rbs if import prime signature.

```
../../../../.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/rbs-2.2.2/stdlib/prime/0/prime.rbs:47:2: [error] Module self type constraint in type `::Prime` doesn't satisfy: `::Prime <: ::_Each[::Integer]`
│ Diagnostic ID: RBS::ModuleSelfTypeError
│
└   include Enumerable[Integer]
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

```rb
Prime.each
Prime.instance.each # Ruby::NoMethod (Type `::Prime` does not have method `each`)
```

In prime.rbs defined `each` method with `self?` prefix.
https://github.com/ruby/rbs/blob/47468ad1a5a9de3bc0947184d8db07475472ee99/stdlib/prime/0/prime.rbs#L79-L80

The documentation states the following.
https://github.com/ruby/rbs/blob/master/docs/syntax.md#method-definition
> self? method definition adds two methods: a public singleton method and a private instance method, which is equivalent to module_function in Ruby.

Is `self?` intentional?
Is this an rbs problem? Or is it a steep problem?

soutaro:
Nice catch! 

I'm sorry for the error. This problem is occurred by the change of the semantics of `self?` syntax. It was to define two methods, with the same visibility.

I think we need two `def`s, for public singleton method and public instance method respectively.

ksss:
Thank you for the response.
Fixed ✨

PR #934, opened 2022-03-22
Fix Prime signature

ksss:
Fix https://github.com/ruby/rbs/issues/924

And I reimplement testing code using by generator.

```
$ bundle exec rake 'generate:stdlib_test[Prime]'
```

It will also be possible to run tests on Ruby 3.0 over.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem "goodcheck"
 gem "dbm"
 gem 'digest'
 gem 'tempfile'
+gem "prime"
 
 # Test gems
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -18,6 +18,7 @@ gem "goodcheck"
 gem "dbm"
 gem 'digest'
 gem 'tempfile'
+gem "prime"
```
soutaro, line 21: 🙏

soutaro (approved):
Thanks! 👏

PR #936, opened 2022-03-23
Add method `to_d` (convert into BigDecimal) to some classes

takahashim:
Target classes: Integer, Float, String, BigDecimal, Rational, Complex and NilClass

### DIFF
--- a/stdlib/bigdecimal/0/big_decimal.rbs
+++ b/stdlib/bigdecimal/0/big_decimal.rbs
@@ -1187,6 +1187,18 @@ class BigDecimal < Numeric
   #
   def zero?: () -> bool
 
+  # call-seq:
+  #     a.to_d -> bigdecimal
+  #
+  # Returns self.
+  #
+  #     require 'bigdecimal/util'
+  #
+  #     d = BigDecimal("3.14")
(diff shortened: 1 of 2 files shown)


stdlib/bigdecimal/0/big_decimal.rbs
```diff
@@ -1187,6 +1187,18 @@ class BigDecimal < Numeric
   #
   def zero?: () -> bool
 
+  # call-seq:
+  #     a.to_d -> bigdecimal
+  #
```
ksss, line 1192: Maybe `call-seq` is not necessary.
I think you should use the `rbs annotate` tool.

ksss (commented):
One way to do it is to create `bigdecimal-util` like `bigdecimal-math`, but RBS seems to be a tool that takes convenience over correctness, so this seems to be the way to go.

soutaro (dismissed):
Could you try copying RDoc using `rbs annotate`?

soutaro (changes requested):
Could you try copying RDoc using rbs annotate?

soutaro (approved):
Merging this PR. Will update the docs myself.

PR #937, opened 2022-03-24
Const types as literal types

ksss:
Sorry for multi fix.
All commit includes testing code.

Currently, const(CDECL node) use the same type resolution as the optional arguments of the method(`node_type`).
However, most constants are immutable, and it is unnatural to treat them the same as method optional arguments.

For example

````
E = nil
#=> prototype out `E: untyped?`, But is it useful?
````

I've modified the case of const and optional arguments to be separated, and const are now attempted to be resolved as literals like method return type.


soutaro (approved):
👍

PR #940, opened 2022-03-29
Support if/unless node on prototype rb

ksss:
When starting with `if` or `unless`, it was being untyped.
So I have made it as typed as possible.


soutaro:
Assigning to @pocke. Looks like too adhoc while it would be useful. Not sure about the direction of the tool...

soutaro (approved):
Talked with @pocke, and we agreed to approve this PR. 🎉

PR #943, opened 2022-03-29
Fix return value type for `IO#wait_writable`

koic:
`IO#wait_writable` does not return bool unlike `IO#wait_readable`.
https://docs.ruby-lang.org/en/3.1/IO.html#method-i-wait_writable

### DIFF
--- a/core/io/wait.rbs
+++ b/core/io/wait.rbs
@@ -54,5 +54,5 @@ class IO
   # -->
   # Waits until IO is writable and returns `true` or `false` when times out.
   #
-  def wait_writable: (?Numeric? timeout) -> (self | bool | nil)?
+  def wait_writable: (?Numeric? timeout) -> boolish
 end


soutaro:
Let me confirm:

1. This is the fix of method type, which is not changed since Ruby 3.1.1, and
2. #942 is the fix of document, which is changed since Ruby 3.1.1

RBS is based on the latest release of Ruby, 3.1.1 as of today. So, I will merge this, but won't #942.

Does it make sense?

core/io/wait.rbs
```diff
@@ -54,5 +54,5 @@ class IO
   # -->
   # Waits until IO is writable and returns `true` or `false` when times out.
   #
-  def wait_writable: (?Numeric? timeout) -> (self | bool | nil)?
+  def wait_writable: (?Numeric? timeout) -> self?
```
soutaro, line 57: `boolish` would be better since the doc says `truthy or falsy`.
```suggestion
  def wait_writable: (?Numeric? timeout) -> boolish
```

koic: I get it and update it!

koic:
> 2. https://github.com/ruby/rbs/pull/942 is the fix of document, which is changed since Ruby 3.1.1
>
> RBS is based on the latest release of Ruby, 3.1.1 as of today. So, I will merge this, but won't https://github.com/ruby/rbs/pull/942.

I don't have a strong opinion. OTOH, ruby/io-wait seems to be able to `bundle update` independently of the Ruby versions. So, I think it is possible to fix the document independently of Ruby 3.1.1.
https://github.com/ruby/io-wait/commit/d0721e300a

Either way, that's just my point of view, and I'll leave the decision to the maintainers anyway. Thank you for your maintenance.

soutaro:
> ruby/io-wait seems to be able to bundle update independently of the Ruby versions

Ah... it totally makes sense...
Yes, it's the problem of having gems in `stdlib` directory. So, let me leave it for now.

So, we will update the type now, but will update the doc later. Thanks again! 🙏

PR #952, opened 2022-03-31
Move rb_global_variable

soutaro:
Trying to fix random test failure:

```
/__w/rbs/rbs/lib/rbs/parser_aux.rb:12:in `gsub!': no implicit conversion of Array into String (TypeError)
	from /__w/rbs/rbs/lib/rbs/parser_aux.rb:12:in `_parse_signature'
	from /__w/rbs/rbs/lib/rbs/parser_aux.rb:12:in `parse_signature'
```

### DIFF
--- a/ext/rbs_extension/unescape.c
+++ b/ext/rbs_extension/unescape.c
@@ -19,6 +19,7 @@ void rbs_unescape_string(VALUE string) {
 
   if (!HASH) {
     HASH = rb_hash_new();
+    rb_global_variable(&HASH);
     rb_hash_aset(HASH, rb_str_new_literal("\\a"), rb_str_new_literal("\a"));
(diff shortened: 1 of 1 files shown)


ext/rbs_extension/unescape.c
```diff
@@ -19,6 +19,7 @@ void rbs_unescape_string(VALUE string) {
 
   if (!HASH) {
     HASH = rb_hash_new();
+    rb_global_variable(&HASH);
```
soutaro, line 22: GC may be started from inside `rb_str_new_literal` call where the `HASH` is not placed on the stack. It means the value of `HASH` may be freed before the execution of `rb_global_variable` at the lsat place.

So, moving the `rb_global_variable` here will ensure it is marked before any possible GC start. 💪

PR #962, opened 2022-04-02
[Fixes 961] Use default value also for literal_to_type

### LINKED ISSUE #961: Undefined local variable or method `default'
When running `rbs prototype rb` on class like this:

```ruby
class HelloWorld
  def self.world(str)
      str + 'world'
  end

  def hello
    self.class.world('hello')
  end
end
```

And exception is thrown
```
/usr/local/lib/ruby/gems/3.0.0/gems/rbs-2.3.0/lib/rbs/prototype/rb.rb:545:in `literal_to_type': undefined local variable or method `default'
```

lauratpa:
Fixes #961 

It seems that when this piece of code  https://github.com/ruby/rbs/pull/937/files#diff-207b29733058e35f6d05687cd78c9e8e4aec1d51e86f794ea1f5d5620d66027fR545 was moved to a new place, the `default` value variable got left behind.

This is a naive implementation to fix it.

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -542,7 +542,7 @@ module RBS
           when :freeze, :tap, :itself, :dup, :clone, :taint, :untaint, :extend
             literal_to_type(receiver)
           else
-            default
+            untyped
           end
         else
           untyped
--- a/test/rbs/rb_prototype_test.rb
+++ b/test/rbs/rb_prototype_test.rb
@@ -754,6 +754,32 @@ end
     RBS
   end
 
+  def test_calling_class_method_from_instance
+    parser = RB.new
+
+    rb = <<~'RUBY'
+class HelloWorld
+  def self.world(str)
+    str + 'world'
+  end
+
+  def hello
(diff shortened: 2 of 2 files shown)


ksss (commented):
OMG, It's my mistake. 😱
How about use `untyped` instead of `default`?

lib/rbs/prototype/rb.rb
```diff
@@ -542,7 +542,7 @@ def literal_to_type(node)
           when :freeze, :tap, :itself, :dup, :clone, :taint, :untaint, :extend
             literal_to_type(receiver)
           else
-            default
+            untyped
```
lauratpa, line 545: @ksss Like this? :)

soutaro (approved):
Looks good. Will merge this tonight.
@ksss Let me know if it's not good for you.

lib/rbs/prototype/rb.rb
ksss: Thanks🌸

ksss:
No problem. Thanks 🌸

PR #970, opened 2022-04-05
`\e` is not defined in C90

nobu:
https://github.com/ruby/rbs/runs/5827948727?check_suite_focus=true
```
../../../../ext/rbs_extension/unescape.c(25): warning C4129: e: unrecognized character escape sequence
```


soutaro (approved):
🙇‍♂️

PR #973, opened 2022-04-07
Use -std=gnu99 instead of -std=c99

nurse:
If you use -std=c99, GCC doesn't use optimizations which use compiler
builtin functions instead of C library.

https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
https://bugs.ruby-lang.org/issues/12336

### DIFF
--- a/ext/rbs_extension/extconf.rb
+++ b/ext/rbs_extension/extconf.rb
@@ -1,4 +1,4 @@
 require 'mkmf'
 $INCFLAGS << " -I$(top_srcdir)" if $extmk
-append_cflags ['-std=c99']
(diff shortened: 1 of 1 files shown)


soutaro:
@nurse Seems like it's still failing? https://github.com/ruby/ruby/runs/5869073187?check_suite_focus=true#step:16:301

nobu:
The reason is `test-bundled-gems.rb` runs the built ruby without `fake.rb` to build the gems.
Now these extensions are generated under `.bundle` directory by `make build-exts`, the test can use them.
I’m working on it.

PR #977, opened 2022-04-12
Use Random::Base for super_class of Random

ksss:
Starting with ruby v3.0, Random::Base was created to facilitate algorithm changes,
and the inheritance relationship was changed.
ref: https://github.com/ruby/ruby/pull/3024

Also, since there was no test, I created a new one.

### DIFF
--- a/core/random.rbs
+++ b/core/random.rbs
@@ -20,9 +20,7 @@
 # of 2**19937-1.  As this algorithm is *not* for cryptographical use, you must
 # use SecureRandom for security purpose, instead of this PRNG.
 #
-class Random < Object
-  include Random::Formatter
-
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thank you so much! 🙏

soutaro:
Oh no... What is happening??? How can we have that circular definition in Ruby...

soutaro:
Hmm, the Ruby source code is doing equivalent to the following:

```rb
base = Class.new
Random = Class.new(base)
Random::Base = base
```

ksss:
Yes.

`Random::Base` implemented from https://github.com/ruby/ruby/pull/3024 .
Probably, `Random::Base` is a public API.

PR #978, opened 2022-04-12
Define only name Singleton::SingletonClassMethods

ksss:
`Singleton::SingletonClassMethods` is internal module.
But it does exist.
Therefore, I propose to register only the module name.
This module name will prevent the problem of a non-existent module name when the RBS is mechanically generated.
If this module name does not exist, the developer will have to write code to work around the error.


pocke (approved):
👍

PR #979, opened 2022-04-12
Add Exception for pattern-matching

ksss:
I'm considering `DidYouMean` as a use case.

https://github.com/ruby/did_you_mean/blob/80c74fe3e80da672e6d94a3eaf6ade803c4a22fd/lib/did_you_mean/spell_checkers/pattern_key_name_checker.rb#L5-L8


pocke (approved):
Thanks!

PR #989, opened 2022-05-08
Prefer `IO::Buffer.new` for mutable buffer tests.

ioquatix:


soutaro:
The `Buffer.for` returns immutable instance and the tests which calls mutations raise errors. Using `.new` fixes the tests.

soutaro (approved):
❄️

ioquatix:
Thanks!

PR #1000, opened 2022-05-18
Add *batch* mode to `rbs prototype`

soutaro:
Let `prototype-rb` and `prototype-rbi` to process multiple files at one execution with `--out-dir` option.

```
# Process .rb files in lib dir and save the output in `sig` directory
$ rbs prototype rb --out-dir=sig lib

# Use --base-dir option to specify the base path where the relative path calculation starts from
$ rbs prototype rb --out-dir=sig/test --base-dir=test/rbs test 
```

### DIFF
--- a/lib/rbs/cli.rb
+++ b/lib/rbs/cli.rb
@@ -590,7 +590,7 @@ EOU
 
       case format
       when "rbi", "rb"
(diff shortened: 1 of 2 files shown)


pocke (approved):
LGTM :star2:

PR #1010, opened 2022-05-29
Use manifest.yaml instead of manually for validate task

ksss:
and add missing dependency for `stdlib/cgi`.


soutaro (approved):
Nice improvement! 🙏

PR #1011, opened 2022-05-31
Hash#dig requires at least one key

AaronLasseigne:
The signature for Hash#dig is `hash.dig(key, *identifiers)`. It requires at least one key to be passed and then optionally accepts more identifiers.


soutaro (approved):
Thanks! 🙏

PR #1013, opened 2022-06-02
Load dependencies on -r option also

pocke:
This PR allows you to load dependency for standard libraries loaded with `-r` out of the box.

# Problem

`rbs collection` is the manager of third-party gems' RBSs. It resolves dependencies of gems/stdlibs.
But it requires defining the application dependency with `rbs_collection.yaml`. It means it doesn't resolve the dependencies on `-r`.

So, we needed to resolve the dependencies manually on `-r`.

# Solution

With this change, the dependencies are loaded automatically even if we use `-r`. It uses the same way as `rbs collection` but for `-r`.

RBS-related tools do not need to care about this PR. The dependency will be resolved without any changes for the related tools such as typeprof.

### DIFF
--- a/lib/rbs/environment_loader.rb
+++ b/lib/rbs/environment_loader.rb
@@ -34,16 +34,32 @@ module RBS
       @core_root = core_root
       @repository = repository
 
-      @libs = []
+      @libs = Set.new
       @dirs = []
     end
 
-    def add(path: nil, library: nil, version: nil)
+    def add(path: nil, library: nil, version: nil, resolve_dependencies: true)
       case
       when path
         dirs << path
       when library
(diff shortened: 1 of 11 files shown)


lib/rbs/environment_loader.rb
```diff
@@ -34,16 +34,32 @@ def initialize(core_root: DEFAULT_CORE_ROOT, repository: Repository.new)
       @core_root = core_root
       @repository = repository
 
-      @libs = []
+      @libs = Set.new
```
pocke, line 37: It is necessary to avoid an infinite loop.


I know EnvironmentLoader also checks duplication in `each_decl`, but this change is also necessary to avoid an infinite loop during `add` phase.

PR #1015, opened 2022-06-02
Add support for RSpec's verifying doubles in test setup helper

meqif:
Besides regular doubles, RSpec also provides verifying doubles (created with `instance_double`), which are stricter and check that the stubbed methods actually do exist in the original class or object.

In my experience, they tend to be preferred over regular doubles and thus adding support for them is essential for RBS test hooks to be useful out of the box, without specifying both the test suite and additional classes to be ignored.

All classes listed in https://github.com/rspec/rspec-mocks/blob/v3.11.1/lib/rspec/mocks/verifying_double.rb have been added in this PR.

Please let me know if there's anything I should do before this PR is ready to be merged. Thanks!

References:
- https://relishapp.com/rspec/rspec-mocks/docs/verifying-doubles

### DIFF
--- a/lib/rbs/test/setup_helper.rb
+++ b/lib/rbs/test/setup_helper.rb
@@ -30,7 +30,12 @@ module RBS
 
         case double_suite.downcase.strip
         when 'rspec'
-          ['::RSpec::Mocks::Double']
+          [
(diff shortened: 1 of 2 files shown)


soutaro (approved):
🙏

PR #1016, opened 2022-06-05
Hash#reject does not return self

AaronLasseigne:


pocke:
`Hash#reject` returns a Hash even if the receiver is a child class of Hash since Ruby 2.2.

```ruby
class H < Hash
end

p H.new.reject{}.class # => Hash
```


So it should be `Hash` instead of `self`. Thanks!


---


By the way, I guess we have the same mistakes on the RBS code base, but I haven't investigated yet.

PR #1021, opened 2022-06-06
Lean to the optional side

ksss:
In `gem_rbs_collection`, I think there are a few fixes coming so far that will make the blocks optional.

https://github.com/ruby/gem_rbs_collection/pull/88
https://github.com/ruby/gem_rbs_collection/pull/132
https://github.com/ruby/gem_rbs_collection/pull/135
https://github.com/ruby/gem_rbs_collection/pull/136
https://github.com/ruby/gem_rbs_collection/pull/138
https://github.com/ruby/gem_rbs_collection/pull/165

The cause is due to prototype output.

I feel it is not practical to make the block required.
I suggest leaning towards the optional side and relaxing the requirement.

### DIFF
--- a/lib/rbs/prototype/helpers.rb
+++ b/lib/rbs/prototype/helpers.rb
@@ -12,10 +12,7 @@ module RBS
 
         if block
           method_block = Types::Block.new(
-            # HACK: The `block` is :& on `def m(...)` syntax.
-            #       In this case the block looks optional in most cases, so it marks optional.
-            #       In other cases, we can't determine which is required or optional, so it marks required.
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks! 👍 
I agree that making blocks optional would be a better default.

pocke (approved):
LGTM!

Previously I tried to make the same change, but I didn't merge it. Because it introduces false negatives. The user cannot notice a missing block for a generated method.

But now I agree with optional blocks because the benefit is important even if there is a trade-off.

PR #1024, opened 2022-06-07
Clarify GHA step name for `rake annotate`

pocke:
Previously, when the step is failed, a developer needs to find how to fix the error from the action definition.
By this change, the developer can find the rake task more easily.


pocke:
Ah, it's the 0b10000000000th PR on this repository 🎉

PR #1025, opened 2022-06-07
Support to call collection command in child dir

ksss:
I am inconvenienced by the specification that I have to go to the root directory every time when I run the rbs collection commands.

I propose that the rbs collection command be able to be executed in the child directory, like bundler or rake.

### DIFF
--- a/lib/rbs/cli.rb
+++ b/lib/rbs/cli.rb
@@ -17,7 +17,7 @@ module RBS
 
         @libs = []
         @dirs = []
-        @config_path = Collection::Config::PATH
+        @config_path = Collection::Config.find_config_path || Collection::Config::PATH
       end
 
       def loader
@@ -1001,12 +1001,12 @@ EOB
       case args[0]
       when 'install'
         unless params[:frozen]
-          Collection::Config.generate_lockfile(config_path: config_path, gemfile_lock_path: Pathname('./Gemfile.lock'))
+          Collection::Config.generate_lockfile(config_path: config_path, gemfile_lock_path: Bundler.default_lockfile)
         end
         Collection::Installer.new(lockfile_path: lock_path, stdout: stdout).install_from_lockfile
       when 'update'
         # TODO: Be aware of argv to update only specified gem
-        Collection::Config.generate_lockfile(config_path: config_path, gemfile_lock_path: Pathname('./Gemfile.lock'), with_lockfile: false)
+        Collection::Config.generate_lockfile(config_path: config_path, gemfile_lock_path: Bundler.default_lockfile, with_lockfile: false)
         Collection::Installer.new(lockfile_path: lock_path, stdout: stdout).install_from_lockfile
       when 'init'
         if config_path.exist?
--- a/lib/rbs/collection/config.rb
+++ b/lib/rbs/collection/config.rb
@@ -14,6 +14,15 @@ module RBS
 
       PATH = Pathname('rbs_collection.yaml')
 
+      def self.find_config_path
+        current = Dir.pwd
+        until File.exist?(File.join(current, PATH))
+          return nil if current == File.expand_path("..", current)
+          current = File.expand_path("..", current)
+        end
+        Pathname(current) / PATH
+      end
+
       # Generate a rbs lockfile from Gemfile.lock to `config_path`.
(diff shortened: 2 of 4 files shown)


lib/rbs/cli.rb
```diff
@@ -1001,12 +1001,12 @@ def run_collection(args, options)
       case args[0]
       when 'install'
         unless params[:frozen]
-          Collection::Config.generate_lockfile(config_path: config_path, gemfile_lock_path: Pathname('./Gemfile.lock'))
+          Collection::Config.generate_lockfile(config_path: config_path, gemfile_lock_path: Bundler.default_lockfile)
```
pocke, line 1004: I think it will have unexpected behavior when the directories are different between `rbs_collection.yaml` and `Gemfile.lock`.

For example:

```
parent/
├── Gemfile
├── Gemfile.lock
├── child
│   ├── Gemfile
│   └── Gemfile.lock
└── rbs_collection.yaml
```

In this case, it's ok when the current directory is `parent`.
But it will have unexpected behavior when the current directory is `child`.

In `child`, `rbs` command finds `parent/rbs_collection.yaml` and `parent/child/Gemfile.lock`. But the `parent/rbs_collection.yaml` is generated with `parent/Gemfile.lock`. They're mismatched. So `rbs collection install` command and so on work unexpectedly in the `child` directory.



I have two ideas to avoid the problem.
First, search `Gemfile.lock` only from the same directory as `rbs_collection.yaml`. For example, gemfile_lock_path will be `config_path.dirname.join('Gemfile.lock')`.

Second, validate the directories. We can check the directories for them are the same. If the directories are different, it raises an error.

I prefer the first idea. Sometimes a child directory has a `Gemfile.lock`, so the validation may be too strict.


What do you think?

ksss: @pocke Thank you for reviewing.

You are right. In this case we have a problem with the current change.

We often run commands with `$ bundle exec`.

If we use `$ bundle exec rbs collection` in the children directory, bundler will refer to the Gemfile in the children directory.

In the case of method 1, if the programmer runs `bundle exec rbs collection` in the children directory, the Gemfile that bundler refers to and the Gemfile that rbs refers to will be different. This also seems to create confusion in this case.

In the case of method 2, it would be inconvenient because it would be necessary to generate an unnecessary rbs_collection.yaml.
The fact that the repository has separate Gemfiles means that the parent and child are likely to be different scripts or applications. In some cases, the child script may not need type support.

How about referring to rake as my suggestion for method 3?
In rake, if programmer run `$ bundle exec rake` in the child directory of the situation presented, it will print the absolute path of the referenced Rakefile.

```shell
$ cd /parent/child

$ bundle exec rake -T
(in /parent)
...
```

https://github.com/ruby/rake/blob/73a21161bbd0db02804bbd11606a7529e2a0aaa2/lib/rake/application.rb#L690-L693

pocke: > How about referring to rake as my suggestion for method 3?
In rake, if programmer run $ bundle exec rake in the child directory of the situation presented, it will print the absolute path of the referenced Rakefile.

In your suggestion, how does rbs command behave? Which files is it load from the same directory or different directories?
If it loads the files from different directories, I think the message doesn't solve the problem.

`rbs collection` command generates `rbs_collection.lock.yaml` from `rbs_collection.yaml` and `Gemfile.lock`.
It also updates `rbs_collection.lock.yaml` if `rbs_collection.yaml` and/or `Gemfile.lock` have some updates. It means if `rbs collection` command refers different `Gemfile.lock`, it changes `rbs_collection.lock.yaml` unexpectedly.
So I think it always uses `rbs_collection.yaml` and `Gemfile.lock` from the same directory. But I can understand your concern, which rbs uses different Gemfile.lock between the runtime and `rbs collection`.


Now I also prefer the idea two. As you said the child directory is for another application in most cases. So it will be safer if rbs raises an error for this situation.





----



By the way, I think displaying the message is a good idea. We should add the message despite the behavior.

pocke: Thanks for your reply 🙌 
I'll work on this problem next week 👍

ksss: Sounds great! Thanks!

PR #1027, opened 2022-06-09
Fix method commenting

ksss:
The result of the comment is broken because there is no distinction between a code ending comment and a comment without a code.

------

t.rb

```rb
# Foo doc
module Foo # :nodoc:
  # foo doc
  def foo
  end
end
```

### Before

```rb
module Foo
  # Foo doc
  # :nodoc:
  # foo doc
  def foo: () -> nil
end
```

### After

```rb
# Foo doc
module Foo
  # foo doc
  def foo: () -> nil
end
```


soutaro (approved):
🙇‍♂️

soutaro (approved):
🙏

PR #1030, opened 2022-06-15
Fix decls of output from Prototype::Runtime

ksss:
I have a problem when I give the decls output by `RBS::Prototype:Runtime` to `RBS::Environment`.

The `RBS::Environment` expects `RBS::TypeName` for the const `name`, but it is set to Symbol.


soutaro (approved):
🙏

PR #1033, opened 2022-06-17
Add Time#to_{time,date,datetime} to stdlib/date

ksss:
stdlib `date` define `Time#to_time` `Time#to_date` and `Time#to_datetime`.

https://github.com/ruby/ruby/blob/2c190863239bee3f54cfb74b16bb6ea4cae6ed20/ext/date/date_core.c#L9957-L9959

These methods should be defined in the `date` library.


soutaro (approved):
🙏

PR #1034, opened 2022-06-18
Fix Regexp.new argument

nobu:
`code` must be third argument and `option` cannot be ommitted.
If the second argument is not an Integer, it has been assumed as just true.
Ruby 3.2 will extend the second argument `option`, and unknown flags will cause errors.


soutaro (approved):
🙏

PR #1041, opened 2022-06-20
Move `ENVClass` definition under `RBS::Unnamed`

soutaro:
Some Ruby objects like `ENV` doesn't have corresponding class exposed to user scripts. However, defining such a class is not allowed in RBS.

We introduce `RBS::Unnamed` pseudo module to define classes of those objects. Note that there are no objects defined corresponding to the classes.


core/random.rbs
```diff
@@ -20,8 +20,8 @@
 # of 2**19937-1.  As this algorithm is *not* for cryptographical use, you must
 # use SecureRandom for security purpose, instead of this PRNG.
 #
-class Random < Object
-  include Random::Formatter
+class Random < RBS::Unnamed::Random_Base
+  Base: singleton(RBS::Unnamed::Random_Base)
```
ksss, line 24: Are there plans to make `Random::Base` inheritable?

```rb
class Sample < Random::Base
end
```

I know it is not a priority since there are few examples of its use, but it would be better if it were considered.

ksss (commented):
This is an idea to cut the namespace under `RBS`.
It is a very good idea for `ARGF`, `ENV`, etc.
But `Random::Base` exists as a real Ruby class, so we may need to consider about inheritance.

core/random.rbs
soutaro: Hmm, good point.
Currently defining `Random::Base` as a class would be the solution?

```rbs
class Random::Base < RBS::Unnamed::Random_Base
end
```

But, it's not straightforward...

soutaro: Hmm... Cycle still exists with `Random_Base` and `Random::Formatter`.

soutaro: Because `Random::Formatter` is a module, not class, we cannot inherit a module.

soutaro: So, I would give up for this. We would take one of the two:

1. Ignore `Random::Base` for now
2. Define the module and class as a constant

Neither allows using the classes/module from RBS. (This should be a good motivating example for defining alias of classes/modules.)

soutaro: We can include the `Unnamed::Random_Formatter` module.

ksss: I don't see this as much of a problem in RBS, since superclasses can be omitted during open class.
Thanks for your consideration!

PR #1046, opened 2022-06-23
Improve Logger type signatures related to `level`.

myronmarston:
As shown in the current docs, `level` can be a String, Symbol, or Integer:

https://github.com/ruby/ruby/blob/b180ffa62270dc32bb69167822b4c698def5c8f7/lib/logger.rb#L388-L393

### DIFF
--- a/stdlib/logger/0/logger.rbs
+++ b/stdlib/logger/0/logger.rbs
@@ -500,7 +500,7 @@ class Logger
   # `severity`
   # :   The Severity of the log message.
   #
-  def level=: (Integer | String severity) -> Integer
+  def level=: (Integer | String | Symbol severity) -> Integer
 
   # <!--
   #   rdoc-file=lib/logger.rb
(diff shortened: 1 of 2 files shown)


myronmarston:
> How about using `_ToS` interface?
> 
> https://github.com/ruby/ruby/blob/232e2f598103c8eda37d08913665b72b6f787e3f/lib/logger.rb#L401

If I used that, then I believe code like this would type check:

``` ruby
logger = Logger.new("some/file.log", level: Object.new)
logger.level = Object.new
```

In fact, _any_ object that implements `to_s` (which everything except things that inherit directly from `BasicObject`, IIRC) would type check, but runtime errors will usually result.

An exception is thrown if the string returned by `to_s`, when downcased, doesn't exactly match one of `debug`, `info`, `warn`, `error`, `fatal` or `unknown`.  As a general rule, the _only_ types on which `to_s` returns a string that will match one of those are `String` and `Symbol`.  Of course, someone could override `to_s` on their custom type to have it return one of those values (or someone can monkey patch a built-in type to have it return one of those values...) but in practice I've never seen anyone do that.

In short, while the line you linked to does use `.to_s`, I believe it does that because it's the "least effort" way to make the method support strings or symbols, not because it intends (or expects) to support any arbitrary object that implements `.to_s`.

ksss:
@myronmarston  You are right.

The `_ToS` interface is too broad and the current definition is too narrow.
There is a way to specify literals, but it is downcase and there are too many combinations.
The patch way would be a good balance.

---

I found the original discussion.

https://github.com/ruby/ruby/pull/1101
https://bugs.ruby-lang.org/issues/11695

> support symbol and string log level setting

It is clear from this that only String and Symbol are intended.

soutaro (approved):

PR #1047, opened 2022-06-24
Fix constant name resolving

ksss:
There is a pattern in which class names fail to resolve if they are connected by `::`.
Shouldn't it be possible to resolve names in cases like the following?

### Current

```rb
class Hello < String
  def hello: (String) -> Hello
end

class Hello::World
  def world: () -> World
end

# after run `env.resolve_type_names`

class ::Hello < String
  def hello: (String) -> ::Hello
end

class ::Hello::World
  def world: () -> World #<= cannot resolve
end
```

### Expect

```rb
class Hello < String
  def hello: (String) -> Hello
end

class Hello::World
  def world: () -> World
end

# after run `env.resolve_type_names`

class ::Hello < ::String
  def hello: (::String) -> ::Hello
end

class ::Hello::World
  def world: () -> ::Hello::World #<= resolved
end
```

### DIFF
--- a/lib/rbs/environment.rb
+++ b/lib/rbs/environment.rb
@@ -256,7 +256,11 @@ module RBS
 
       context = (outer + [decl]).each.with_object([Namespace.root]) do |decl, array|
         head = array.first or raise
-        array.unshift(head + decl.name.to_namespace)
+        namespace = decl.name.to_namespace
+        namespace.path.each_with_index do |path, i|
+          part = Namespace.new(path: namespace.path[0, i + 1], absolute: false)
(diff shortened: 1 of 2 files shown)


soutaro:
@ksss In fact, this is expected behavior simulating Ruby.

```rb
class String::Foo
  pp Foo            # => uninitialized constant String::Foo::Foo (NameError)
end
```

However, I'm not sure if someone really wants this behavior.

ksss:
Oh sorry, I didn't know this behavior...
Thanks!

PR #1053, opened 2022-07-01
Add overloads for math operation with rhs being BigDecimal

### LINKED ISSUE #1052: Mathematical operations don't play well with BigDecimal
I am trying to add typings to a ruby library that has a method like:
```ruby
# def my_method: (Integer int) -> BigDecimal
def my_method(int)
  int / BigDecimal(2)
end
```

The current signature of `Integer#/ is:`
```rbs
  def /: (Integer) -> Integer
       | (Float) -> Float
       | (Rational) -> Rational
       | (Complex) -> Complex
```
However, when using BigDecimal, it should be: 
```rbs
  def /: (Integer) -> Integer
       | (Float) -> Float
       | (Rational) -> Rational
       | (Complex) -> Complex
       | (BigDecimal) -> BigDecimal
```
~I'm not an expert in RBS, but it seems we cannot add an overload to an already defined method (or can we?).~
Edit: Found it!

~As Steep don't have the correct typings, it complains about it. rather loudly. I could write `BigDecimal(int) / BigDecimal(2)` to silence it, but if there is an alternative, I'm all for it!~

I'm doing a PR to add the definitions to the `big_decimal.rbs` typings.

piotaixr:
Closes #1052

### DIFF
--- a/stdlib/bigdecimal/0/big_decimal.rbs
+++ b/stdlib/bigdecimal/0/big_decimal.rbs
@@ -1408,6 +1408,18 @@ class Integer
   # See also BigDecimal::new.
   #
   def to_d: () -> BigDecimal
+
+  def /: (BigDecimal) -> BigDecimal
+    | ...
+
+  def *: (BigDecimal) -> BigDecimal
+    | ...
+
+  def +: (BigDecimal) -> BigDecimal
+    | ...
+
+  def -: (BigDecimal) -> BigDecimal
+    | ...
 end
 
 %a{annotate:rdoc:skip}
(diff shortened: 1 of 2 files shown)


pocke:
Could you avoid mixing cosmetic changes? It makes reviewing a bit harder, and the changed format in `.rbs` files is a different style from other `.rbs` files in this repository. So could you revert the cosmetic changes and only leave the core changes?

piotaixr:
Oh! Of course. Sorry. Will update the PR ASAP.
Asking rubymine to format becomes muscle memory at this point :-)

piotaixr:
Updated!

pocke (approved):
Thanks!

PR #1055, opened 2022-07-03
Fix some error on prototype runtime

ksss:
As a milestone, the following commands have been corrected so that errors do not occur.

```
$ bundle exec rbs prototype runtime '*'
```


soutaro (approved):
Thanks!

PR #1056, opened 2022-07-03
Fix Warning.#warn and Kernel.#warn signatures

kachick:
Since Ruby 3.0, Warning and Kernel.#warn can take optional category.
And current RBS Warning.#warn is defined only on instance. This PR aims to fix both.

### Referenced followings

[Warning.#warn accept only 1 String for message](https://github.com/ruby/ruby/blob/cd34f56d450f2310cceaf4c5f34d23eddfda58e8/error.c#L267)
[Category names](https://github.com/ruby/ruby/blob/cd34f56d450f2310cceaf4c5f34d23eddfda58e8/error.c#L3061-L3062)
[deprecated  | experimental | nil](https://github.com/ruby/ruby/blob/cd34f56d450f2310cceaf4c5f34d23eddfda58e8/error.c#L3071-L3078)
[Can't define different signatures in different RUBY_VERSIONS](https://github.com/ruby/rbs/issues/613)

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -2135,7 +2135,7 @@ module Kernel : BasicObject
   # :experimental
   # :   Used for experimental features that may change in future releases.
   #
-  def self?.warn: (*untyped msg, ?uplevel: Integer | nil) -> NilClass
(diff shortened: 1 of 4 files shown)


pocke (approved):

kachick:
Thank you!

PR #1064, opened 2022-07-16
Fixing Module#refine type signature

piotaixr:
The `Module#refine` should allow to pass a `Module` to refine, and not only classes, as the following is legal:
```ruby
module A
  refine Enumerable do
    def test
      42
    end
  end
end
```

Also:
- `refine` accepts a block but does not provide any argument to it
- `refine` returns an instance of `Refinement`
```ruby
a = module A
  refine String do |arg|
    puts arg.class
  end
end
puts a.class
# NilClass
# Refinement
# => nil
```


test/stdlib/test_helper.rb
```diff
@@ -81,7 +81,13 @@ def wrapped_object
...
 
           begin
-            return_value = spy.object.__send__(spy.method_name, *args, &spy_block)
+            if spy_block
+              return_value = spy.object.__send__(spy.method_name, *args) do |*a, **k, &b|
+                spy_block.call(*a, **k, &b)
+              end
+            else
+              return_value = spy.object.__send__(spy.method_name, *args, &spy_block)
+            end
```
piotaixr, line 90: This to go around the fact that [the ruby source](https://github.com/ruby/ruby/blob/master/eval.c#L1398) seems to require the block given to `refine` have to be a proper block a not a `Proc`.

soutaro: Thanks.

I would skip the test rather than adding this code, but we can do that when this actually causes a problem. 💪

soutaro (approved):
🎉

PR #1065, opened 2022-07-16
Make the signature of StringIO#write compatible with _Writer interface

piotaixr:
I don't know how to write the test for the signature change. The way I added it make it pass locally for me (Archlinux/rbenv/ruby3.1.2).

What would be the correct solution?


soutaro (approved):
Thanks! 🙏

PR #1070, opened 2022-07-27
adding IO.pipe sig

HoneyryderChuck:
This was missing from the IO module.

Couldn't find a less verbose way to set the conversion options. Suggestions welcome.


soutaro:
We already have definition of `IO.pipe`.

PR #1074, opened 2022-07-28
Add ERB::Util and ERB::DefMethod signature

ksss:
Add undefined module signature in ERB.


stdlib/erb/0/erb.rbs
```diff
@@ -446,4 +446,99 @@ class ERB
...
+  #
+  module DefMethod
+    # <!--
+    #   rdoc-file=lib/erb.rb
+    #   - def_erb_method(methodname, erb_or_fname)
+    # -->
+    # define *methodname* as instance method of current module, using ERB object or
+    # eRuby file
+    #
+    def self.def_erb_method: (String methodname, (String | ERB) erb_or_fname) -> untyped
```
ksss, line 542: It is practically impossible to call it as an instance method of `def_erb_method`, so I just used `self`.

stdlib/erb/0/erb.rbs
```diff
@@ -446,4 +446,99 @@ class ERB
...
+    #
+    def self?.html_escape: (String str) -> String
+
+    # <!--
+    #   rdoc-file=lib/erb.rb
+    #   - h(s)
+    # -->
+    #
+    alias h html_escape
+    alias self.h self.html_escape
```
ksss, line 477: Do we need the `alias self?.method` syntax?

soutaro (approved):
🙏

Issue #1077, opened 2022-08-03
Type of `self` in block/proc


soutaro:
https://hackmd.io/xLrYaqUtQ1GhgTHODkYypw

This is a feature for `BasicObject#instance_eval` and `scope` in ActiveRecord.

soutaro:
Done!

PR #1078, opened 2022-08-05
Block self binding

soutaro:
#1077


hurricup:
Missing update to https://github.com/ruby/rbs/blob/master/docs/syntax.md

PR #1085, opened 2022-08-13
Update Kernel.raise to take an Exception instance

kejadlen:


pocke:
`Exception` instance implements `_Exception` interface, so the added code looks redundant. Could you describe why you need this change?

kejadlen:
Ack, I thought this was why I was getting a type error on `raise`, but I guess it must be something else. Apologies!

(On further investigation, it appears that only `(String message) -> bot` is typechecking in my code.)

PR #1090, opened 2022-08-22
Towards Rubygems 4.0

soutaro:
💎


soutaro:
This should fix [a test failure on ruby/ruby](https://github.com/ruby/ruby/runs/7943911307?check_suite_focus=true#step:16:2929).

PR #1092, opened 2022-08-23
dbm is no longer bundled since ruby 3.1

nobu:


soutaro:
We cannot drop DBM from RBS for now, because YAML depends on it.

The actual problems are:

* YAML is a gem but included as a standard library in RBS gem
* There is no way to require library RBS files selectively -- loading `yaml` without `yaml/dbm` in this case

PR #1100, opened 2022-08-29
Add frozen_string_literal: true

pocke:
RBS allocates many objects. It makes `steep check` slow.

So this PR reduces String allocations with `frozen_string_literal: true` magic comment.

# Benchmarking

I benchmarked `bin/steep -j1`.

```
before: 
[13.82, 13.87, 13.73].sum / 3 # => 13.806666666666667

after:
[13.43, 13.27, 13.21].sum / 3 # => 13.303333333333333

(1 - 13.303333333333333 / 13.806666666666667) * 100 # => 3.645581844519563
```

3.6% faster :rocket:

I've confirmed the effect with memory_profiler gem. It reduces 163918 strings (from 3711262 to 3547344).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,3 +13,8 @@ Lint/DuplicateMethods:
   Enabled: true
   Include:
     - 'test/**/*_test.rb'
+
+Style/FrozenStringLiteralComment:
+  Enabled: true
+  Include:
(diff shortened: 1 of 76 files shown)


soutaro (approved):
🧊 🙏

PR #1103, opened 2022-09-02
Dedup method comments

### LINKED ISSUE #1054: Duplicate comments?
Following the example:

```
> instance.methods[:gsub].comments.size
=> 4
```

Did I do something wrong? The 4 comments all seem identical?

soutaro:
Fixes #1054

### DIFF
--- a/lib/rbs/definition.rb
+++ b/lib/rbs/definition.rb
@@ -120,7 +120,7 @@ module RBS
       end
 
       def comments
-        @comments ||= _ = defs.map(&:comment).compact
+        @comments ||= defs.map(&:comment).compact.uniq
       end
 
       def annotations
--- a/test/rbs/definition_builder_test.rb
+++ b/test/rbs/definition_builder_test.rb
@@ -454,6 +454,27 @@ end
     end
   end
 
+  def test_build_comment_dedup
+    SignatureManager.new do |manager|
+      manager.files.merge!(Pathname("foo.rbs") => <<-EOF)
+class Hello
+  # doc1
+  def foo: () -> String
+         | (Integer) -> String
+end
+      EOF
+      manager.build do |env|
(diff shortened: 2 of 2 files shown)


ioquatix:
Do you know why duplicates existed in the first place?

soutaro:
I don't think there was a good reason to have it duplicated. It should be just because of a confusion caused by the complicated structure and bad names... 😓

A def declaration with several overloads has duplicated documentations given to the def declaration.

```rbs
# Doc is for def syntax
def foo: () -> String                  (Overload 1)
       | (Integer) -> Integer          (Overload 2)

# Another doc for another def syntax
def foo: (Integer, Integer) -> void    (Overload 3)
       | ...
```

And the `defs` attribute contains the overloads, 1, 2, and 3, not def syntaxes.

This caused the duplication.

ioquatix:
Would it make sense to have a single source of truth for unique documentation?

In your example case, it seems reasonable that there are 2 distinct "documentation" instances.

soutaro:
Not very sure what do you mean with _a single source of truth_. The `RBS::AST::***#comment` would be the source of truth, and this method is a utility to help collect every comments attached to a method.

In the example above, there will be two comments, for `#foo` method.

* `RBS::AST::Members::MethodDefinition` object for the first def has the first doc
  * The second object has the second doc
  * This is the source of truth
* The comments will be distributed to method overloads, so that we can provide _hover_ feature to show the documents from the source code
  * Overload 1 and 2 have the first doc
  * Overload 3 has the second doc
* The `#comments` method collects the docs through overloads
  * This is the reason of duplication: There are three overloads and we got three comments
  * Deduping the comments will result in two comment object: the first doc and second doc

ioquatix:
I'll try it out and see how it works in practice. Thanks for your help.

Issue #872, opened 2022-01-18
Move stdlib/rubygems/ to core/rubygems/


pocke:
I propose to move RubyGems library's RBS from stdlib to core.

In Ruby, `rubygems` is loaded by default unless `--disable-gems` is given.
`--disable-gems` option is provided only for debugging Ruby core. So we can suppose `rubygems` is loaded always.

But currently RBS provides `rubygems` as a standard library, so the user needs to load `rubygems` explicitly in RBS. It is different from Ruby.

So I think RBS should provide `rubygems` as a core library. 

# Cons

* Performance
  * Loading RBSs will be slower by this change to parse/load `rubygems.rbs`.
  * I guess the parsing cost is small so we can ignore it.
* Compatibility
  * If `rubygems` library is used in existing code, it will causes a load error (or something) by this change.
  * For example, `library 'rubygems'` in `Steepfile` will raise an error if `rubygems` library is gone.

soutaro:
Sounds good! 👍

PR #1107, opened 2022-09-10
Move rubygems to core from stdlib

pocke:
Fix https://github.com/ruby/rbs/issues/872

This PR moves `rubygems` to core from stdlib.

This change keeps compatibility. By this change, `library 'rubygems'` in Stepfile become to load nothing but it just displays a warning message.

In the future (maybe RBS v3.0?), we should remove the warning and make `library 'rubygems'` failure.


Steepfile
```diff
@@ -11,10 +11,12 @@ target :lib do
 
   library "set", "pathname", "json", "logger", "monitor", "tsort", "uri", 'yaml', 'dbm', 'pstore', 'singleton', 'shellwords', 'fileutils', 'find', 'digest'
   signature "stdlib/strscan/0/"
-  signature "stdlib/rubygems/0/"
   signature "stdlib/optparse/0/"
   signature "stdlib/rdoc/0/"
 
+  # Please remove the following line after releasing #1107
+  signature "core/rubygems/"
```
pocke, line 18: Steep uses released RBS, so we need to specify the `core/rubygems` directory until this PR is released to RubyGems.

soutaro (approved):
🎉 💎

PR #1109, opened 2022-09-14
Fix `RDoc::Markup` type

paracycle:
`RDoc::Markup` is a class and not a module:
```ruby
> require "rdoc"
=> true
> RDoc::Markup.class
=> Class
```


soutaro (approved):
🙏

Issue #1111, opened 2022-09-18
`rbs parse` prints `(ErrorToken)` `Syntax error` at CR+LF line ending


ParadoxV5:
(Don’t question why I’m so picky, Windows and WSL users are always prone to line-ending inconsistencies, especially newbies who are unaware of the LF convention.)

### To Reproduce
```bash
$ echo -e "module Test\r\nend" > test.rbs
$ rbs parse test.rbs
```

### Expected
No errors on this blank module.

### Actual
```
` (ErrorToken)..1:12: Syntax error: unexpected token for class/module declaration member, token=`
```

### RBS version
`2.6.0`

PR #1112, opened 2022-09-20
Always skip CR explicitly

soutaro:
Fixes https://github.com/ruby/rbs/issues/1111

* Newlines in Windows is `\r\n`, but WSL is `\n`
* This causes a confusion when: the source code is written in Windows and read in WSL

This PR is explicitly ignore `\r` even in non-windows environment.

Note that this workaround is unnecessary when everything is done correctly: the source code is written in Windows and read in Windows because Ruby converts `\r\n` to `\n` automatically when reading with text mode API. (And `rbs parse` reads the source in text mode.) (This assumption is unrealistic especially for newbies, according to #1111 and I agree with it.)

### DIFF
--- a/test/rbs/parser_test.rb
+++ b/test/rbs/parser_test.rb
@@ -641,4 +641,10 @@ RBS
       assert_equal "foo?: String, bar!: Integer", method_type.type.param_to_s
     end
   end
+
+  def test_neline_inconsitency
+    code = "module Test\r\nend"
+
+    RBS::Parser.parse_signature(code)
+  end
 end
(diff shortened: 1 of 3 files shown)


test/rbs/parser_test.rb
```diff
@@ -641,4 +641,10 @@ def test_parse_method_type2
       assert_equal "foo?: String, bar!: Integer", method_type.type.param_to_s
     end
   end
+
+  def test_neline_inconsitency
```
ParadoxV5, line 645: wait.
```suggestion
  def test_newline_inconsitency
```

PR #1113, opened 2022-09-21
[DOC] Tweak the doc for `Process.kill` signature

koic:
Follow https://github.com/ruby/ruby/pull/5928

This also matches the name used in the RBS signature.

```
def self.kill: (Integer | Symbol | String signal, *Integer pids) -> Integer
```

### DIFF
--- a/core/process.rbs
+++ b/core/process.rbs
@@ -485,7 +485,7 @@ module Process
 
   # <!--
   #   rdoc-file=process.c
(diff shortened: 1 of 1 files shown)


ksss:
This comment is automatically generated and does not require a separate documentation update.

koic:
@ksss I get it. Thank you for the useful information!

PR #1114, opened 2022-09-28
Add method overload of OptionParser#on

joker1007:
Add overload signature of `OptionParser#run`.

support following patterns:

- `on(short, long, pat = /.*/, desc = "") {|v| ...} -> self`
- `on(short, long, klass = String, desc = "") {|v| ...} -> self`
- `on(short, long, *rest) {|v| ...} -> self`


soutaro (approved):
Thanks! 🙏

PR #1128, opened 2022-10-23
Relax need for block

ksss:
Deciding that block is essential is difficult and often wrong.

This method of following type signature,

```rb
def foo
  yield if block_given?
end
```

I expected to following.

```
def foo: () ?{ () -> untyped } -> untyped 
```

However, it is actually required because of the existence of yield.

```
def foo: () { () -> untyped } -> untyped 
```

### DIFF
--- a/lib/rbs/prototype/helpers.rb
+++ b/lib/rbs/prototype/helpers.rb
@@ -23,7 +23,7 @@ module RBS
         if body_node
           if (yields = any_node?(body_node) {|n| n.type == :YIELD })
             method_block = Types::Block.new(
-              required: true,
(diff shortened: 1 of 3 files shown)


ksss:
Actual case https://github.com/ruby/gem_rbs_collection/pull/16

ksss:
@soutaro @pocke How about this?

soutaro:
#1180 improved this problem.

ksss:
@soutaro I don't think the problem has been improved.
Please see my comments https://github.com/ruby/rbs/pull/1180#issuecomment-1360942094

PR #1133, opened 2022-10-26
Add missing constants for Socket

ksss:
Refer to https://github.com/ruby/ruby/blob/131c31a9209c61f84d318aa18b61f468f48b8219/ext/socket/mkconstants.rb


soutaro (approved):
🔌

PR #1137, opened 2022-10-27
Add missing error class of IO::Buffer

ksss:
https://github.com/ruby/ruby/blob/5129ca3e056e1ce3189ba39fa311d4d687b97b45/io_buffer.c#L3167-L3171


soutaro (approved):
🙏

PR #1144, opened 2022-11-13
Add more types for Random (fix conflict #470)

unasuke:
## What I do
* Fix types by comments on original pull request
* Merge master into this branch and resolve conflicts

## Why I do
I need changesets of the #470 (especially `Random.urandom`) but original author did not respond mainteiner's review comments from 2 years.


core/random.rbs
```diff
@@ -71,6 +91,7 @@ class Random < RBS::Unnamed::Random_Base
   def self.rand: () -> Float
                | (Integer | ::Range[Integer] max) -> Integer
                | (Float | ::Range[Float] max) -> Float
+               | (::Range[T]) -> T
```
unasuke, line 94: > Yeah, the type we want would be [T <: Numeric] (Range[T]) -> T, but RBS currently doesn't allow writing it.
>
> I think there are two options. [T] (Range[T]) -> T or (Range[Numeric]) -> Numeric. The first option would make sense in my opinion, because testing the code will uncover the problem anyway.
> https://github.com/ruby/rbs/pull/470#discussion_r521191243

Is `[T] (Range[T])` invalid syntax?

pocke: `[T] (Range[T])` is not a valid piece of code. It becomes a valid method type by adding the returned type, like `[T] (Range[T]) -> T` as soutaro said.

By the way, `[T < Numeric] (Range[T]) -> T` is more appropriate because the upper bound type for a type parameter has been implemented.

soutaro (approved):
Thanks! 👍

PR #1146, opened 2022-11-15
Fix SEGV when parse with invalid content

ksss:
I came across a case that SEGV.
After investigating the minimal code, I found the following

```rb
buffer = RBS::Buffer.new(content: 1, name: nil)
RBS::Parser.parse_signature(buffer)
```

I think it should be validated for String before parsing.


soutaro (approved):
Thanks! 👍

Issue #1148, opened 2022-11-16
Stop loading non-library RBS files


soutaro:
Using RBS based tools like `rbs_rails` and `steep` requires adding `gem` entries in `rbs_collection.yaml`.

```yaml
gems:
  - name: rbs
    ignore: true
  - name: steep
    ignore: true
```

This is annoying and causes problems, and it looks wired because the `gems` section is mainly used to *ignore* libraries.

I'd like to propose changing the library RBS loading protocol:

* Skip loading RBS files from gems that is declared in `Gemfile` with `require: false`

`require: false` is usually written for gems that is not loaded from the application, especially in Rails apps, and we can use it as a marker that implies RBS files of the gems are unnecessary.

The steps to load RBS of libraries will be as following:

1. Require gems that is declared in `gems` section of `rbs_collection.yaml` without `ignore: true`
2. Require gems and their dependencies from Bundler context that doesn't have `require: false` (and skip if it's ignored)

With this change, the `rbs_rails` and `steep` users would change their `Gemfile` lines as:

```rb
gem 'steep', require: false
gem 'rbs_rails', require: false
```

It looks better and more intuitive than having `ignore: true` lines in `rbs_collection.yaml`. 😄

ParadoxV5:
> With this change, the `rbs_rails` and `steep` users would change their `Gemfile` lines as:
> ```ruby
> gem 'steep', require: false
> gem 'rbs_rails', require: false
> ```

For real – because I `require` manually the traditional way instead of `Bundler.require`, I omit writing the `require: false` part.

Though, how would this change impact the `gemspec` delegate in the `Gemfile`?
Additionally, how shall the new format treat [`add_development_dependency` and `add_runtime_dependency`](https://guides.rubygems.org/specification-reference/)?

---

An alternate solution is to refactor `rbs_collection.yaml` (to the same or a new file) to explicitly list gem RBSs to *exclude* without declaring `ignore: true` repeatedly. E.g.: `ignore: [rbs, steep]`

soutaro:
@ParadoxV5 Thank you for your feedback!

I know `require: false` is optional because some of us, including me, don't use `Bundler.require`. In this case `require: false` is a RBS collection specific marker. I feel it's acceptable to *abuse* that feature for RBS, but if you want, you can continue having `ignore: true` lines in `rbs_collection.yaml`.

[...]

PR #1149, opened 2022-11-16
Refactor collection

soutaro:
This is a step toward https://github.com/ruby/rbs/issues/1148, it refactor the Ruby code and RBS type definition.

* More strict `Sources::*` method types, to declare if `version` is required or not.
* `Collection::Config::Lockfile` is introduced to make `rbs_collection.lock.yaml` data structure clear.
* `Sources::Git` is refactored that reduces `git checkout` operation as much as possible

The `Collection::*` API changes a lot. Not very sure if we need to bump the major version of the gem...

### DIFF
--- a/lib/rbs/collection/config/lockfile.rb
+++ b/lib/rbs/collection/config/lockfile.rb
@@ -81,24 +81,7 @@ module RBS
           if gems = data["gems"]
             gems.each do |gem|
               src = gem["source"]
-              source =
-                case src["type"]
-                when "git"
-                  # @type var src: Sources::Git::source_entry
-                  Sources::Git.new(
-                    name: src["name"],
-                    revision: src["revision"],
-                    remote: src["remote"],
-                    repo_dir: src["repo_dir"]
-                  )
-                when "stdlib"
-                  Sources::Stdlib.instance
-                when "rubygems"
-                  Sources::Rubygems.instance
-                else
-                  raise
-                end
-
+              source = Sources.from_config_entry(src)
               lockfile.gems[gem["name"]] = {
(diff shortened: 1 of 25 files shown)


pocke:
> The Collection::* API changes a lot. Not very sure if we need to bump the major version of the gem...

I think we don't need to bump the major version. Because `rbs collection` CLI command is only intended to expose end users, I do not intend to expose other APIs such as `RBS::Collection::*` for end users.

lib/rbs/collection/config/lockfile.rb
```diff
@@ -0,0 +1,123 @@
...
+
+          data = {
+            "sources" => sources.each_value.sort_by {|s| s.name }.map do |source|
+              # @type block: git_source_data
+              {
+                "name" => source.name,
+                "remote" => source.remote,
+                "revision" => source.resolved_revision,
+                "repo_dir" => source.repo_dir
+              }
```
pocke, line 50: Is this replaceable with `Source#to_lockfile`?

soutaro: Maybe? But, the `Source#to_lockfile` includes `"type"` while existing lockfiles don't.

pocke: Hmm, I think it's not a problem because it's not a breaking change. Including `type` is more descriptive, and we can keep it more DRY.

soutaro: We will have `type` attribute. We are planning to add another `type` of sources soon. So having `git` type here is nice.

PR #1153, opened 2022-11-17
Fix SEGV when give invalid object to `variables`.

ksss:
I found another pattern to SEGV.

```rb
RBS::Parser.parse_type("bool", variables: 1)
```

This patch raises a `TypeError` and produces the following message

```
wrong argument type Integer (must be array or nil) (TypeError)
```

### DIFF
--- a/ext/rbs_extension/parserstate.c
+++ b/ext/rbs_extension/parserstate.c
@@ -299,6 +299,12 @@ parserstate *alloc_parser(VALUE buffer, int start_pos, int end_pos, VALUE variab
   parser_advance(parser);
 
   if (!NIL_P(variables)) {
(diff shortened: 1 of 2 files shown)


soutaro:
@ksss Thank you very much. 🙏 I'm curious what if we pass an array of something not symbol. 

```ruby
RBS::Parser.parse_type("bool", variables: [1])
```

I'm not sure if I implemented something to handle that case...

soutaro:
Confirmed no SEGV happens. 

```
TypeError: wrong argument type Object (expected Symbol)
```

soutaro (approved):
Thanks!

ksss:
It seems that `SYM2ID` checks if the value is other than a symbol 👍

PR #1166, opened 2022-12-01
Support #detailed_message for RBS::ParsingError

ksss:
`Exception#detailed_message` is a new feature supported from Ruby v3.2.0 and above.
https://bugs.ruby-lang.org/issues/18564

I would like to propose that this feature be applied to RBS.

CLI usage.

```
$ ruby -v
ruby 3.2.0preview2 (2022-09-09 master 35cfc9a3bb) [arm64-darwin21]

$ cat parsing_error.rbs
class Foo
  deff aaa: () -> void
end

$ rbs -I parsing_error.rbs validate --silent
/Users/ksss/src/github.com/ksss/rbs/lib/rbs/parser_aux.rb:17:in `_parse_signature': parsing_error.rbs:2:2...2:6: Syntax error: unexpected token for class/module declaration member, token=`deff` (tLIDENT) (RBS::ParsingError)

    deff aaa: () -> void
    ^^^^

	from /Users/ksss/src/github.com/ksss/rbs/lib/rbs/parser_aux.rb:17:in `parse_signature'
	from /Users/ksss/src/github.com/ksss/rbs/lib/rbs/environment_loader.rb:162:in `block (2 levels) in each_decl'
	from /Users/ksss/src/github.com/ksss/rbs/lib/rbs/environment_loader.rb:132:in `each_file'
        (snip)
```

Script usage.

```
$ bundle exec irb -r rbs
irb(main):001:0> RBS::Parser.parse_type("singleton(foo)")
/Users/ksss/src/github.com/ksss/rbs/lib/rbs/parser_aux.rb:7:in `_parse_type': a.rbs:1:10...1:13: Syntax error: expected one of class/module/constant name, token=`foo` (tLIDENT) (RBS::ParsingError)

  singleton(foo)
            ^^^

	from /Users/ksss/src/github.com/ksss/rbs/lib/rbs/parser_aux.rb:7:in `parse_type'
	from (irb):1:in `<main>'
        (snip)
```

This feature is inspired by [error_highlight](https://github.com/ruby/error_highlight).

Scope is limited to Ruby 3.2.0 or later and `ParsingError` combinations.

To support Ruby 3.1 or earlier, incompatible changes need to be made, so they are out of scope.

It is likely that other errors could also be supported, but I have limited the scope to `ParsingError` only to keep the PR scope small.

### DIFF
--- a/lib/rbs/errors.rb
+++ b/lib/rbs/errors.rb
@@ -20,7 +20,25 @@ module RBS
   class LoadingError < BaseError; end
   class DefinitionError < BaseError; end
 
+  module DetailedMessageable
+    def detailed_message(highlight: false, **)
+      indent = " " * location.start_column
+      marker = "^" * (location.end_column - location.start_column)
+
+      io = StringIO.new
+      io.puts super
+      io.puts
+      io.print "\e[1m" if highlight
+      io.puts "  #{location.buffer.lines[location.end_line - 1]}"
+      io.puts "  #{indent}#{marker}"
+      io.print "\e[m" if highlight
+      io.string
+    end
+  end
+
   class ParsingError < BaseError
+    include DetailedMessageable
+
     attr_reader :location
     attr_reader :error_message
     attr_reader :token_type
--- a/test/rbs/errors_test.rb
+++ b/test/rbs/errors_test.rb
@@ -1,0 +1,56 @@
+require "test_helper"
+
+class RBS::ErrorsTest < Test::Unit::TestCase
+  def buffer(source)
+    RBS::Buffer.new(content: source, name: "test.rbs")
+  end
+
+  def test_parse_signature_with_parsing_error_detailed_message
+    omit "Exception#detailed_message does not supported" unless Exception.method_defined?(:detailed_message)
+
+    assert_raises RBS::ParsingError do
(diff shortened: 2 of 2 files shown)


soutaro:
@ksss Thank you for this PR. Looks great!

Could you explain about the _incompatible changes_ to support Ruby 3.1? It's totally fine to provide this feature for only 3.2+, but introducing incompatible changes also would be ok when we release a major update -- RBS 3.0.

ksss:
An incompatibility in Ruby 3.1 and earlier is a change in the `#to_s` of each error class.
The return value of `#to_s` will be multi-line and tests for matching error messages will fail.

e.g. 
https://github.com/ruby/rbs/blob/57c07bb9abf86e95694224174ade7521b593926a/test/rbs/parser_test.rb#L494-L497

More detailed explanation of the incompatibility can be found here.
https://bugs.ruby-lang.org/issues/18564#Motivation

However, since RBS is a relatively new project, the impact may be minimal.

If we accept this incompatibility, we would put in a branch like error_highlight.

https://github.com/ruby/error_highlight/blob/5275078dc6f33148c0241160bed0b43753aae863/lib/error_highlight/core_ext.rb#L11-L36

PR #1178, opened 2022-12-15
Escaping for keywords is no longer necessary

ksss:
When the Ruby-implemented parser was used, reserved words had to be escaped.
However, when the C-implemented parser was introduced, escaping is no longer required even if the variable is a reserved word.

I removed the legacy of the Ruby-implemented parser in `RBS::Types::Function::Param#to_s`.

### DIFF
--- a/lib/rbs/types.rb
+++ b/lib/rbs/types.rb
@@ -757,11 +757,7 @@ module RBS
 
         def to_s
           if name
-            if Parser::KEYWORDS.include?(name.to_s)
-              "#{type} `#{name}`"
-            else
-              "#{type} #{name}"
-            end
+            "#{type} #{name}"
           else
             "#{type}"
           end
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👏

ksss:
I don't know why fail testing with master-nightly-focal on CI...

```
Error: test_select(IOSingletonTest): Errno::EBADF: Bad file descriptor
test/stdlib/IO_test.rb:114:in `close'
test/stdlib/IO_test.rb:114:in `ensure in test_select'
test/stdlib/IO_test.rb:115:in `test_select'
     112:       IO, :select, [r], [w], [r], 0.5
     113:   ensure
     114:     r.close
  => 115:     w.close
     116:   end
     117: end
     118: 
```

soutaro:
Updated the branch on the top of #1180 that skips the failing test.

PR #1185, opened 2022-12-23
Correct Socket.new signature

tisonkun:
I have a code snippet that works but doesn't fit in this signature:

```ruby
socket = Socket.new(Socket::AF_INET, Socket::SOCK_STREAM, 0)
...
```


tisonkun:
cc @soutaro

soutaro (approved):
🙏

PR #1186, opened 2022-12-25
Add rbs_ prefix to skip function

nurse:
The function name `skip` is too common and it confliects macOS's
libsystem_c.dylib. Though it doesn't cause a problem under two-level
namespaces, it causes a crash with flat_namespace or other environments.


soutaro (approved):
🙏

PR #1187, opened 2022-12-25
Make Array#zip to accept Enumerable

abicky:
This PR supports the last two calls in the following example:

```ruby
Array(3).zip([1] * 3)
Array(3).zip([1] * 3, [2] * 3)
Array(3).zip([1].cycle) # incompatible
Array(3).zip([1].cycle, [2].cycle) # incompatible
```

Here is the output of `steep check` for the preceding script:

```
% cat Steepfile
target :lib do
  signature "sig"
  check "main.rb"

  library "set", "pathname", "json", "logger", "monitor", "tsort", "uri", 'dbm', 'pstore', 'singleton', 'shellwords', 'fileutils', 'find', 'digest'
  signature 'stdlib/yaml/0'
  signature "stdlib/strscan/0/"
  signature "stdlib/optparse/0/"
  signature "stdlib/rdoc/0/"
end
% steep check
# Type checking files:

..................................................................................................................F......................................................................

main.rb:3:0: [error] Cannot find compatible overloading of method `zip` of type `::Array[::Integer]`
│ Method types:
│   def zip: [U] (::Array[U]) -> ::Array[[::Integer, (U | nil)]]
│          | (::Array[untyped], *::Array[untyped]) -> ::Array[::Array[untyped]]
│          | [U] (::Array[U]) { ([::Integer, (U | nil)]) -> void } -> void
│          | (::Array[untyped], *::Array[untyped]) { (::Array[untyped]) -> void } -> void
│
│ Diagnostic ID: Ruby::UnresolvedOverloading
│
└ Array(3).zip([1].cycle) # incompatible
  ~~~~~~~~~~~~~~~~~~~~~~~

main.rb:4:0: [error] Cannot find compatible overloading of method `zip` of type `::Array[::Integer]`
│ Method types:
│   def zip: [U] (::Array[U]) -> ::Array[[::Integer, (U | nil)]]
│          | (::Array[untyped], *::Array[untyped]) -> ::Array[::Array[untyped]]
│          | [U] (::Array[U]) { ([::Integer, (U | nil)]) -> void } -> void
│          | (::Array[untyped], *::Array[untyped]) { (::Array[untyped]) -> void } -> void
│
│ Diagnostic ID: Ruby::UnresolvedOverloading
│
└ Array(3).zip([1].cycle, [2].cycle) # incompatible
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Detected 2 problems from 1 file
```

### DIFF
--- a/test/stdlib/Enumerable_test.rb
+++ b/test/stdlib/Enumerable_test.rb
@@ -83,11 +83,6 @@ class EnumerableTest < StdlibTest
     enumerable.each_entry { |x| x }
   end
 
-  def test_zip
-    enumerable.zip([4,5,6])
-    enumerable.zip([4,5,6]) { |arr| arr.sum }
-  end
-
   def test_chunk
     enumerable.chunk
     enumerable.chunk { |x| x.even? }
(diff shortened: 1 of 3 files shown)


test/stdlib/Enumerable_test.rb
```diff
@@ -83,11 +83,6 @@ def test_each_entry
     enumerable.each_entry { |x| x }
   end
 
-  def test_zip
```
abicky, line 86: I didn't understand why these test methods existed. According to the following PRs, I guess none of the runtime tests are executed intentionally:

* https://github.com/ruby/rbs/pull/320
* https://github.com/ruby/rbs/pull/405

soutaro:
Thank you for opening this PR. I want to keep both definitions of `Array#zip` and `Enumerable#zip`, because at least they have two different implementations. (While I don't know why we need the two implementations for them...)

Also updated the method types with another PR, #1232.

PR #1188, opened 2022-12-26
Remove Encoding#replicate from Ruby 3.3

mame:
https://bugs.ruby-lang.org/issues/18949


soutaro (approved):
🙏

eregon:
Maybe we should remove the signature too? https://github.com/ruby/rbs/pull/1203

PR #1189, opened 2022-12-26
Add signature for RubyVM::AbstractSyntaxTree

ksss:
Added method types for `RubyVM::AbstractSyntaxTree` and `RubyVM::AbstractSyntaxTree::Node`.

RubyVM::AbstractSyntaxTree` itself is marked as experimental, but we believe that there are a certain number of uses, including rbs.

However, The following methods marked as *internal use* have not been added.

- `RubyVM::AbstractSyntaxTree::Node#node_id`
- `RubyVM::AbstractSyntaxTree::Node#script_lines`
- `RubyVM::AbstractSyntaxTree::Node#source`

The documentation is from ruby v3.1, but the types are based on ruby v3.2 code.


soutaro (approved):
🎉 🎉 🎉

soutaro:
Added a commit to make the tests pass.

PR #1191, opened 2022-12-29
Support interface in class/module for Sorter

ksss:
Fixed an error that occurred when an interface was defined in a class/module and could no longer be processed.

Sort order was adapted to the top-level case.


soutaro (approved):
🎉

PR #1193, opened 2022-12-30
Add ripper in stdlib

ksss:
Ripper is the standard Ruby parser and is used in many gems.
The gem I am developing also uses Ripper, and there was a need to use Ripper's RBS in the development of this gem.


soutaro (approved):
Thank you so much! 🎉

PR #1195, opened 2022-12-30
Support Symbol and _ToStr for Regexp#=~

ksss:
`Regexp#match` and `Regexp#match?` already support Symbol and _ToStr.
Fixed so that `Regexp#=~` can also support them.

### DIFF
--- a/core/regexp.rbs
+++ b/core/regexp.rbs
@@ -948,7 +948,7 @@ class Regexp
   #     "  x = y  " =~ /(?<lhs>\w+)\s*=\s*(?<rhs>\w+)/
   #     p lhs, rhs # undefined local variable
   #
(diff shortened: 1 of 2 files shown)


soutaro (approved):
🙏

core/regexp.rbs
```diff
@@ -948,7 +948,7 @@ class Regexp
   #     "  x = y  " =~ /(?<lhs>\w+)\s*=\s*(?<rhs>\w+)/
   #     p lhs, rhs # undefined local variable
   #
-  def =~: (String? str) -> Integer?
+  def =~: (String? | Symbol | _ToStr str) -> Integer?
```
ParadoxV5, line 951:
```suggestion
  def =~: (string? | Symbol) -> Integer?
```

PR #1203, opened 2023-01-06
Remove Encoding#replicate

eregon:
* See https://bugs.ruby-lang.org/issues/18949
* See https://github.com/ruby/ruby/pull/7079

I wish it was faster and simpler to update `rbs`, maybe it should be a default gem?

### DIFF
--- a/core/encoding.rbs
+++ b/core/encoding.rbs
@@ -437,16 +437,6 @@ class Encoding < Object
 
   # <!--
   #   rdoc-file=encoding.c
-  #   - enc.replicate(name) -> encoding
-  # -->
-  # Returns a replicated encoding of *enc* whose name is *name*. The new encoding
-  # should have the same byte structure of *enc*. If *name* is used by another
-  # encoding, raise ArgumentError.
-  #
-  def replicate: (String name) -> Encoding
-
-  # <!--
-  #   rdoc-file=encoding.c
   #   - enc.name -> string
   #   - enc.to_s -> string
   # -->


core/encoding.rbs
```diff
@@ -435,16 +435,6 @@ class Encoding < Object
...
 
-  # <!--
-  #   rdoc-file=encoding.c
-  #   - enc.replicate(name) -> encoding
-  # -->
-  # Returns a replicated encoding of *enc* whose name is *name*. The new encoding
-  # should have the same byte structure of *enc*. If *name* is used by another
-  # encoding, raise ArgumentError.
-  #
-  def replicate: (String name) -> Encoding
```
eregon, line 446: Removing should be fine according to https://github.com/ruby/rbs/blob/master/docs/CONTRIBUTING.md#how-can-we-handle-incompatibilities-of-core-apis-and-standard-libraries-between-rubies

eregon:
Ah https://github.com/ruby/rbs/commit/84526b58a88818661541f10de40074d63f298541 already removes the test, thanks @mame, that might be enough if there is a release with it.

soutaro:
@eregon Thank you for opening this PR, but I plan to keep the definition until the release of Ruby 3.3 (3.3 will actually drop the definition, right?).

Generally, the core library RBS definition targets the latest release of Ruby. And it is 3.2 for now. (While we haven't finished working for 3.2...)

But, If you still have an opinion to drop the definition right now, I'm open in this case. Because, the `#replicate` method is rarely used and it's deprecated already.

eregon:
Yeah it will be removed in Ruby 3.3.0 release, and of course on ruby/ruby master before that.

If it doesn't cause CI failures, it's up to you until when you want to keep the definition.

OK, so I close this PR.
Will there be something to remember to remember to remove it on 3.3.0 release? Maybe the CI checks the signatures all match to an existing method, but only for releases and not master?

PR #1207, opened 2023-01-12
Make `rake stdlib_test` with single process

hsbt:
The current `rake stdlib_test` is always spawning ruby process each tests. It shows over 1800+ lines of logfile at https://github.com/ruby/ruby/actions/runs/3880327335/jobs/6618526842#step:16:316.

I hope to reduce its size. I make `rake stdlib_test` with single process. How about this?

I'm not sure we stay to use `__FILE__` in the future. It seems to enough with stub file instead of test file itself.


soutaro (approved):
👍

PR #1218, opened 2023-01-27
`Module#attr` returns an Array of Symbol after 3.0

ksss:
Since RBS is supposed to have a policy of matching types to the latest Ruby releases, I fixed the type of Module#attr to 3.0 or later.

See also https://bugs.ruby-lang.org/issues/17314


soutaro (approved):
👍

PR #1222, opened 2023-02-01
Skip loading `require: false` gems in collection

soutaro:
Assuming gems declared in `Gemfile` with `require: false` are apps -- not libraries, we can skip loading the RBS files of that gems in collection.

1. The `LockfileGenerator` starts from the gems declared in `Gemfile` without `require: false`, and adds them and their dependencies recursively. 
2. It skips a gem if it's ignored in `rbs_collection.yaml`.

### DIFF
--- a/test/rbs/collection/config_test.rb
+++ b/test/rbs/collection/config_test.rb
@@ -233,65 +233,6 @@ class RBS::Collection::ConfigTest < Test::Unit::TestCase
     end
   end
 
-  def test_generate_lock_from_collection_repository_specified
-    mktmpdir do |tmpdir|
-      config_path = tmpdir / 'rbs_collection.yaml'
-      config_path.write [CONFIG, <<~YAML].join("\n")
-        gems:
-          - name: ast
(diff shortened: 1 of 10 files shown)


test/rbs/collection/config_test.rb
```diff
@@ -212,66 +233,16 @@ def test_generate_lock_from_collection_repository_ignoring
     end
   end
 
-  def test_generate_lock_from_collection_repository_specified
```
soutaro, line 215: We cannot delete this test because a `default` gem may be required without declared in `Gemfile`...

soutaro: This should be ok. If a *non-default* gem is not included in the Bundler context, it cannot be `require`-ed nor `Gem::Specification.find_by_name` cannot find the gem.

If a *non-default* gem is implicitly loaded, it should be included in the `gems` section of `rbs_collection.yaml` or `dependencies:` section of `manifest.yaml` of a gem.

PR #1223, opened 2023-02-04
give `_ToProc#to_proc` a return type

ParadoxV5:
I don’t know what purpose are the interfaces in `builtin.rbs`, but I *do* use them for implicit conversion. The return of `_ToProc#to_proc`, however, is untyped and I cannot see why. This is the contract for implicit conversion to a *block when prefixed with `&`* (`an_array.map(&a_to_proc)`).

### DIFF
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -23,7 +23,7 @@ interface _ToHash[K, V]
 end
 
 interface _ToProc
-  def to_proc: () -> untyped
+  def to_proc: () -> Proc
 end
 
 interface _ToPath


ParadoxV5:
[Discussion] just typing it as `Proc` isn’t very useful since RBS has “Proc type”s.
* We could make it `_ToProc[^(specific_arg_types) -> specific_return type]` so one can specify the resulting Proc’s contract.
* The current solution is to write a custom `interface` that defines the desired contract: `def to_proc: ^(specific_arg_types) -> specific_return type`.
Regardless of which bullet point to adopt, `_ToProc#to_proc` doesn’t make sense not having a return type – unless  `_ToProc` shouldn’t exist fundamentally.

soutaro (approved):

soutaro:
Thanks @ParadoxV5, 

Having `Proc` is an improvement while it doesn't make much sense, as you said. It makes sense that passing a proc type to the type parameter, but still it's not very useful because we don't have something to express the constraints on the arguments of a proc/method.

PR #1225, opened 2023-02-07
Add link to RBS by Example to README.md

Adam-Stomski:
Most of the people I work with prefer learning based on examples. I think it might be helpful to have the link to "RBS by Example" document in the top level README


soutaro (approved):
Thanks!

PR #1226, opened 2023-02-08
Add `implicitly-returns-nil` annotations

soutaro:
The annotation tells the `Array#[]` and `Hash#[]` may return `nil`.

RBS itself doesn't do anything for this annotation. The type checkers may give precise types of the methods with the annotations.

### DIFF
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -850,7 +850,7 @@ class Array[unchecked out Elem] < Object
   #
   # Array#slice is an alias for Array#[].
   #
-  def []: (int index) -> Elem
+  def []: %a{implicitly-returns-nil} (int index) -> Elem
         | (int start, int length) -> ::Array[Elem]?
         | (::Range[::Integer?] range) -> ::Array[Elem]?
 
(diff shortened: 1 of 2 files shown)


hjwylde:
Hey @soutaro, just had a quick question if you don't mind; which cases are you intending this annotation to be used for? We have a user asking whether `Array#first` should also be annotated with `{implicitly-returns-nil}`. I've answered that I suspect it is an intentional design choice that it's only added to accesses (`[]`), but I'd just like to double check with you to find out for certain.

soutaro:
@hjwylde Hi! Adding the annotation to `Array#first` and likes would make sense. I had planned to implement support for the annotation in Steep to test the type checking experience with it, but not finished yet...

ParadoxV5:
I instead prefer something more explicit… a way for fixed-size tuples and records to narrow the return of their `#[]` and co.… and possibly expand to variable-size Arrays and Hashes.

hjwylde:
Thanks @soutaro, I'll make a pull request to add a few more for the user then :).

I think I do half agree with @ParadoxV5's that there are use-cases for more elaborate contracts (and I've had some ideas on others for RubyMine), but I think `implicitly-returns-nil` is good to start and usable right now.

PR #1229, opened 2023-02-11
Module#attr* returns an Array of Symbol after 3.0

sanfrecce-osaka:
Since below methods returns an Array of Symbol after 3.0, I fixed the return type of them.

- Module#attr_reader
- Module#attr_writer
- Module#attr_accessor

cf. https://bugs.ruby-lang.org/issues/17314
cf. https://github.com/ruby/rbs/pull/1218


soutaro (approved):
🙏

PR #1245, opened 2023-02-20
Add "Expected one argument." to rbs ancestors command

hana-da:
If we run `rbs methods` command with no arguments, It shows "Expected one argument."   
On the other hand, we run `rbs ancestors` command without any arguments, TypeError is raised.

Therfore, I have added "Expected one argument." to `rbs ancestors` command also.


soutaro (approved):
💪

PR #1252, opened 2023-02-24
Add -e, --type, and --method-type options to `rbs parse`

### LINKED ISSUE #925: Add -e, --type, and --method-type options to `rbs parse` command
I propose adding `-e` option and so on to `rbs parse` command.

# Why

Currently we need to put a `.rbs` file in order to run a syntax check. It is not convenient.

```bash
# prepare a temporary file with the content I want to check
$ echo 'class Foo def x: () -> void end' > test.rbs

# Run syntax check
$ rbs parse test.rbs

# Clear the temporary file
$ rm test.rbs
```

So I propose adding the options to `rbs parse` command. We will can avoid putting a file by this change.

```bash
# We can run the syntax check out of the box.
$ rbs parse -e 'class Foo def x: () -> void end'

# It calls `RBS::Parser.parse_type` instead.
# We can focus on the code we want to check by reducing wrapper such as `class Foo .... end`
$ rbs parse --type -e 'Foo::Bar'

# It calls `RBS::Parser.parse_method_type` instead.
# It has the same benefit as `--type` option.
$ rbs parse --method-type -e '() -> void'
```

pocke:
Fix #925

### DIFF
--- a/lib/rbs/cli.rb
+++ b/lib/rbs/cli.rb
@@ -922,6 +922,7 @@ Options:
 
     def run_parse(args, options)
       parse_method = :parse_signature
+      # @type var e_code: String?
       e_code = nil
 
       OptionParser.new do |opts|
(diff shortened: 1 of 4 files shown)


soutaro:
https://github.com/ruby/rbs/pull/1282 would help? @pocke

soutaro (approved):
👏

soutaro (approved):

PR #1255, opened 2023-02-26
Convert `include` in `class << self` to `extend` by `prototype rb`

### LINKED ISSUE #1196: `rbs prototype rb` misunderstand `include` in `class << self `
```Ruby
#test.rb
module A
  def foo = p :foo
end

class B
  class << self
    include A
  end
end
```
```shell
$ rbs -v
rbs 2.8.2

$ rbs prototype rb test.rb

module A
  def foo: () -> untyped
end

class B
  include A
end
```

Maybe `include` declared in `class << self` should be `extend` in rbs's class declaration?
Or it would be nice if rbs could share the same syntax form of `class << self` with ruby
> Please correct me if I miss something

pocke:
Fix #1196

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -215,7 +215,8 @@ module RBS
           when :include
             args.each do |arg|
               if (name = const_to_name(arg, context: context))
-                decls << AST::Members::Include.new(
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍

PR #1273, opened 2023-03-13
Gem.datadir has been removed.

hsbt:
See https://github.com/rubygems/rubygems/pull/1732 and https://github.com/rubygems/rubygems/pull/6469


hsbt:
@soutaro `Gem.datadir` is deprecated since 2016. rbs shouldn't have this signature.

PR #1274, opened 2023-03-13
Skip testing Gem.datadir if not defined

k0kubun:
This was removed in https://github.com/ruby/ruby/commit/8c6b349805e2f17a57576b8dfad31e5681d6b0e9.


k0kubun:
ah, I didn't notice https://github.com/ruby/rbs/pull/1273. Closing.

PR #1279, opened 2023-03-16
Validate duplicate type aliases

pocke:
An environment should reject duplicate type aliases, but actually, it just ignores the duplication due to a forgotten `raise`.

This PR fixes the validation to raise an error on duplicate type aliases, like the following.

```rbs
type foo = String
type foo = Integer
```

## Additional Information

Interface has the same problem. 
https://github.com/ruby/rbs/blob/0dd11d38c11f55ead0a0d78bb9b19d2eaf3a2104/lib/rbs/environment.rb#L364-L371

For example, the second `_I` conceals the first one in the following example.

```rbs
interface _I
  def foo: () -> void
end

interface _I
  def bar: () -> void
end

module M
  include _I
end
```

We can confirm this problem with the following command.

```console
# It should print `foo` and `bar`,
# but actually it prints only `bar`.
$ rbs -I. methods --no-inherit M
bar (public)
```

~~I think the interface should behave the same as the open class, instead of rejecting duplicated interfaces.~~
~~I will create a PR to fix the interface problem soon.~~

### DIFF
--- a/lib/rbs/environment.rb
+++ b/lib/rbs/environment.rb
@@ -374,7 +374,7 @@ module RBS
         name = decl.name.with_prefix(namespace)
 
         if entry = type_alias_decls[name]
-          DuplicatedDeclarationError.new(name, decl, entry.decl)
+          raise DuplicatedDeclarationError.new(name, decl, entry.decl)
         end
 
         type_alias_decls[name] = TypeAliasEntry.new(name: name, decl: decl, outer: outer)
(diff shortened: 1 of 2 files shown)


pocke:
I and @soutaro discussed duplicated interfaces. It should cause an error also.
So I made it causes a duplication error on interfaces too. https://github.com/ruby/rbs/pull/1279/commits/388e6c869c690149f07cf317794b266f4213346c

PR #1284, opened 2023-03-20
Enable Steep in test directory only for completion

pocke:
This PR enables Steep in the test directory only for completion.

# Problem

Currently, Steep does nothing in the test directory.

I think a lack of type-checking is acceptable. Because we need to write many types definitions for tests. Type-checking may be helpful but we need to spend the cost.

On the other hand, we can enable the completion more easily. The completion only works with written type definitions, but it's useful enough.

# Solution

Add a new Steep target for the test directory. This setting does not enable type-checking.

### DIFF
--- a/Steepfile
+++ b/Steepfile
@@ -1,5 +1,19 @@
 D = Steep::Diagnostic
 
+libs = [
+  "set", "pathname", "json", "logger", "monitor",
+  "tsort", "uri", 'dbm', 'pstore', 'singleton',
+  'shellwords', 'fileutils', 'find', 'digest', 'abbrev',
+]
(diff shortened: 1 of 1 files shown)


Steepfile
```diff
@@ -23,6 +35,21 @@ target :lib do
   end
 end
 
+target :test do
+  signature 'sig'
+  check 'test'
+
+  # Ignore the following file because Steep causes SystemStackError on this file.
+  ignore "test/stdlib/enumerator/Product_test.rb"
```
pocke, line 43: I'll report this problem to soutaro/steep soon

PR #1286, opened 2023-03-20
Support class/module alias by RBS::Writer

pocke:
RBS has supported the Class/Module alias since v3, but the writer ignores the Class/Module alias unexpectedly.

This PR adds the support of the Class/Module alias to the writer.


soutaro (approved):
🙏

PR #1287, opened 2023-03-20
Implement `rbs subtract` command

pocke:
This PR adds `rbs subtract` command.

Design: https://hackmd.io/seoMijXwRdG2uFITm2lLqw

I'm trying this feature on my private project, and it works well.

## Limitations

This PR doesn't solve the limitations described in the design doc.
The limitations rarely affect the real applications. So I think we do not need to remove the limitations in this PR.


soutaro (approved):
:+1:

PR #1291, opened 2023-03-25
Avoid to test `Pathname#{taint,untaint}` when it is not defined

pocke:
`Pathname#{taint,untaint}` will be removed in Ruby 3.2, so we do not need to test them on the latest Ruby.

related https://github.com/ruby/rbs/pull/860

-----

By the way, `Pathname#{taint,untaint}` have not been removed yet in Ruby 3.2 in spite of the following warning.

```console
$ ruby -rpathname -e 'Pathname("a").taint'
-e:1: warning: Pathname#taint is deprecated and will be removed in Ruby 3.2.
```

I think we forgot to remove these methods for Ruby 3.2. I'll open an issue to ruby/pathname soon.
ADD: Opened https://github.com/ruby/pathname/issues/28

### DIFF
--- a/test/stdlib/Pathname_test.rb
+++ b/test/stdlib/Pathname_test.rb
@@ -731,8 +731,10 @@ class PathnameInstanceTest < Test::Unit::TestCase
   end
 
   def test_taint
-    assert_send_type '() -> Pathname',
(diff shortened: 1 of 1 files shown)


soutaro (approved):
We may want to delete calling the method simply. 🔪

PR #1299, opened 2023-04-06
Should allow String and _ToStr for `Object#define_singleton_method`

ksss:
define_singleton_method call C function `rb_check_id` for argument. It allows Symbol, String and _ToStr.


soutaro (approved):
👍

PR #1300, opened 2023-04-06
Should align specifications for attr family

ksss:
`attr*` family call C function `rb_check_id`.
These methods are consolidated as `Object::name` in the object.rbs method used, so it is considered appropriate to specify the same type if we want to conform to the same specification.

Another approach would be to prepare a `type Module::name` definition that is exactly the same as `Object::name`, and use Module::name as the type for Module methods.

What do you think?


soutaro (approved):
Thanks! 👍

PR #1302, opened 2023-04-09
Correct docs for Comparable

waiwaing:


pocke:
This documentation is synced with the latest released Ruby by `rake annotate`. So we do not need to update this document for now.

Thanks for your PR but I'll close it.

PR #1308, opened 2023-04-17
Run `bundle install` automatically in bin/steep

pocke:
This PR adds `bundle install` into bin/steep to ensure install gems for Steep.

# Problem

Currently, we need to run `bundle install` manually to run Steep. Because steep uses a different Gemfile.

But IDEs, such as VSCode, execute `bin/steep` transparently. Sometimes I overlooked the missing gems error because VSCode conceals the error (maybe VSCode displays the error as a small popup, but I never check the popup contents 😇).

# Solution

Run `bundle install` in `bin/steep`. By this change, `bin/steep` doesn't fail by the missing gems error.

I've applied the same solution to our company project, and it works well.


soutaro (approved):
👍

PR #1316, opened 2023-04-24
Fix symbol and string literal types' `to_s`

pocke:


lib/rbs/types.rb
```diff
@@ -1212,7 +1212,57 @@ def to_json(state = _ = nil)
...
+        else
+          literal.inspect
+        end
+      end
+
+      private
+
+      def string_to_s(str)
+        content = str.chars.map do |c|
+          case c
```
sampersand, line 1229: Wouldnt a hash and `[]` work better? like `str.chars.map { |c| SPECIAL_CHARS.fetch(c, c) }` ?

PR #1321, opened 2023-04-28
Allow Symbol | string to Module methods

ksss:
CRuby's C APIs, `rb_check_id` and `rb_to_id`, allow Symbols and strings. The following methods that use this C API internally should allow these types.

- Module#attr
- Module#attr_accessor
- Module#attr_reader
- Module#attr_writer
- Module#autoload
- Module#autoload?
- Module#alias_method
- Module#class_variable_defined?
- Module#class_variable_get
- Module#class_variable_set
- Module#const_defined?
- Module#const_get
- Module#const_set
- Module#const_source_location
- Module#deprecate_constant
- Module#define_method
- Module#instance_method
- Module#method_defined?
- Module#module_function
- Module#public_instance_method
- Module#public_method_defined?
- Module#public_class_method
- Module#public_constant
- Module#private_method_defined?
- Module#private_class_method
- Module#private_constant
- Module#protected_method_defined?
- Module#protected_class_method
- Module#remove_const
- Module#remove_class_variable
- Module#remove_method
- Module#undef_method

### DIFF
--- a/core/module.rbs
+++ b/core/module.rbs
@@ -235,7 +235,7 @@ class Module < Object
   #
   #     Exiting with code 99
   #
-  def alias_method: (::Symbol | ::String new_name, ::Symbol | ::String old_name) -> ::Symbol
+  def alias_method: (id new_name, id old_name) -> ::Symbol
 
   # <!--
   #   rdoc-file=object.c
@@ -284,7 +284,7 @@ class Module < Object
   #     end
   #     Mod.instance_methods.sort   #=> [:one, :one=, :two, :two=]
   #
-  def attr_accessor: (*Object::name arg0) -> Array[Symbol]
+  def attr_accessor: (*id arg0) -> Array[Symbol]
 
   # <!--
   #   rdoc-file=object.c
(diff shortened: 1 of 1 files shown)


core/module.rbs
```diff
@@ -1629,5 +1629,7 @@ class Module < Object
   # `attr_reader(name)` but deprecated. Returns an array of defined method names
   # as symbols.
   #
-  def attr: (*Object::name arg0) -> Array[Symbol]
+  def attr: (*id arg0) -> Array[Symbol]
+
+  type id = Symbol | string
```
ParadoxV5, line 1634: `Object::name`?

ksss: Yes, It's very similar to `Object::name`.
However, I am questioning the correctness of `Object::name` name.
The expression is ambiguous with `name`.
The context here is `Module`, and `id` is used inside CRuby, as in `rb_check_id`.

soutaro (approved):
👍

core/module.rbs
ParadoxV5: Should those using `Object::name` switch to `Module::id` instead?

PR #1329, opened 2023-05-13
Print detailed_message on parse command

ksss:
Extends the error display when parsing fails with the `parse` command.
This improves the user experience by making it easier to identify which part was at fault.


soutaro (approved):
👍

PR #1337, opened 2023-05-20
Allow _ToInt to argument for Time#{round,floor,ceil}

ksss:
```rb
o = Object.new; def o.to_int = 1; Time.now.floor(o)
#=> 2023-05-20 14:18:40.1 +0900
```


soutaro (approved):
🙏

PR #1342, opened 2023-05-25
Fix Kernel.p and Kernel.pp type with no argument

tompng:
`Kernel.p(T)` returns `T`, `Kernel.p(*more_than_two)` returns array, `Kernel.p()` retuns nil.
I added the missing definition `() -> nil`.

I think `(*untyped arg0) -> Array[untyped]` conflicts with `(T arg0) -> T` and `() -> nil`, so I changed it to `(untyped, untyped, *untyped) -> Array[untyped]`

```ruby
# I referenced core/module.rbs
class Module < Object
  def public: () -> nil
            | (Symbol method_name) -> Symbol
            | (Symbol, Symbol, *Symbol method_name) -> Array[Symbol]
            | ...
end
```

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -1241,7 +1241,8 @@ module Kernel : BasicObject
   #     [0..4, 0..4, 0..4]
   #
   def self?.p: [T] (T arg0) -> T
-             | (*untyped arg0) -> Array[untyped]
+             | (untyped, untyped, *untyped) -> Array[untyped]
+             | () -> nil
 
   # <!--
   #   rdoc-file=lib/pp.rb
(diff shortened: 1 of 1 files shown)


core/kernel.rbs
```diff
@@ -1241,7 +1241,9 @@ module Kernel : BasicObject
   #     [0..4, 0..4, 0..4]
   #
   def self?.p: [T] (T arg0) -> T
-             | (*untyped arg0) -> Array[untyped]
+             | (untyped, untyped, *untyped) -> Array[untyped]
+             | () -> nil
+
```
ksss, line 1246: It appears that this line is the difference and the test is failing.

tompng: Thanks, I remove it

PR #1343, opened 2023-05-25
prototype rb: Extract instance variables and class variables

tk0miya:


tk0miya:
This patch generates the definition of instance variables and class variables just after the method definition:

```
# source.rb
class Foo
  def meth
    @_meth = 1
  end

  private

  def private_meth
    @_private_meth = 2
  end
end
```
```
# generated.rbs
class Foo
  def meth: () -> untyped

  @_meth: untyped
end
```

It seems fine.

But in another case, it's a bit strange for me if variables are defined inside a private method.

```
# source.rb
class Foo
  def meth
    @_meth = 1
  end

  private

  def private_meth
    @_private_meth = 2
  end
end
```
```
# generated.rbs
class Foo
  def meth: () -> untyped

  @_meth: untyped

  private

  def private_meth: () -> untyped

  @_private_meth: untyped
end
```

Is this acceptable? Or, do we need to reorder the definition before generation?
Please let me know your opinion.

Thanks,

tk0miya:
>Is this acceptable? Or, do we need to reorder the definition before generation?
>Please let me know your opinion.

Note: After this post, I and @soutaro talked about this topic in the ruby-jp slack. And we determined to move these variable definitions to the top of the class. So I'll update this PR this weekend.

tk0miya:
I just update my PR to sort the member of class definitions. Now instance variables and class variables will be generated at the top of each class.

pocke:
Would you be able to support class instance variables?

```ruby
class C
  def self.foo
    @foo = 42
  end

  class << self
    def bar
      @bar = 42
    end
  end
end
```

In this case, `@foo` and `@bar` should be class instance variables, but `rbs prototype rb` outputs them as instance variables.

tk0miya:
It seems CI for the nightly image failed because installing the stringio gem failed.

lib/rbs/prototype/rb.rb
```diff
@@ -195,6 +196,11 @@ def process(node, decls:, comments:, context:)
 
           decls.push member unless decls.include?(member)
 
+          new_ctx = Context.initial.tap { |ctx| ctx.singleton = kind == :singleton }
```
pocke, line 199:
```suggestion
          new_ctx = context.dup.tap { |ctx| ctx.singleton = kind == :singleton }
```

Because the new context should inherit the namespace context.
Currently it does not use the namespace context inside the method definition. But the new context should keep the parent context's namespace for the future. 


---


BTW, I'd like to refactor the context class to implement more concrete methods to update the context. Currently it just provides primitive interfaces and it introduces confusion.

lib/rbs/prototype/rb.rb
```diff
@@ -377,6 +383,32 @@ def process(node, decls:, comments:, context:)
             comment: comments[node.first_lineno - 1]
           )
 
+        when :IASGN
+          if context.singleton || !context.namespace.empty?
+            member = AST::Members::ClassVariable.new(
```
pocke, line 388: It should be `ClassInstanceVariable` instead.

tk0miya:
Updated. But, after the update, I don't have a way to promote class instance variables placed just inside the class definition:

```
class C
 @var = nil
end
```

Before the update, I used the context not having a namespace...

pocke (approved):
Thanks!

pocke:
> Updated. But, after the update, I don't have a way to promote class instance variables placed just inside the class definition:


I understand. I'll try fixing this problem on another PR.

Issue #670, opened 2021-04-21
What is a full name of the rbs?


junaruga:
Is the rbs an abbreviation of something right? What is the full name of the rbs?
Is the full name "RuBy Structure" (the `README.md` says "RBS is a language to describe the structure of Ruby programs") or "RuBy type Signature" (the `rbs.gemspec` summary says "Type signature for Ruby.") ?

For example I assume a full name of the irb is "Interactive RuBy".

junaruga:
When I wanted to create a document to explain the rbs, I had the question.

soutaro:
I don't think RBS is an acronym of a _full name_. It implies a combination of RuBy and Signature, but it's not the full name.

I guess it's similar to RPM (is not Redhat Package Manager), PHP (is not Personal Homepage Page tools), or OCaml (is not Objective Caml).

junaruga:
All right. Thanks for the answer. I would close this ticket.

As a reference, in case of RPM, I think it was an acronym of "Red hat Package Manager", now  it is an acronym of "RPM Package Manager".

https://en.wikipedia.org/wiki/RPM_Package_Manager
> RPM Package Manager (RPM) (originally Red Hat Package Manager, now a recursive acronym)

PR #1350, opened 2023-06-10
Add description of the abbreviated term "RBS"

piroor:
I was not familiar to the context of this project so I've been confused what means the project name at the first time. How about adding description about the name "RBS" itself?


pocke:
Thanks for your PR!
Right, I think you are still confused because RBS is not an acronym for "RuBy Signature".
ref: https://github.com/ruby/rbs/issues/670


I think your proposal will introduce another confusion. How about improving the description with the linked issue?

piroor:
Oh, sorry I missed the thread. I've updated the description (added "originated from") and linked to the issue. How about this?

piroor:
And I agree it is better way that adding a FAQ topic like some major projects instead adding to the first description, if there was FAQ list.

README.md
```diff
@@ -1,6 +1,6 @@
 # RBS
 
-RBS is a language to describe the structure of Ruby programs.
+RBS (*R*u*B*y *S*ignature) is a language to describe the structure of Ruby programs.
```
ParadoxV5: How about:
```suggestion
RBS (“`.rb` *s*ignature”) is a language to describe the structure of Ruby programs.
```

PR #1353, opened 2023-06-13
Fix broken CI

tk0miya:
Our CI has been broken because of changes in Ruby 3.3.

This fixes them and gets green again.

* dep: Update stringio to 3.0.7 to support Ruby 3.3

  stringio-3.0.7 was released at 2023/6/2 to support Ruby 3.3.
  https://github.com/ruby/stringio/blob/master/NEWS.md#307---2023-06-02

  This upgrade it from 3.0.6 to 3.0.7 to resolve errors for
  master-nightly-* tests.

* Workaround for Parser 3.2.2.2 or lower with Ruby 3.3.0dev

  Follow up ruby/ruby#7877.

  Ruby 3.3.0dev is set to promote Racc to a bundled gem. Therefore,
  this PR provides a workaround for Parser 3.2.2.2 or lower with Ruby
  3.3.0dev to prevents the Ruby 3.3.0dev CI matrix error from rubocop.

  refs: https://github.com/rubocop/rubocop/pull/11941

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,11 @@ gem 'tempfile'
 gem "prime"
 gem "rdoc", "~> 6.4.0"
 
+# FIXME: Workaround for Parser 3.2.2.2 or lower with Ruby 3.3.0dev.
+# When the Praser gem releases a new version of Racc that includes the runtime dependencies,
+# it will be able to upgrade the Parser gem dependency and remove the workaround.
+gem 'racc', '>= 1.6.2'
+
 # Test gems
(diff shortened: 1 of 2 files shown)


tk0miya:
Oops. The new racc and rubocop package were released last week. So it's better to upgrade rubocop instead of pinning racc.
https://github.com/rubocop/rubocop/pull/11942

koic:
Exactly. Racc is installed as a dependency via Parser gem if RuboCop 1.52.1+ is used.

PR #1355, opened 2023-06-15
Removed Prime signature and tests

hsbt:
Extracted from https://github.com/ruby/rbs/pull/1354.

We shouldn't test Prime because it's the bundled gems, not stdlib.


soutaro:
Thanks. We will move the type definition to gem_rbs_collection.

hsbt:
Thanks!

>We will move the type definition to gem_rbs_collection.


I'll do that.

PR #1356, opened 2023-06-16
Handle connection error with Resolv_test.rb

hsbt:
see https://github.com/ruby/rbs/actions/runs/5285494174/jobs/9564033234#step:10:26

We sometimes faced "connection refused" issue in our CI containded `ruby/ruby` repo. I added rescue block with `Errno::ECONNREFUSED` for this.


pocke (approved):
Thanks!

PR #1365, opened 2023-07-06
Added setup instructions for testing the core/stdlib types.

yalab:
Hi.

I tried to set up to see if I could contribute, but I encountered some issues along the way. 
I have added them to the documentation for reference.

thanks.


soutaro:
In fact, you can just run `bin/setup` instead, which is explained in [README.md](https://github.com/ruby/rbs#development:~:text=After%20checking%20out%20the%20repo%2C%20run%20bin/setup%20to%20install%20dependencies.). You may want to explain it in `stdlib.md` too.

PR #1368, opened 2023-07-10
Improve type signature for Kernel.exec

camertron:
According to the official Ruby documentation, `Kernel.exec` should have the same type signature as `Kernel.spawn` except for the return value.


soutaro (approved):
Thanks!

PR #1371, opened 2023-07-12
Add URI Errors type definition 

S-H-GAMELINKS:
I'm using RBS at work now, and type check fails because there is no type definition for URI errors. 
So, I added it.

I am new to contributing to RBS, so if you need anything additional, I would appreciate your comments!


soutaro (approved):
Thanks! 👍

PR #1373, opened 2023-07-13
Make TypeNameResolver more compatible with Ruby

soutaro:


test/rbs/resolver/type_name_resolver_test.rb
```diff
@@ -42,6 +42,26 @@ class Bar
...
+  end
+
+  class Bar
+  end
+end
+EOF
+      manager.build do |env|
+        resolver = Resolver::TypeNameResolver.new(env)
+
+        assert_nil resolver.resolve(type_name("Foo::Bar"), context: [[nil, TypeName("::Foo")], TypeName("::Foo::Foo")])
```
soutaro, line 59: It tried `::Foo::Foo::Foo::Bar`, `::Foo::Foo::Bar`, and `::Foo::Bar` and resolved to `::Foo::Bar`, but it was not how Ruby resolves constants.

With this PR, it tries only with the first component of given type name `Foo` as `::Foo::Foo::Foo` and `::Foo::Foo`, and resolves to `::Foo::Foo`. Then it adds the tail part `Bar` as `::Foo::Foo::Bar`, but it doesn't exist. So, it returns `nil`.

PR #1384, opened 2023-07-21
subtract: Filter redundant access modifiers

tk0miya:
Filter access modifiers (public and private) from subtracted RBS code.

* repeated access modifiers
* access modifiers not having any members


soutaro (approved):
👍

soutaro:
@pocke Do you have any comments? or it will be merged in a few days.

PR #1386, opened 2023-07-26
Fix type name for prototype rbi

ksss:
This PR solves two problems.

## Fix untyped in case of COLON3

```rbi
class Test
  sig { returns(Foo) }
  def m1; end

  sig { returns(::Foo) }
  def m2; end
end
```

```rbs
class Test
  def m1: () -> Foo

  # before
  def m2: () -> untyped

  # after
  def m2: () -> ::Foo
end
```

## Fixed a problem where namespace disappears when an absolute path is specified

```rbi
class Test
  sig { returns(Foo::Bar) }
  def m3; end

  sig { returns(::Foo::Bar) }
  def m4; end
end
```

```rbs
class Test
  def m3: () -> Foo::Bar

  # before
  def m4: () -> ::Bar

  # after
  def m4: () -> ::Foo::Bar
end
```

### DIFF
--- a/lib/rbs/prototype/rbi.rb
+++ b/lib/rbs/prototype/rbi.rb
@@ -479,7 +479,7 @@ module RBS
           else
             Types::ClassInstance.new(name: const_to_name(type_node), args: [], location: nil)
           end
-        when type_node.type == :COLON2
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍

PR #1396, opened 2023-07-27
Make the arg for `Kernel::exit!` optional

### LINKED ISSUE #1298: Kernel exit! has an optional argument
Investigating a seemingly spurious `Ruby::InsufficientPositionalArguments` from steep, I think I find an incorrect signature for `exit!`, in the [kernel source](https://github.com/ruby/ruby/blob/v3_2_0/process.c#L4493):
```C
static VALUE
rb_f_exit_bang(int argc, VALUE *argv, VALUE obj)
{
    int istatus;

    if (rb_check_arity(argc, 0, 1) == 1) {
        istatus = exit_status_code(argv[0]);
    }
    else {
        istatus = EXIT_FAILURE;
    }
    _exit(istatus);

    UNREACHABLE_RETURN(Qnil);
}
```
but the [signature](https://github.com/ruby/rbs/blob/master/core/kernel.rbs#L810) requires an argument
```ruby
  def self?.exit!: (Integer | TrueClass | FalseClass status) -> bot
```

ParadoxV5:
Fixes #1298
The associated test is blank because – you can’t test this…

#good-first-issue

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -807,7 +807,7 @@ module Kernel : BasicObject
   #
   #     Process.exit!(true)
   #
-  def self?.exit!: (Integer | TrueClass | FalseClass status) -> bot
+  def self?.exit!: (?Integer | TrueClass | FalseClass status) -> bot
 
   # <!-- rdoc-file=eval.c -->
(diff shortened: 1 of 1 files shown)


soutaro (approved):
👍

PR #1397, opened 2023-07-27
Allow `Float` & `nil` for `ConditionVariable#wait` 2nd arg

### LINKED ISSUE #1306: ConditionVariable#wait signature is too narrow
[Signature is currently](https://github.com/ruby/rbs/blob/5d60b1d3b887a22876191e78c06a3b5536a6b6b4/core/thread.rbs#L1417):

```
  def wait: (Thread::Mutex mutex, ?Integer timeout) -> self
```

But it accepts `nil` (which is the default value) and, tracking it down to [`time_timespec(VALUE num, int interval)`](https://github.com/ruby/ruby/blob/6c1b604e29c32baff612d8e26947255a5f3c70ac/time.c#L2770), also `Float`.

So it looks like the signature should actually be:

```
  def wait: (Thread::Mutex mutex, ?(Integer || Float)? timeout) -> self
```

ParadoxV5:
Fixes #1306

#good-first-issue

### DIFF
--- a/core/thread.rbs
+++ b/core/thread.rbs
@@ -1414,7 +1414,7 @@ class Thread::ConditionVariable < Object
   #
   # Returns the slept result on `mutex`.
   #
-  def wait: (Thread::Mutex mutex, ?Integer timeout) -> self
+  def wait: (Thread::Mutex mutex, ?Integer | Float? timeout) -> self
 end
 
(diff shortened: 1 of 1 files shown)


ParadoxV5:
P.S. there seems to be no associated test

soutaro (approved):
Thanks! 👍

PR #1398, opened 2023-07-27
Syntax highlight `README.md` example as `rbs`

ParadoxV5:
It’s still highlighted as Ruby (`rb`) currently (while the rest have switched to `rbs` at b829cc4)

(One-char PR lol)


ParadoxV5:
On that note, I noticed that some code blocks highlight as `rb` while others `ruby`.

pocke (approved):
Thanks!

PR #1399, opened 2023-07-28
Update `io.rbs`

ParadoxV5:
* Fill in `untyped` as appropriate
  * Many of them are simply `boolish` or `void`
  * Patching in `::popen` is my main motivation. Yes, it is ridiculous.
  * Copy-pasted kwargs for `#initialize` and `::open`
* Use `nil` rather than `NilClass`
* Delete nonexistent methods `#bytes`, `#chars`, `#codepoints` and `#lines`
* Actually make aliases `alias`es
* Fill in arg names for `arg0`

### DIFF
--- a/core/io.rbs
+++ b/core/io.rbs
@@ -1202,7 +1202,7 @@ class IO < Object
   #     IO.new(fd, internal_encoding: nil) # => #<IO:fd 3>
   #     IO.new(fd, autoclose: true)        # => #<IO:fd 3>
   #
-  def initialize: (int fd, ?string | int mode, ?path: string?, ?external_encoding: String external_encoding, ?internal_encoding: String internal_encoding, ?encoding: String encoding, ?textmode: boolish textmode, ?binmode: boolish binmode, ?autoclose: boolish autoclose, ?mode: String mode) -> void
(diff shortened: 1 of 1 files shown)


core/io.rbs
```diff
@@ -1309,7 +1202,7 @@ class IO < Object
   #     IO.new(fd, internal_encoding: nil) # => #<IO:fd 3>
   #     IO.new(fd, autoclose: true)        # => #<IO:fd 3>
   #
-  def initialize: (int fd, ?string | int mode, ?path: string?, **untyped opts) -> void
+  def initialize: (int fd, ?string | int mode, ?path: string?, ?external_encoding: String external_encoding, ?internal_encoding: String internal_encoding, ?encoding: String encoding, ?textmode: boolish textmode, ?binmode: boolish binmode, ?autoclose: boolish autoclose, ?mode: String mode) -> void
```
soutaro, line 1205:
```suggestion
  def initialize: (int fd, ?string | int mode, ?path: string?, ?external_encoding: String | Encoding | nil, ?internal_encoding: String | Encoding | nil, ?encoding: String | Encoding | nil, ?textmode: boolish, ?binmode: boolish, ?autoclose: boolish, ?mode: String) -> void
```

soutaro (approved):
Thanks! 👍

PR #1400, opened 2023-07-28
`rake annotate`

ParadoxV5:
The rest of `rake annotate` changes from a184eda & dc59809


ParadoxV5:
The files were generated with Ruby 3.2.2.

soutaro:
@ParadoxV5 Can you explain where the diffs come from? (It looks like CI generates the old docs.)

ParadoxV5:
I’ve identified the deviation. When I ran `bundle exec rake annotate`, the script fails at the line
`bundle exec rbs annotate --no-system --no-gems --no-site --no-home -d /tmp/rdoc-3.2.2 core stdlib`:
```
bundler: failed to load command: rbs (/usr/local/lib/ruby/gems/3.2.0/bin/rbs)
<internal:/usr/local/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require': cannot load such file -- rbs_extension (LoadError)
```

Thus in order to publish the PRs associated with the aforementioned commits, I ran
`rbs annotate --no-system --no-gems --no-site --no-home -d /tmp/rdoc-3.2.2 core stdlib`
without `bundle exec`. The more up-to-date gems in my system may be the cause of these extra diffs.

Since you have prompted a further investigation into these, I have retraced the steps. I noticed that replacing `rbs` with `exe/rbs` does not make a difference. That is, the `LoadError` still raises with a manually run of
`bundle exec exe/rbs annotate --no-system --no-gems --no-site --no-home -d /tmp/rdoc-3.2.2 core stdlib`,
and the diffs still generate with
`exe/rbs annotate --no-system --no-gems --no-site --no-home -d /tmp/rdoc-3.2.2 core stdlib`.

ParadoxV5:
Checking the `bundle install` logs, I noticed that every fetch/install line is for older versions of dependent gems. What’s the dep update guideline for this project?

soutaro:
Thanks for the investigation. It seems because of the version difference between your local gems and locked gems.

I recently found the `rbs annotate` code depending on RDoc has some potential problems. And plan to fix them in next version.

ParadoxV5:
Obsolete, except maybe [#issuecomment-1683366042](#issuecomment-1683366042)

PR #1415, opened 2023-08-08
Super class should be absolute paths.

ksss:
## Problem

The signature generated by the `rbs prototype runtime` for the `ActiveSupport::Logger` as follows is unreadable due to an error.
`Detected recursive ancestors: ::ActiveSupport::Logger < ::ActiveSupport::Logger (RBS::RecursiveAncestorError)`

```rbs
$ bundle exec rbs prototype runtime -r active_support/all ActiveSupport::Logger
module ActiveSupport
  class Logger < Logger
...snip...
```

In the actual Rails code, super class are specified by absolute paths, as shown below.

https://github.com/rails/rails/blob/2942958827f1934dfcba284d074e6d61104d3e7c/activesupport/lib/active_support/logger.rb#L7-L8

## Proposal

I propose to solve this problem by specifying the superclass as an absolute path even in the `rbs prototype runtime`.
There may be other similar problems, but I have limited the impact to specifying the superclass only.

```rbs
$ bundle exec rbs prototype runtime -r active_support/all ActiveSupport::Logger
module ActiveSupport
  class Logger < ::Logger
...snip...
```

### DIFF
--- a/lib/rbs/prototype/runtime.rb
+++ b/lib/rbs/prototype/runtime.rb
@@ -380,7 +380,7 @@ module RBS
           RBS.logger.warn("Skipping anonymous superclass #{mod.superclass} of #{mod}")
           nil
         else
-          super_name = to_type_name(const_name(mod.superclass), full_name: true)
+          super_name = to_type_name(const_name(mod.superclass), full_name: true).absolute!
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍

PR #1419, opened 2023-08-09
prototype rb: Fix a crash by "include foo"

### LINKED ISSUE #1416: `rbs prototype rb` fails when include target contains variables
## Steps to reproduce

It works with v3.1.3, but an error occurs with rbs on the master branch.

### code

```console
$ cat test.rb
module Devise
  def self.include_helpers(scope)
    ActiveSupport.on_load(:action_controller) do
      include scope::Helpers if defined?(scope::Helpers)
      include scope::UrlHelpers
    end
  end
end
```

### v3.1.3

```console
$ rbs prototype rb test.rb
module Devise
  def self.include_helpers: (untyped scope) -> untyped
end
```

### master 90ea1493d1417e7bb14840056c8d4e9d6a0055ef

```console
$ ./exe/rbs prototype rb test.rb
/Users/sinsoku/ghq/github.com/ruby/rbs/lib/rbs/prototype/rb.rb:454:in `const_to_name!': unhandled exception
	from /Users/sinsoku/ghq/github.com/ruby/rbs/lib/rbs/prototype/rb.rb:445:in `const_to_name!'
```

I found commit 69f84f88c68220279c5317214326e41cfa1806c1 with `git bisect`.

## System configuration

```console
$ ruby --version
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin22]
```

tk0miya:
RBS::Prototype::RB#const_to_name() raises an exeception if non-constant reference given.  So it will cause a crash when the parser tries to parse the non-constant argument for the include statement.

Close #1416

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -461,7 +461,7 @@ module RBS
           when :SELF
             context.namespace.to_type_name
           when :CONST, :COLON2, :COLON3
-            const_to_name!(node)
+            const_to_name!(node) rescue nil
           end
         end
       end
--- a/test/rbs/rb_prototype_test.rb
+++ b/test/rbs/rb_prototype_test.rb
@@ -1035,6 +1035,21 @@ end
     end
   end
 
+  def test_const_to_name
+    parser = RBS::Prototype::RB.new
+    [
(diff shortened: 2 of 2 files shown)


soutaro (approved):
👍

PR #1421, opened 2023-08-10
Avoid to generate anonymous or internal module

ksss:
Anonymous or internal modules cannot have their names defined as RBS, so they cannot be used even if they get a name. These should be treated as if the name could not be obtained.

## Before

```rbs
$ rbs prototype runtime "NameError*"
class NameError < StandardError
  include DidYouMean::Correctable

  include ErrorHighlight::CoreExt

  public

  def local_variables: () -> untyped

  def name: () -> untyped

  def receiver: () -> untyped

  private

  def initialize: (*untyped) -> void

  class message   #=> Syntax error: expected one of class/module/constant name, token=`message` (tLIDENT)
    def self._load: (untyped) -> untyped

    public

    def ==: (untyped) -> untyped

    def _dump: (untyped) -> untyped

    def to_str: () -> untyped

    private

    def initialize_copy: (untyped) -> untyped
  end
end
```

## After

```rbs
$ bundle exec rbs prototype runtime "NameError*"
class NameError < StandardError
  include DidYouMean::Correctable

  include ErrorHighlight::CoreExt

  public

  def local_variables: () -> untyped

  def name: () -> untyped

  def receiver: () -> untyped

  private

  def initialize: (*untyped) -> void
end
```

### DIFF
--- a/lib/rbs/prototype/runtime.rb
+++ b/lib/rbs/prototype/runtime.rb
@@ -541,7 +541,17 @@ module RBS
 
       def const_name(const)
         @module_name_method ||= Module.instance_method(:name)
-        @module_name_method.bind(const).call
+        name = @module_name_method.bind(const).call
+        return nil unless name
+
+        begin
+          Object.const_get(name)
+        rescue NameError
+          # Should generate const name if anonymous or internal module (e.g. NameError::message)
+          nil
+        else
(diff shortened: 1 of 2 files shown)


soutaro (approved):
🙏

PR #1430, opened 2023-08-18
Convert `| nil` to `?` in global_variables.rbs

sampersand:
Nearly all of the old-style `| nil` annotations have been replaced with `?`. Only `$-F` and `$;` have been left as-is, as they're `Regex | String | nil`.


soutaro (approved):
💅 🎉

PR #1432, opened 2023-08-18
Remove TRUE, FALSE, and NIL constants

sampersand:
These constants were removed in Ruby 3.0; Ruby 2.7 is now 4 months EOL, and nobody was using these constants in 2.7 anyways.


soutaro (approved):
🙏

PR #1433, opened 2023-08-18
Corrected Kernel#Array's signature

sampersand:
This updates `Kernel#Array` to the correct signature.

The `Array()` function works as follows:
1. If the argument is an array, it's returned immediately
2. If the argument defines `to_ary`, then the return value of that is returned
3. If the argument defines `to_a`, then the return value of that is returned.
4. Otherwise, the argument is boxed up in an array and returned.

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -403,13 +403,11 @@ module Kernel : BasicObject
   #
   #     Array(:foo)             # => [:foo]
   #
-  def self?.Array: (NilClass x) -> [ ]
-                 | [T] (::Array[T] x) -> ::Array[T]
-                 | [T] (::Range[T] x) -> ::Array[T]
-                 | [T] (_Each[T] x) -> ::Array[T]
-                 | [K, V] (::Hash[K, V] x) -> ::Array[[ K, V ]]
-                 | [T] (T x) -> ::Array[T]
-
+  def self?.Array: (nil) -> [ ]
+                 | [T] (Array[T] x) -> Array[T]
+                 | [T] (_ToAry[T] x) -> Array[T]
+                 | [T] (_ToA[T] x) -> Array[T]
+                 | [T] (T x) -> Array[T]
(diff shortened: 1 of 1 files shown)


core/kernel.rbs
```diff
@@ -403,13 +403,11 @@ module Kernel : BasicObject
...
-  def self?.Array: (NilClass x) -> [ ]
-                 | [T] (::Array[T] x) -> ::Array[T]
-                 | [T] (::Range[T] x) -> ::Array[T]
-                 | [T] (_Each[T] x) -> ::Array[T]
-                 | [K, V] (::Hash[K, V] x) -> ::Array[[ K, V ]]
-                 | [T] (T x) -> ::Array[T]
-
+  def self?.Array: (nil) -> [ ]
+                 | [T] (Array[T] x) -> Array[T]
+                 | [T] (_ToAry[T] x) -> Array[T]
```
ParadoxV5, line 408:
```suggestion
  def self?.Array: [T] (_ToAry[T] x) -> Array[T]
```

`nil` is already a `_ToA` (a nit, `NilClass#to_a` RBS *also* gives `[ ]` rather than `[]`),
and Array is already a `_ToAry` and a `_ToA`

Though we can keep them for readability.
* [x] `Array(an_array)`… programmers probably know what they are doïng.
* [ ] `Array(nil)` … will rubyists think this give `[nil]` instead?
  * [x] `Array(nilable)`… I guess this one doesn’t matter that much since there’s `Array(array_or_element_or_nil)`

sampersand: Yeah, I was keeping it for readability. I don't think people will think `Array nil #=> [nil]` personally

ParadoxV5: If the lack of clarifications does not confuse people then it’s fine to omit them. WDYT, they who merges PRs?

sampersand:

PR #1434, opened 2023-08-18
Corrected Kernel#Complex's signature

sampersand:
This updates `Kernel#Complex`'s signature.

In particular, if the first argument defines `.to_c`, and no second argument is supplied, it calls and returns that. (In the case where `exception: false` is given, _all_ exceptions (even `Exception` itself) that would be thrown by `.to_c` are swallowed, and `nil` is returned.)

Two additional variants, for when `exception: false` is explicitly given, were added. In these cases, the return value is `Complex?`.

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -446,7 +446,11 @@ module Kernel : BasicObject
   #
   # See String#to_c.
   #
-  def self?.Complex: (Numeric | String x, ?Numeric | String y, ?exception: bool exception) -> Complex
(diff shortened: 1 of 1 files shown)


sampersand:
Technically, there could also be an `| (untyped, ?untyped, exception: false) -> nil` variant, but that seems a bit awkward.

sampersand:
I've added in the `| (untyped, ?untyped, exception: false) -> nil` variant for completeness

sampersand:

PR #1436, opened 2023-08-18
Updated Kernel#Hash

sampersand:
This updates `Kernel#Hash`'s signature.

This is probably the second-simplest of the kernel conversion methods, after `Kernel#String`: 
- If the argument is `nil` or an empty array, an empty hash is returned.
- Otherwise, `.to_hash` is called on the object.

I was debating on whether I should make a separate `(nil | []) -> {}` case, but I decided against it. (Using `[]` / `{}` literals in practice is pretty difficult, as it's a `Hash[bot, bot]`.)

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -488,7 +488,7 @@ module Kernel : BasicObject
   #     Hash(nil)              # => {}
   #     Hash([])               # => {}
   #
-  def self?.Hash: [K, V] (Object x) -> ::Hash[K, V]
+  def self?.Hash: [K, V] (nil | [] | _ToHash[K, V])  -> Hash[K, V]
 
   # <!--
   #   rdoc-file=object.c


core/kernel.rbs
```diff
@@ -488,7 +488,7 @@ module Kernel : BasicObject
   #     Hash(nil)              # => {}
   #     Hash([])               # => {}
   #
-  def self?.Hash: [K, V] (Object x) -> ::Hash[K, V]
+  def self?.Hash: [K, V] (nil | [] | _ToHash[K, V])  -> Hash[K, V]
```
ParadoxV5, line 491: > I was debating on whether I should make a separate `(nil | []) -> {}` case, but I decided against it.

No need to, at all, since `nil` and empty arrays already have `#to_h`, let their own RBSs do the `[K, V]` magic. (I’ll check RBS for `NilClass#to_h`)

```suggestion
  def self?.Hash: [K, V] (_ToHash[K, V])  -> Hash[K, V]
```

ParadoxV5: `NilClass#to_h` gives `Hash[untyped, untyped]` :+1:

sampersand: It's not the exact same @ParadoxV5—`_ToHash` requires `to_hash`, not `to_h`. So the `nil | []` case is distinct.

sampersand:

PR #1438, opened 2023-08-18
Updated Kernel#Rational

sampersand:
Updated `Kernel#Rational` to be more accurate.

Phew, this one is a mess. In "normal operation," the signature would look like this:

```rbs
type num = Complex|Float|int|_ToR|String
def self?.Rational: (num numer, num denom, exception: false) -> Rational?
                  | (num numer, num denom, ?exception: bool) -> Rational
```

However, `Kernel#Rational` _also_ tries to be a "division operator" for some reason: If the first argument is `Numeric`, but defines neither `to_int` nor `to_i` nor even `to_r`, `Kernel#Rational` will return `arg1 / arg2`. That is, it'll actually call the `/` operator on the first argument, with the second.

But it gets worse. If the second argument is exactly `1`, then `/` won't even be called. 

In neither case does `exception: false` actually work—unlike the other conversion methods, exceptions are _not_ caught when `Kernel#Rational` is trying to act like a division operator.

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -622,7 +622,6 @@ module Kernel : BasicObject
   interface _RationalDiv[T]
     def /: (Numeric) -> T
   end
-  #: (Numeric | String | Object x, ?Numeric | String y, ?exception: bool exception) -> Rational
 
   # <!--
   #   rdoc-file=object.c
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -614,7 +614,15 @@ module Kernel : BasicObject
   #
   # See also String#to_r.
   #
-  def self?.Rational: (Numeric | String | Object x, ?Numeric | String y, ?exception: bool exception) -> Rational
+  def self?.Rational: (_ToInt | _ToR numer, ?_ToInt | _ToR denom, exception: false) -> Rational?
+                    | (_ToInt | _ToR numer, ?_ToInt | _ToR denom, ?exception: bool) -> Rational
+                    | [T] (Numeric&_RationalDiv[T] numer, Numeric denom, ?exception: bool) -> T
+                    | [T < Numeric] (T value, 1, ?exception: bool) -> T
+                    | (untyped, ?untyped, exception: false) -> nil
+  interface _RationalDiv[T]
+    def /: (Numeric) -> T
+  end
+  #: (Numeric | String | Object x, ?Numeric | String y, ?exception: bool exception) -> Rational
 
   # <!--
   #   rdoc-file=object.c


ParadoxV5:
> `type num = Complex|Float|int|_ToR|String`

Complex, Float and String are already `_ToR`s. So is the `Integer` part of `int`, but not necessarily the `_ToInt` part. So `num` can simply be `_ToInt|_ToR`.

sampersand:
updated to take that into account

ParadoxV5:
that’s quick … I was looking at the changes and it didn’t quite match your desc :smile:

ParadoxV5:
> However, `Kernel#Rational` _also_ tries to be a "division operator" for some reason: If the first argument is `Numeric`, but defines neither `to_int` nor `to_i` nor even `to_r`, `Kernel#Rational` will return `arg1 / arg2`. That is, it'll actually call the `/` operator on the first argument, with the second.

> Returns x/y or arg as a [Rational](https://rubyapi.org/3.2/o/rational).
> ⸺ [Ruby docs](https://rubyapi.org/o/kernel#method-i-Rational)

So this behavoir is kinda but ambiguously defined. I see you have decided to reflect it in your patch.

core/kernel.rbs
```diff
@@ -614,7 +614,15 @@ module Kernel : BasicObject
...
   #
-  def self?.Rational: (Numeric | String | Object x, ?Numeric | String y, ?exception: bool exception) -> Rational
+  def self?.Rational: (_ToInt | _ToR numer, ?_ToInt | _ToR denom, exception: false) -> Rational?
+                    | (_ToInt | _ToR numer, ?_ToInt | _ToR denom, ?exception: bool) -> Rational
+                    | [T] (Numeric&_RationalDiv[T] numer, Numeric denom, ?exception: bool) -> T
+                    | [T < Numeric] (T value, 1, ?exception: bool) -> T
+                    | (untyped, ?untyped, exception: false) -> nil
+  interface _RationalDiv[T]
+    def /: (Numeric) -> T
+  end
```
ParadoxV5, line 624: Tested and confirmed, `Kernel#Rational` indeed can return non-Rationals, like how the RBS describes. Smells more a Ruby bug than a feature.

core/kernel.rbs
```diff
@@ -614,7 +614,15 @@ module Kernel : BasicObject
...
+  #: (Numeric | String | Object x, ?Numeric | String y, ?exception: bool exception) -> Rational
```
ParadoxV5, line 625: Not sure what this’s for, but it’s conflicting the comments checker CI.

```suggestion
```

core/kernel.rbs
```diff
@@ -614,7 +614,15 @@ module Kernel : BasicObject
   #
   # See also String#to_r.
   #
-  def self?.Rational: (Numeric | String | Object x, ?Numeric | String y, ?exception: bool exception) -> Rational
+  def self?.Rational: (_ToInt | _ToR numer, ?_ToInt | _ToR denom, exception: false) -> Rational?
+                    | (_ToInt | _ToR numer, ?_ToInt | _ToR denom, ?exception: bool) -> Rational
```
ParadoxV5, line 618: With `exception: false` already covered, can this simply be
```suggestion
                    | (_ToInt | _ToR numer, ?_ToInt | _ToR denom, ?exception: true) -> Rational
```
?

Also applies to other conversion methods in Kernel.
Does not apply to the `[T]` ones if they don’t squelch exceptions as you said.

BTW,
```ruby
Rational 1, 0, exception: false #=! ZeroDivisionError: divided by 0
```

sampersand: the `: false` case is for when a typechecker can explicitly determine that it's falsey. the `: bool` variant is when the typechecker doesn't have enough information.

sampersand: It might make sense to "default" to having `Rational?` be the default

ParadoxV5: > the `: false` case is for when a typechecker can explicitly determine that it's falsey. the `: bool` variant is when the typechecker doesn't have enough information.

Not falsey, just `false`. Ruby rejects `exception: nil`. It also rejects `boolish`es, only accepting `true`.

ParadoxV5: Could there be a world where `bool` has three values? 💭

sampersand:

PR #1439, opened 2023-08-18
Updated Kernel#String

sampersand:
Simple signature cleanup for `Kernel#String`


sampersand:

PR #1441, opened 2023-08-18
Updated Kernel#srand

sampersand:
Updated `Kernel#srand` to be more correct.

Technically its first argument must respond to `to_int`, not be `Numeric`. (`srand Class.new(Numeric)` fails.) It also always returns an `Integer`.


soutaro (approved):
👍

PR #1442, opened 2023-08-18
Updated Kernel#fork

sampersand:
Minor update to `fork`: In the block form, `fork` always returns an `Integer` for the process id. The return value of the block is also unused


sampersand:
Consolidated into https://github.com/ruby/rbs/pull/1444

PR #1443, opened 2023-08-19
Updated Kernel write methods

sampersand:
This updates the signatures for all of the `Kernel` output methods (`print`, `printf`, `putc`, `puts`, `p`, and `warn`—`pp` was fine as-is), and `Object#display`.

Additionally, a `Warnings::category` type alias was added.

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -1137,7 +1137,7 @@ module Kernel : BasicObject
   #     gets  # Sets $_ to the most recent user input.
   #     print # Prints $_.
   #
-  def self?.print: (*Kernel args) -> nil
+  def self?.print: (*_ToS args) -> nil
 
   # <!--
   #   rdoc-file=io.c
@@ -1171,9 +1171,9 @@ module Kernel : BasicObject
   #
   # With no arguments, does nothing.
   #
-  def self?.printf: (IO arg0, String arg1, *untyped args) -> nil
-                  | (String arg1, *untyped args) -> nil
-                  | () -> nil
+  def self?.printf: () -> nil
+                  | (String fmt, *untyped args) -> nil
+                  | (_Writer io, string fmt, *untyped args) -> nil
 
   # <!--
   #   rdoc-file=proc.c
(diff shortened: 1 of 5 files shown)


core/kernel.rbs
```diff
@@ -1240,8 +1240,8 @@ module Kernel : BasicObject
   #     0..4
   #     [0..4, 0..4, 0..4]
   #
-  def self?.p: [T] (T arg0) -> T
-             | (untyped, untyped, *untyped) -> Array[untyped]
+  def self?.p: [T < _Inspect] (T arg0) -> T
+             | (_Inspect arg0, _Inspect arg1, *_Inspect rest) -> Array[_Inspect]
```
ParadoxV5, line 1244: `[T < _Inspect]`? Wynaut

```suggestion
             | [T < _Inspect] (T arg0, T arg1, *T rest) -> Array[T]
```

sampersand: Because that'd would require all the different arguments are the same type then

ParadoxV5: it won’t. It would fall back to `_Inspect`, since `_Inspect` is also `< _Inspect`.

core/warning.rbs
```diff
@@ -29,6 +29,7 @@
 # The `warning` gem provides convenient ways to customize Warning.warn.
 #
 module Warning
+  type category = :deprecated | :experimental | :performance
```
soutaro, line 32: Do we have `:performance` category?

```ruby
3.2.2 :002 > Warning[:performance]
(irb):2:in `[]': unknown category: performance (ArgumentError)
	from (irb):2:in `<main>'
	from /Users/soutaro/.rvm/gems/ruby-3.2.2/gems/irb-1.7.4/exe/irb:9:in `<top (required)>'
	from /Users/soutaro/.rvm/gems/ruby-3.2.2/bin/irb:25:in `load'
	from /Users/soutaro/.rvm/gems/ruby-3.2.2/bin/irb:25:in `<main>'
```

soutaro (changes requested):
Could you confirm if we have `:performance` category in Ruby 3.2?

core/warning.rbs
sampersand: In 3.3.0 we do, it was just added

soutaro: Hmm. RBS definition is based on the latest release of Ruby, 3.2 now.
So..., could you delete it? (because test is failing anyway.)

sampersand: Sure, I'll do that with a few other things I've added from 3.3

soutaro (approved):
👍

PR #1444, opened 2023-08-19
Updated control flow type definitions

sampersand:
Update `Kernel` "control flow" method definitions:
- `catch`: (nothing to do)
- `throw`: make the tag `untyped`
- `at_exit`: block return type is `void`
- `exit` and `exit!`: use `bool` instead of `TrueClass | FalseClass`; Also, use `int` instead of `Integer` for status
- `fork`: block return type is `void`, return type in block form is no longer `Integer?` but just `Integer`
- `loop`: block return type is `void`
- `sleep`: Add in support for Ruby-3.3.0 preview's `sleep(nil)`
- `raise` / `fail`: Add support for `raise has_a_to_str_method`, Made the `message` in the long-form `_ToS`, and allow for `nil` and `String` backtraces.

Technically, the `message` in the long form can be any type, as long as the first argument's `exception` function allows for it. So, we could do `[T] (_Exception[T], ?T?, ...`, but that seems like overkill. (Especially since `_Exception` is used in `Thread` and `Fiber`. Maybe when I get around to I could change the signature?)

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -383,7 +383,7 @@ module Kernel : BasicObject
   # 4. Therefore you should use spawn() instead of fork().
   #
   def self?.fork: () -> Integer?
-                | () { () -> untyped } -> Integer?
+                | () { () -> void } -> Integer
 
   def initialize_copy: (self object) -> self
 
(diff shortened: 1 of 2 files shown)


core/kernel.rbs
```diff
@@ -757,6 +757,7 @@ module Kernel : BasicObject
   #
   def self?.binding: () -> Binding
 
+
```
soutaro, line 760: Delete this line to make CI pass.

soutaro (changes requested):
Extra empty line breaks the CI.

soutaro (approved):
👍

PR #1445, opened 2023-08-19
Updated Kernel conversion methods

sampersand:
This is a merging of the previous individual PRs for conversion methods into one, larger PR. (#1433 , #1434, #1435, #1436, #1437, #1438, #1439). Look at them for a little more in-depth details, adn some discussion topics. Summary of changes is below:

- `Kernel#Array`: Updated to represent the true definition: Only `.to_ary` and `.to_a` are used for conversions, and everything else is boxed up in an array.
- `Kernel#Complex`: Added the `.to_c` variant.
- `Kernel#Float`: Changed from `Numeric | String` to just be `_ToF`, which is what's used `internally.
- `Kernel#Hash`: The argument is no longer an `Object`: Instead, `.to_hash` is required for its argument, with the special case of `nil` and empty arrays (and _only_ empty arrays) returning empty hashes.
- `Kernel#Integer`: Updated to reflect the ability to pass implicit `.to_str` and `.to_int` to all variants, and only `.to_i` (not being a subclass of `Numeric`) for the one-argument case
- `Kernel#Rational`: Oh boy, this method is a mess. Check out the [original issue](https://github.com/ruby/rbs/pull/1438) for more details. tl;dr: If the first argument isn't `_ToInt | _ToR`, but is `Numeric`, it attempts to do division with it, which allows for returning arbitrary types.
- `Kernel#String`: Simply cleaned up naming; it was semantically correct beforehand.

For all conversion methods with an `exception: bool` field, two variants have been added: `(..., ?exception: true) -> type` and `(..., exception: bool) -> type?`. This reflects the fact that with `exception: false` (or an unknown boolean value, which may be false), the return type is `nil`.

Additionally, all the `exception: bool` variants, when given invalid types but with `exception: false`, will return `nil`. That's also been accounted for.

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -403,12 +403,10 @@ module Kernel : BasicObject
   #
   #     Array(:foo)             # => [:foo]
   #
-  def self?.Array: (NilClass x) -> [ ]
-                 | [T] (::Array[T] x) -> ::Array[T]
-                 | [T] (::Range[T] x) -> ::Array[T]
-                 | [T] (_Each[T] x) -> ::Array[T]
-                 | [K, V] (::Hash[K, V] x) -> ::Array[[ K, V ]]
-                 | [T] (T x) -> ::Array[T]
+  def self?.Array: (nil) -> []
+                 | [T] (Array[T] ary) -> Array[T]
+                 | [T] (_ToAry[T] | _ToA[T] array_like) -> Array[T]
+                 | [T] (T ele) -> [T]
 
   # <!--
   #   rdoc-file=complex.c
@@ -446,8 +444,11 @@ module Kernel : BasicObject
   #
   # See String#to_c.
   #
-  def self?.Complex: (Numeric | String x, ?Numeric | String y, ?exception: bool exception) -> Complex
-
+  def self?.Complex: (_ToC complex_like, ?exception: true) -> Complex
+                   | (_ToC complex_like, exception: bool) -> Complex?
+                   | (Numeric | String real, ?Numeric | String imag, ?exception: true) -> Complex
+                   | (Numeric | String real, ?Numeric | String imag, exception: bool) -> Complex?
+                   | (untyped, ?untyped, exception: false) -> nil
   # <!--
   #   rdoc-file=kernel.rb
   #   - Float(arg, exception: true)    -> float or nil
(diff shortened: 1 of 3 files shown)


core/kernel.rbs
```diff
@@ -488,7 +490,8 @@ module Kernel : BasicObject
   #     Hash(nil)              # => {}
   #     Hash([])               # => {}
   #
-  def self?.Hash: [K, V] (Object x) -> ::Hash[K, V]
+  def self?.Hash: [K, V] (nil | [] _empty) -> Hash[K, V]
```
ParadoxV5, line 493: `nil` and `[]` are already `_ToHash[untyped, untyped]`s. But if you want to add them here explicitly, note the `[K, V]` are unused.

```suggestion
  def self?.Hash: (nil | [] _empty) -> Hash[untyped, untyped]
```

ParadoxV5: Oh right they aren’t `_ToHash`es, just `_ToH`s.
So this method only checks for `#to_hash` but not `#to_h`, unlike `Integer` and `Array`?

core/kernel.rbs
```diff
@@ -614,7 +619,16 @@ module Kernel : BasicObject
...
   #
-  def self?.Rational: (Numeric | String | Object x, ?Numeric | String y, ?exception: bool exception) -> Rational
+  def self?.Rational: (_ToInt | _ToR rational_like, ?exception: true) -> Rational
+                    | (_ToInt | _ToR rational_like, exception: bool) -> Rational?
+                    | (_ToInt | _ToR numer, ?_ToInt | _ToR denom, ?exception: true) -> Rational
+                    | (_ToInt | _ToR numer, ?_ToInt | _ToR denom, exception: bool) -> Rational?
+                    | [T] (Numeric&_RationalDiv[T] numer, Numeric denom, ?exception: bool) -> T
+                    | [T < Numeric] (T value, 1, ?exception: bool) -> T
+                    | (untyped, ?untyped, exception: false) -> nil
+  interface _RationalDiv[T]
```
soutaro, line 629: Need a new line before `_RationalDiv`.

soutaro (changes requested):
Looks good. Some newline problems to make the CI pass.

soutaro (approved):
I will take a look at the CI fails.

PR #1450, opened 2023-08-21
Cleaned up Errno, added a few missing ones

sampersand:
This cleans up the definition of `errno.rbs`, making the classes live under the `Errno` module, and aren't defined outside.

In addition, a few missing errors were added: `NOERROR`, `EPWROFF`, `EQFULL`, `ESHLIBVERS`, `EBADARCH`, `EBADEXEC`, `EBADMACHO`, `EDEVERR`, `ELAST`, and `ENOPOLICY`. 

You can find the list of [errno constants here](https://github.com/ruby/ruby/blob/master/defs/known_errors.def), with the caveat that `NOERROR` is defined [here](https://github.com/ruby/ruby/blob/master/error.c#L3588).


soutaro:
Indentation mismatch caused the failure. Could you fix that with two spaces?

sampersand:
Oh, oops! I forgot to set subliem to always us e2 spaces for rrbs

soutaro (approved):
Thanks! 🎉

PR #1452, opened 2023-08-22
updated dir.rbs

sampersand:
This updates `Dir.rbs`, as well as adds in the `path` type alias.

The interface `_ToPath | _ToStr` is frequently used in file-system related functions, so it makes sense to have an alias for that.

All of the `void` functions were replaced with their actual return types (ie `0` or `nil`).

Detailed breakdown:
- `Dir#initialize`: `dir` is now `path`.
- `Dir.[]`: Both `*patterns` and `base:` now accept `_ToPath`; `sort` was added.
- `Dir.chdir`: `dir` is now `path`.
- `Dir.children`: `dirname` is now `path`.
- `Dir.chroot`: `root` is now `path`.
- `Dir.delete`: `dirname` is now `path`.
- `Dir.each_child`: `dirname` is now `path`.; Also, the `Enumerator` form's `Return` param is now `nil`, not `void`.
- `Dir.empty?`: `path_name` is now `path`.
- `Dir.entries`: `dirname` is now `path`.
- `Dir.exist?`: `file_name` now accepts `path`, and strangely enough, `_ToIo`. (It uses `rb_stat` internally, which can take `_ToIo`s.)
- `Dir.glob`: `pattern` now accepts `_ToAry[path]` and `path`; `flags` is now `int`; `base` now accepts `path`; `sort` was added.
- `Dir.home`: The `user` is now optional
- `Dir.mkdir`: `dirname` now accepts `path`; Also, `permissions` now is `int`.
- `Dir.open`: `dirname` now accepts `path`; Also, now returns `instance`.
- `Dir.pwd`: Now an alias of `Dir.getwd`, like it is in the source code.
- `Dir#pos=`: Now accepts `int`, and returns the original type.
- `Dir#seek`: Now accepts `int`.
- `Dir#tell`: Now an alias of `Dir#pos`, like it is in the source code.


sampersand:
(Note I didn't update the tests yet)

soutaro (approved):
👍 

> All of the void functions were replaced with their actual return types (ie 0 or nil).

I talked with @sampersand about this and understand it's better, because some Ruby code actually depends on the return values.
I don't have plan to fix everything right now, but I'm open to merge the changes.

PR #1453, opened 2023-08-22
doc: Add annotations notation to syntax.md

tk0miya:
I found some `parse_annotations()` calls in the [parser.c](https://github.com/ruby/rbs/blob/master/ext/rbs_extension/parser.c).

* parser_member_def()
* parse_module_members()
* parse_module_members()
* parse_decl()

Additionally, [signature_parsing_test.rb](https://github.com/ruby/rbs/blob/master/test/rbs/signature_parsing_test.rb) also tells us where we can insert annotations.

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -764,3 +764,19 @@ interface _Foo
   def new: () -> Foo
 end
 ```
+
+### Annotations
+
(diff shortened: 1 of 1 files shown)


soutaro (approved):
Thanks!

PR #1455, opened 2023-08-22
Add array[E], hash[K,V] and documents most interfaces

sampersand:
This adds the `array[E] = Array[E] | _ToAry[E]` and `hash[K, V] = Hash[K, V] | _ToHash[K, V]` types, as well as adds documentation to all `_To*` interfaces and all type declarations within `builtin.rbs`

### DIFF
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -1,60 +1,172 @@
+# A type that's convertible to a `Complex`.
+#
 interface _ToC
+  # Convert `self` to a `Complex`.
+  #
   def to_c: () -> Complex
 end
 
+# A type that's convertible to a `Rational`.
+#
 interface _ToR
+  # Convert `self` to a `Complex`.
+  #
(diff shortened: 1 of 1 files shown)


ParadoxV5:
I would rather like to know the purposes of `int`, `string` and `io` besides one word beïng prettier than `_ToXXX`.

sampersand:
> I would rather like to know the purposes of `int`, `string` and `io` besides one word beïng prettier than `_ToXXX`.

I talked with soutaro last night about this actually. The idea is that while technically redundant, they exist for documentation purposes: If you accept `string x`, that means the cannonical case is accepting a `String`, but it also can accept a `.to_str` if needed. Compare to if you just accept `_ToStr`, that implies that you generally want a class that defines `.to_str`.

So the idea is, in the stdlib, you always want to use `int` / `string` / `io` and friends

soutaro (approved):
👍

Issue #1253, opened 2023-02-24
Struct has almost none of its method defined


sdaubert:
In rbs file for Struct class, there is no definition for:

* []
* []=
* deconstruct
* deconstruct_keys
* dig
* length
* members (there is a definition on class level, but not on instance level)
* size
* values
* values_at

sampersand:
You're right, that's bizarre. I'll work on that soon

sampersand:
https://github.com/ruby/rbs/pull/1456

PR #1456, opened 2023-08-22
Add type annotations for Struct

sampersand:
This rewrites the entirety of `Struct`'s annotations.

It relies on #1492 to be merged so that it can use the `with_xxx` methods.

The type annotation on `Struct` is a bit unfortunate: It requires all instance variables to be of the same type (or union type), such as `Struct[Integer | String | bool]`. Note that `Struct.new(...).new` can take fewer arguments than fields, and the remaining fields are just filled in. To accommodate for this behaviour, I originally had things like `[]` return `Elem?`. However, I realized that you can just do something like `Struct[Integer?]` to allow for this.

There's still some weirdness with how `Struct.new` and `Struct.new(...).new` play out with type interfaces, so an extra branch `| (*Elem fields, **Elem keyword_fields) -> instance` was added to `Struct.new`'s definition. Unit testing this proved to be annoying, and I didn't end up being able to test the "Struct subclass" `new`.

More concretely, the changes are:
- Added tests for all of `Struct`'s methods
- Imported documentation for all of `Struct`
- Removed redundant `< Object` from `Struct`
- Struct includes `Enumerable[Elem]` not `Enumerable[Elem?]` now. (See above for how optional arguments to `new` works)
- Removed `attribute_name` as it was only used (a bit incorrectly) in `initialize`
- Added `type index = String | Symbol | int` which is used in `[]`, `[]=`, `dig`, and `deconstruct_keys`
- Updated `Struct.new` to more correctly reflect how creation works, and also added the optional struct name. (See above for weirdness with `Struct.new`.)
- `Struct.members`: removed leading `::` from `Array[Symbol]`
- `Struct.keyword_init?`: returns `bool?` now, instead of `true | false | nil`
- Added in definitions for `Struct` instance methods: `==`, `eql?`, `hash`, `inspect`, `to_a`, `to_h`, `size`, `each`, `each_pair`, `[]`, `[]`, `select`, `values_at`, `members`, `dig`, `deconstruct_keys`, `to_s`, `values`, `length`, `filter`, and `deconstruct`.

### DIFF
--- a/core/struct.rbs
+++ b/core/struct.rbs
@@ -46,12 +46,10 @@
 # *   Includes [module Enumerable](rdoc-ref:Enumerable@What-27s+Here), which
 #     provides dozens of additional methods.
 #
-#
 # See also Data, which is a somewhat similar, but stricter concept for defining
 # immutable value objects.
 #
 # Here, class Struct provides methods that are useful for:
-#
 # *   [Creating a Struct
 #     Subclass](rdoc-ref:Struct@Methods+for+Creating+a+Struct+Subclass)
 # *   [Querying](rdoc-ref:Struct@Methods+for+Querying)
@@ -61,25 +59,23 @@
 # *   [Iterating](rdoc-ref:Struct@Methods+for+Iterating)
 # *   [Converting](rdoc-ref:Struct@Methods+for+Converting)
 #
-#
 # ### Methods for Creating a Struct Subclass
 #
-# *   ::new: Returns a new subclass of Struct.
-#
+# *   ::new: Returns a new subclass of Struct.#
 #
 # ### Methods for Querying
 #
 # *   #hash: Returns the integer hash code.
-# *   #length, #size: Returns the number of members.
-#
+# *   #length, #size: Returns the number of members.#
 #
 # ### Methods for Comparing
 #
-# *   #==: Returns whether a given object is equal to `self`, using `==` to
-#     compare member values.
-# *   #eql?: Returns whether a given object is equal to `self`, using `eql?` to
-#     compare member values.
-#
+# [#==](#method-i-3D-3D)
+# :   Returns whether a given object is equal to `self`, using `==` to compare
+#     member values.
+# #eql?
+# :   Returns whether a given object is equal to `self`, using `eql?` to compare
+#     member values.#
 #
 # ### Methods for Fetching
 #
@@ -103,19 +99,24 @@
 #
 # ### Methods for Iterating
 #
-# *   #each: Calls a given block with each member name.
-# *   #each_pair: Calls a given block with each member name/value pair.
+# *   #inspect, #to_s: Returns a string representation of `self`.
+# *   #to_h: Returns a hash of the member name/value pairs in `self`.
 #
 #
 # ### Methods for Converting
 #
-# *   #inspect, #to_s: Returns a string representation of `self`.
-# *   #to_h: Returns a hash of the member name/value pairs in `self`.
+# #inspect, #to_s
+# :   Returns a string representation of `self`.
+# #to_h
+# :   Returns a hash of the member name/value pairs in `self`.
 #
-class Struct[Elem] < Object
-  include Enumerable[Elem?]
+class Struct[Elem]
+  include Enumerable[Elem]
 
-  type attribute_name = interned
+  # The types that can be used when "indexing" into a `Struct` via `[]`, `[]=`, `dig`, and
+  # `deconstruct_keys`.
+  #
+  type index = String | Symbol | int
 
   # <!--
   #   rdoc-file=struct.c
@@ -128,6 +129,8 @@ class Struct[Elem] < Object
   #
   # *   May be anonymous, or may have the name given by `class_name`.
   # *   May have members as given by `member_names`.
+  # *   May have initialization via ordinary arguments (the default) or via
+  #     keyword arguments (if `keyword_init: true` is given).
   # *   May have initialization via ordinary arguments, or via keyword arguments
   #
   #
@@ -203,6 +206,7 @@ class Struct[Elem] < Object
   #         Foo.new(foo: 0, bar: 1, baz: 2)
   #         # Raises ArgumentError: unknown keywords: baz
   #
+  #
   #     Method `::[]` is an alias for method `::new`.
   #
   # *   Method `:inspect` returns a string representation of the subclass:
@@ -219,7 +223,6 @@ class Struct[Elem] < Object
   #
   # By default, the arguments for initializing an instance of the new subclass can
   # be both positional and keyword arguments.
-  #
   # Optional keyword argument `keyword_init:` allows to force only one type of
   # arguments to be accepted:
   #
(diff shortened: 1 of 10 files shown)


ParadoxV5:
might wanna undo the `Gemfile.lock` changes and open a separate PR for those breaking changes (yeuh, `bundle update` is prolly gonna break things)

core/struct.rbs
```diff
@@ -1,299 +1,51 @@
...
+
+  def each: () -> Enumerator[Elem?, self]
+          | () { (Elem? value) -> void } -> self
+
+  def each_pair: () -> Enumerator[[Symbol, Elem?], self]
+               | () { ([Symbol, Elem?] key_value) -> void } -> self
+
+  def []: (index name_or_position) -> untyped
+  def []=: (index name_or_position, Elem value) -> Elem
+         | (index name_or_position, nil) -> nil
```
ParadoxV5, line 35:
```suggestion
  def []=: [E < Elem?] (index name_or_position, E) -> E
```
?

and there ought to be many more `[]=`s and `attr_writer`s out there that don’t make use of the type variable

core/struct.rbs
```diff
@@ -1,299 +1,51 @@
...
-#
-#
-# ### Methods for Converting
-#
-# *   #inspect, #to_s: Returns a string representation of `self`.
-# *   #to_h: Returns a hash of the member name/value pairs in `self`.
-#
-class Struct[Elem] < Object
+class Struct[Elem]
   include Enumerable[Elem?]
```
ParadoxV5, line 2: Why `Elem?` with `?` in so many places? If the user would like to have, say, String or `nil` for values, can’t they have a `Struct[String?]`?

core/struct.rbs
```diff
@@ -1,299 +1,51 @@
...
+               | () { ([Symbol, Elem?] key_value) -> void } -> self
+
+  def []: (index name_or_position) -> untyped
+  def []=: (index name_or_position, Elem value) -> Elem
+         | (index name_or_position, nil) -> nil
+
+  def select: () -> Enumerator[Elem?, Array[Elem?]]
+            | () { (Elem? value) -> boolish } -> Array[Elem?]
+
+  alias filter select
```
ParadoxV5, line 40: do these have the same signatures as the `select` from `Enumerable[Elem?]`?

core/struct.rbs
```diff
@@ -1,299 +1,51 @@
...
-  # Examples:
-  #     Foo = Struct.new(:a)
-  #     Foo.keyword_init? # => nil
-  #     Bar = Struct.new(:a, keyword_init: true)
-  #     Bar.keyword_init? # => true
-  #     Baz = Struct.new(:a, keyword_init: false)
-  #     Baz.keyword_init? # => false
-  #
-  def self.keyword_init?: () -> (true | false | nil)
+  def to_a: () -> Array[Elem?]
```
ParadoxV5, line 19: does this have the same signature as the `#to_a` from `Enumerable[Elem?]`?

core/struct.rbs
```diff
@@ -1,299 +1,51 @@
...
-#
-# *   #each: Calls a given block with each member name.
-# *   #each_pair: Calls a given block with each member name/value pair.
-#
-#
-# ### Methods for Converting
-#
-# *   #inspect, #to_s: Returns a string representation of `self`.
-# *   #to_h: Returns a hash of the member name/value pairs in `self`.
-#
```
ParadoxV5, line 114: The removal of the docs is making the diffs real ugly – yet you need the docs in to at least please the CI

core/struct.rbs
sampersand: That doesn't work, the rhs of `<` needs to be interfaces, otherwise stuff like this would be significantly easier:
```
  # <!--
  #   rdoc-file=vm_method.c
  #   - public                                -> nil
  #   - public(method_name)                   -> method_name
  #   - public(method_name, method_name, ...) -> array
  #   - public(array)                         -> array
  # -->
  # With no arguments, sets the default visibility for subsequently defined
  # methods to public. With arguments, sets the named methods to have public
  # visibility. String arguments are converted to symbols. An Array of Symbols
  # and/or Strings is also accepted. If a single argument is passed, it is
  # returned. If no argument is passed, nil is returned. If multiple arguments are
  # passed, the arguments are returned as an array.
  #
  def public: () -> nil
            | (Symbol method_name) -> Symbol
            | (Symbol, Symbol, *Symbol method_names) -> Array[Symbol]
            | [T < _ToStr] (T method_name) -> T
            | [T < _ToStr] (T, T, *T method_names) -> Array[T]
            | (interned, interned, *interned method_names) -> Array[interned]
            | [T < _ToAry[interned]] (T method_names) -> T
```

ParadoxV5: Why must the RHS be interfaces?
… *tests* RubyMine says the RHS can also be Classes and Modules, but not Type Variables, let alone arbitrary types (here, a `nil`able). Wynaut? !featurerequest

core/struct.rbs
sampersand: Yeah, the docs are the reason why it's still a draft

core/struct.rbs
sampersand: No. Since  you can do `My3dPoint.new(1)`, the extra elements are assigned as `nil`. Thus, anything that interacts with them are `Elem?`.

core/struct.rbs
sampersand: No. Since  you can do `My3dPoint.new(1)`, the extra elements are assigned as `nil`. Thus, anything that interacts with them are `Elem?`.

core/struct.rbs
sampersand: It's cause there's no way to guarantee every element is assigned. Since you can do `My3dPoint.new(1)`, the extra elements are assigned as nil. Thus, anything that interacts with them are `Elem?`.

core/struct.rbs
sampersand: do it

ParadoxV5: #1471

core/struct.rbs
ParadoxV5: Yes, but with `Enumerable[Elem?]`, everything from `Enumerable` are already `Elem?`s. Here I’m pointing out that by applying `Elem?` to `Enumerable`’s type variable `Elem`, both types have the same `#select`, `#filter` and `#to_a`.

core/struct.rbs
ParadoxV5: :arrow_right: `#select` comment https://github.com/ruby/rbs/pull/1456#discussion_r1306588308

ParadoxV5:
Fixes #1253

core/struct.rbs
sampersand: Ah, yes. However, i'm, annotating any methods that are explicitly defined in the base class and not inherited

ParadoxV5: Detailing what’s inherited and what’s overriden isn’t RBS’s job…

13k:
(I suppose my question is out of the scope of this PR, but I'm not sure where to discuss this.)

Has the definition `Struct[K, V]` been considered and discarded before?
 
I'm aware it could raise a lot of issues (`K` would have to be upper bound by `Symbol`; object totality (like mypy's `TypedDict`); etc) and I don't know if rbs is equipped to deal with those.

Anyway, it would be good to have type-checked, well-defined, data structures.

sampersand:
> (I suppose my question is out of the scope of this PR, but I'm not sure where to discuss this.)
> 
> Has the definition `Struct[K, V]` been considered and discarded before?
> 
> I'm aware it could raise a lot of issues (`K` would have to be upper bound by `Symbol`; object totality (like mypy's `TypedDict`); etc) and I don't know if rbs is equipped to deal with those.
> 
> Anyway, it would be good to have type-checked, well-defined, data structures.

How would the `K` work? All the `Symbol`s are supplied internally by Ruby, so there's no way for end-users to modify it.

13k:
> > (I suppose my question is out of the scope of this PR, but I'm not sure where to discuss this.)
> > Has the definition `Struct[K, V]` been considered and discarded before?
> > I'm aware it could raise a lot of issues (`K` would have to be upper bound by `Symbol`; object totality (like mypy's `TypedDict`); etc) and I don't know if rbs is equipped to deal with those.
> > Anyway, it would be good to have type-checked, well-defined, data structures.
> 
> How would the `K` work? All the `Symbol`s are supplied internally by Ruby, so there's no way for end-users to modify it.

I was thinking of `K` being defined as a union of literals that would limit the set of valid keys:

```rbs
class Person < Struct[:name | :age, ElemType]
  # ...
end

person = Person.new()
person.name = "Alice" # valid
person.nickname = "Al" # invalid
```

As it is now, any Symbol satisfies the definition and accessing `nickname` wouldn't raise a type error (but would break at runtime).

13k:
An aside, more general, discussion regarding data types, as I've seen "in the wild", is that there's efforts to support these kinds of strongly typed key/value collections, namely Python's [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) and TypeScript's `interface` creation with [mapped types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html).

I think there's value in considering something like those, but they sit on top of big infrastructures that support them.

- Python's `TypedDict`-like:

```rbs
# Equivalent of `Person = Struct.new(:name, :age)` but with type-checked values
class Person < Struct
  attr_accessor name (): String
  attr_accessor age (): Integer
end
```

Requires a type to "introspect itself" and "expand" their definition from there.

- TypeScripts's approach:

```rbs
# using a record type as shape
class Person < Struct[{ name: String, age: Integer }]
  # `Struct[R]` expands key/value pairs to attributes + Struct's API
end

# or using an interface as shape (if it would be possible to have attributes in interfaces)
interface _Person
  attr_accessor name (): String
  attr_accessor age (): Integer
end

class Person < Struct[_Person]
  # `Struct[I]` expands the interfaces attributes to class attributes + Struct's API
end
```

ParadoxV5:
> I was thinking of `K` being defined as a union of literals that would limit the set of valid keys:
> 
> ```rbs
> class Person < Struct[:name | :age, ElemType]
>   # ...
> end
> ```

That makes sense. Hash is `[K, V]`, so can Struct `[K < Symbol, V]`.

However, `K` would only be used in things like `#[]` and `#each_pair`.

> ```ruby
> person = Person.new()
> person.name = "Alice" # valid
> person.nickname = "Al" # invalid
> ```

These accessor methods, howëver, are not possible with Struct’s RBS, with or without `K`. They are dynamic and unique to the Person Struct subclass.
I see there are two routes:
* Ask your type checker if they can support understanding these Struct accessors dynamically.
* Create your own RBS for Person
  ```rbs
  class Person < Struct[…]
    def name: () -> String
    def name=: [T < String] (T) -> T
  end
  ```

13k:
>     * Create your own RBS for Person
>       ```rbs
>       class Person < Struct[…]
>         def name: () -> String
>         def name=: [T < String] (T) -> T
>       end
>       ```

Yeah, I went this route, specially for structs with heterogeneous value types.

Thanks for clarifying.

sampersand:
The holdup with this was that there was no way to overload the `new` declaration. Talked with @soutaro at RubyConf, and we decided that what I've done is fine, and users can just add their own `new`s in type definitiosn if needed. e.g.:
```rb
Point = Struct.new(:x, :y) do def to_s = "(#{x}, #{y})" end
```
would become
```rbs
class Point < Struct[Numeric]
  attr_reader x: Numeric

  attr_reader y: Numeric
  
  def initialize: (Numeric x, Numeric y) -> void
  
  def to_s: () ->  String
end
```

There's a few minor things I have to finalize but this should be ready for review soon

sampersand:
I cant figure out how to get git to play nice, so I submitted a new PR: https://github.com/ruby/rbs/pull/1456

ParadoxV5:
> I cant figure out how to get git to play nice, so I submitted a new PR: #1456

Did you mean?  #1621

ParadoxV5:
> I cant figure out how to get git to play nice,

Force pushes from rebasing doesn’t play nice with git logs. That’s why I avoid those on pushed commits.

PR #1457, opened 2023-08-23
[prototype runtime] Avoid alias for methods defined in another module

ksss:
`define_method` and `define_singleton_method` can take UnboundMethod in second argument.
In addition, it can be specified even for methods of a completely different module.

For example, erubi uses this technique.

https://github.com/jeremyevans/erubi/blob/4d24561e79e291208996892800a5cb9d11b8d2af/lib/erubi.rb#L23

`rbs prototype runtime` produces the following output for the current erubi.

```rbs
$ rbs prototype runtime -r erubi Erubi
module Erubi
  alias self.h self.html_escape 

  ...snip...
```

However, the specified method does not exist.

```
Unknown method alias name: html_escape => h (::Erubi) (RBS::UnknownMethodAliasError)

    alias self.h self.html_escape
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

To avoid this problem, I suggest that in this case the output be as a method definition, not an alias.

### DIFF
--- a/lib/rbs/prototype/runtime.rb
+++ b/lib/rbs/prototype/runtime.rb
@@ -242,7 +242,16 @@ module RBS
         mod.singleton_methods.select {|name| target_method?(mod, singleton: name) }.sort.each do |name|
           method = mod.singleton_class.instance_method(name)
 
-          if method.name == method.original_name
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks!

PR #1463, opened 2023-08-24
String keys are forbidden.

ioquatix:
https://github.com/ruby/ruby/pull/8273


soutaro:
@ioquatix Will the change in https://github.com/ruby/ruby/pull/8273 be available in 3.2.3 (before 3.3)?

ioquatix:
I hope so, it's marked for backport.

PR #1465, opened 2023-08-24
Fix error when undefined interface or alias

ksss:
If a non-existent interface or alias is specified as shown below, an unexpected error will occur during validation.

```rbs
# t.rbs
class Foo
  def void: () -> _Void
end
```

```
$ rbs -I t.rbs validate --silent
/lib/rbs/variance_calculator.rb:140:in `type': undefined method `decl' for nil:NilClass (NoMethodError)

                          env.interface_decls[type_name].decl.type_params
                                                        ^^^^^
```

I expect the following to work.
This behavior is the same as specifying a non-existent class name.

```
$ rbs -I t.rbs validate --silent
/lib/rbs/variance_calculator.rb:159:in `type': t.rbs:2:18...2:23: Could not find _Void (RBS::NoTypeFoundError)

    def void: () -> _Void
                    ^^^^^
```

Type alias also.

```
$ rbs -I t.rbs validate --silent
/lib/rbs/variance_calculator.rb:159:in `type': t.rbs:2:18...2:23: Could not find voida (RBS::NoTypeFoundError)

    def void: () -> voida
                    ^^^^^
```

I have validated RBS of some size for this change and found no significant degradation in performance.

### DIFF
--- a/lib/rbs/environment.rb
+++ b/lib/rbs/environment.rb
@@ -276,9 +276,9 @@ module RBS
     end
 
     def normalize_type_name?(name)
-      if name.class?
-        normalize_module_name?(name)
-      else
+      return normalize_module_name?(name) if name.class?
+
+      type_name =
         unless name.namespace.empty?
           parent = name.namespace.to_type_name
           parent = normalize_module_name?(parent)
@@ -288,7 +288,8 @@ module RBS
         else
           name
         end
-      end
+
+      type_name?(type_name) ? type_name : false
     end
 
     def normalize_type_name!(name)
--- a/lib/rbs/environment.rb
+++ b/lib/rbs/environment.rb
@@ -289,7 +289,9 @@ module RBS
           name
         end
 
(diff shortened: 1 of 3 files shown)


soutaro (changes requested):
Thank you for investigating the problem and writing a patch.
I think `#normalize_type_name?` should check if the final type name is defined or not, like `normalize_module_name?`.

So, the use-site (changed in this PR) should be left unchanged, and the implementation of `#normalize_type_name?` should be changed instead.

lib/rbs/environment.rb
```diff
@@ -288,7 +288,8 @@ def normalize_type_name?(name)
         else
           name
         end
-      end
+
+      type_name?(type_name) ? type_name : false
```
soutaro, line 292:
```suggestion
      if type_name?(type_name)
        type_name
      end
```

We want `nil` if type name is not present. `false` is for recursive module aliases.

soutaro (approved):

PR #1469, opened 2023-08-25
Added type definitions for float, _ToRange[T], range[T], and interned

sampersand:
This adds in the `_Range[T]` interface, and `range[T]`, `float`, and `interned` aliases.

- `_Range[T]` is the `Range` equivalent of `_ToInt`, and is used in _many_ places.
- `range[T]` is an alias for `Range[T] | _Range[T]`, akin to `string` and `int`.
- `float` is an alias for `Float | _ToF`, which is used in quite a few places (eg `Regexp#timeout=`)
- `interned` is an alias for `Symbol | string`, which is frequently used, even in unexpected areas (eg `Regexp#=~`)

Note that I used `interned` instead of `id`, as I suspect people'll be much more likely to define `id` aliases, and find conflicts, than `interned`.

### DIFF
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -198,6 +198,22 @@ interface _Rewindable
   def rewind: () -> Integer
 end
 
+# A type that's usable like a `Range[T]`.
+#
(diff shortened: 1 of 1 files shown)


soutaro (approved):
👍

Issue #1449, opened 2023-08-21
Feature proposal: `rbs todo` command


ksss:
# Problem

## Finding missing RBS definitions

Identifying which RBS definitions are missing is a difficult and painstaking task.
The person writing the RBS does not know what the missing RBS is.

## Make sure the RBS is practical.

The RBS writer is not confident in the RBS described.
The only way is to start the type checker (steep) and try to write the source code.

# Feature proposal: `rbs todo` command

`rbs todo` command display a list of unimplemented methods.

When check for unimplemented methods in the IO class.

```
$ rbs todo IO
- def pread
- def pwrite
- def to_path
...
```

# Yet another way: `rbs prototype runtime --todo`

`--todo` option acts like a filter to display only methods that are not yet implemented in defined RBS.

```
$ rbs prototype runtime --todo IO
class IO
  def pread: (*untyped) -> untyped
  def pwrite: (untyped, untyped) -> untyped
  def to_path: () -> untyped
end
```

PR #1470, opened 2023-08-26
Implement todo option

ksss:
The todo option excludes predefined output from prototype runtime.

## Issue

https://github.com/ruby/rbs/issues/1449

## Supported features

This PR is implemented to exclude the following.

- Instance method
- Singleton method
- Constant
- Mixin

## Demo

```rbs
$ bundle exec rbs prototype runtime --todo IO
class IO
  public

  def external_encoding: () -> untyped

  def pread: (*untyped) -> untyped

  def pwrite: (untyped, untyped) -> untyped

  def to_path: () -> untyped

  def wait_priority: (*untyped) -> untyped
end
```

### DIFF
--- a/lib/rbs/cli.rb
+++ b/lib/rbs/cli.rb
@@ -673,6 +673,7 @@ EOU
             merge = true
           end
           opts.on("--todo", "Generates only undefined methods compared to objects") do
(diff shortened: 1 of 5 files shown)


soutaro (approved):
Thanks!

PR #1478, opened 2023-08-29
Updated FalseClass, NilClass, and TrueClass

sampersand:
This primarily updates the unit tests for `FalseClass`, `NilClass`, and `TrueClass` to the newer system. It also:
- Removed `public` from the beginning of definitions (it's redundant)
- Removed `def clone: (?freeze: true?) -> self`, as that's defined on `Kernel`
- Collapsed redundant branches (`(nil) -> false | (false) -> false` is now just `(nil | false) -> false`. I decided not to do `(false?) -> false`, as the nil variant's important for the method definition, and doesn't just indicate an absence of a value)
- For `NilClass`, `to_h` now returns `{}`, not `::Hash[untyped, untyped]`. It maintains parity with the other methods (eg `def to_a: () -> []`)

### DIFF
--- a/core/false_class.rbs
+++ b/core/false_class.rbs
@@ -4,8 +4,6 @@
 # operators allowing `false` to participate correctly in logical expressions.
 #
 class FalseClass
-  public
-
   def !: () -> true
 
   # <!--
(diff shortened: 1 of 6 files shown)


soutaro (approved):
👏 👏 👏

PR #1485, opened 2023-08-30
[prototype runtime] Fix error when constant is a BasicObject instance.

ksss:
BasicObject instances are difficult to determine because the methods that can be called are very limited.

There are actual examples, and I think it should be fixed.

https://github.com/Shopify/bootsnap/blob/b75c683feb8c9b4b1da72fe19c788ad112c7fd96/lib/bootsnap/compile_cache.rb#L5

### DIFF
--- a/test/rbs/runtime_prototype_test.rb
+++ b/test/rbs/runtime_prototype_test.rb
@@ -533,29 +533,34 @@ end
     end
   end
 
-  class Unnamed
+  class Constants
     module Name
       class Space
       end
     end
     A = ARGF
     B = ENV
+    C = BasicObject.new
     D = Name::Space.new
+    E = Class.new # skip
+    F = Module.new # skip
   end
 
-  def test_unnamed
+  def test_constants
     SignatureManager.new do |manager|
       manager.build do |env|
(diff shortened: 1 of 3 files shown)


test/rbs/runtime_prototype_test.rb
```diff
@@ -533,22 +533,25 @@ class TestTypeParams
...
 
-  class Unnamed
+  class Constants
     A = ARGF
     B = ENV
+    C = BasicObject.new
   end
 
-  def test_unnamed
+  def test_constants
```
ksss, line 542: I changed the name because the role has changed.

soutaro:
@ksss Thanks!

How about using `UnboundMethod#bind_call` to call `is_a?`, `class`, and other methods?
I'm not sure if it's completely safe, but calling `#is_a?` and `#class` with `self` of `BasicObject` works.

ksss:
Very excellent idea.

The same correspondence may be needed elsewhere, as in `Module#name`.

ksss:
Fixed conflict resolution and use of UnboundMethod.

soutaro (approved):
🙏

PR #1486, opened 2023-09-01
Should keep escape of alias and attr family

ksss:
Passing `RBS::Writer` for the aliases and attr family will cause the backquote to be off and output an RBS that results in a syntax error.

I fixed `RBS::Writer` so that backquotes are kept.


soutaro (approved):
👏

PR #1487, opened 2023-09-01
Clean up the incorrect untyped errors

sampersand:
I erroneously used `| (untyped) -> ...` previously as an "anything but the previous arguments" variant, but in reality it can be "anything, including the previous arguments"—so the new return values reflect the uncertainty of that

### DIFF
--- a/core/false_class.rbs
+++ b/core/false_class.rbs
@@ -25,7 +25,7 @@ class FalseClass
   # statements.
   #
   def ===: (false) -> true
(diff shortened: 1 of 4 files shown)


soutaro (approved):
🙆‍♂️

soutaro:
Some of (at least) the incorrect signatures are introduced before your PR. The reviewer (== @soutaro) should have caught and fixed them...

PR #1488, opened 2023-09-01
Updated previous changes to use the new type aliases

sampersand:
This updates the previous commits I've made to use the new type aliases


soutaro (approved):
💪

PR #1490, opened 2023-09-03
[prototype runtime] Const name should be absolute

ksss:
For example, given the following class, the result of `rbs prototype runtime` will be invalid.

```rb
$ cat t.rb
require 'active_support/all'

class Foo
  T = 1.day
end

$ rbs prototype runtime -R t.rb Foo
class Foo
  T: Duration #=> Could not find ::Duration (RBS::NoTypeFoundError)
end
```

To fix this problem, I modified the constant type to be absolute path.
If we want to keep the expression minimal, we will need to consider context, etc.


soutaro (approved):
👍

PR #1495, opened 2023-09-04
[prototype runtime] Optimize performance

ksss:
### Problem

When the number of modules/classes is large and the path is deep, there is a problem that the performance of `rbs prototype runtime` is extremely degraded.
This is due to the loop of calling `const_name` for all modules to find their names each time the paths are nested.
This situation can easily occur in Rails applications and is worth addressing.

### Solution

To suppress this loop, I modified `@modules` to be `Hash` to suppress the search cost.

### Comparison Code

```rb
$ cat t.rb
module Root
  10.times do |i|
    const_set("A_#{i}", Module.new{
      10.times do |j|
        const_set("B_#{i}_#{j}", Module.new{
          10.times do |k|
            const_set("C_#{i}_#{j}_#{k}", Module.new{
              10.times do |l|
                const_set("D_#{i}_#{j}_#{k}_#{l}", Class.new)
              end
            })
          end
        })
      end
    })
  end
end

$ time bundle exec rbs prototype runtime -R t.rb "Root*" > /dev/null

# before
101.24s user 1.69s system 99% cpu 1:43.11 total

# after
1.05s user 0.14s system 89% cpu 1.330 total
```

### DIFF
--- a/lib/rbs/prototype/runtime.rb
+++ b/lib/rbs/prototype/runtime.rb
@@ -14,7 +14,7 @@ module RBS
       def initialize(patterns:, env:, merge:, owners_included: [])
         @patterns = patterns
         @decls = nil
-        @modules = []
+        @modules = {}
         @env = env
         @merge = merge
         @owners_included = owners_included.map do |name|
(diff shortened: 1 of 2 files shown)


soutaro (approved):
🚄

PR #1496, opened 2023-09-04
Add documentation of `require: false` for rbs collection

pocke:
`rbs collection` ignores gems specified `require: false` in Gemfile. But this behavior is not documented. Therefore this PR adds the documentation.


soutaro (approved):
✏️

PR #1500, opened 2023-09-05
Update FileTest's definitions

sampersand:
This updates the signatures for `FileTest`, and updates the unit tests accordingly.

The current signatures are actually pretty accurate, they are just missing the implicit conversions.

More concretely the following changes were made to `FileTest`'s signatures:
- `blockdev?`, `chardev?`, `directory?`, `exist?`, `file?`, `grpowned?`, `owned?`, `pipe?`, `setgid?`, `setuid?`, `size?`, `size?`, `socket?`, `sticky?`, `world_readable?`, `world_writable?`, and `zero?`: Converted from `(String | IO) -> ...` to `(path | io) -> ...`.
- `empty?`: Changed to be an alias of `zero?`.
- `executable?`, `executable_real?`, `readable?`, `readable_real?`, `symlink?`, `writable?`, and `writable_real?`: Converted from `(String) -> ...` to `(path) -> ...`.
- `identical?`: Changed from `(String | IO, String | IO) -> ...` to `(path | io, path | io) -> ...`.

### DIFF
--- a/core/file_test.rbs
+++ b/core/file_test.rbs
@@ -13,7 +13,7 @@ module FileTest
   #     File.blockdev?('/dev/sda1')       # => true
   #     File.blockdev?(File.new('t.tmp')) # => false
   #
-  def self?.blockdev?: (String | IO file_name) -> bool
+  def self?.blockdev?: (path | io file_name) -> bool
 
   # <!--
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks! 🎉

PR #1501, opened 2023-09-05
Better record type format

soutaro:
Use compact form if a number is included in the attribute name.

### DIFF
--- a/lib/rbs/types.rb
+++ b/lib/rbs/types.rb
@@ -478,7 +478,7 @@ module RBS
         return "{ }" if self.fields.empty?
 
         fields = self.fields.map do |key, type|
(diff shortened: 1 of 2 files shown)


lib/rbs/types.rb
```diff
@@ -478,7 +478,7 @@ def to_s(level = 0)
         return "{ }" if self.fields.empty?
 
         fields = self.fields.map do |key, type|
-          if key.is_a?(Symbol) && key.match?(/\A[A-Za-z_][A-Za-z_]*\z/)
+          if key.is_a?(Symbol) && key.match?(/\A[A-Za-z_][A-Za-z_0-9]*\z/)
```
ParadoxV5, line 481: o:
```suggestion
          if key.is_a?(Symbol) && key.match?(/\A[A-Za-z_]\w*\z/)
```

PR #1503, opened 2023-09-05
Proposal: Support auto extending module

tk0miya:
In Ruby language, developers need to use both "include" and "extend" to implement mix-in having instance methods and class methods.  The auto extending module is a well-known technique to realize it (ex. `ActiveSupport::Concern`, "extend" call in `included` block, and so on).

This supports the auto extending modules via modules having "autoextend:..." annotation.  The `RBS::DefinitionBuilder` searches the extended modules from the annotations of the included modules.


docs/syntax.md
```diff
@@ -780,3 +780,17 @@ _annotation_ ::= `%a{` _annotation-text_ `}`  # Annotation using {}
 
 _annotation-text_ ::= /[^\x00]*/              # Any characters except NUL (and parenthesis)
 ```
+
+#### Auto extending modules
+
+Module having "autoextend:..." annotation is considered as an auto extending module.
+When such auto extending modules are included, the including class will be extended by annotated modules.
```
tk0miya, line 787: We need to find a better name for the "auto extending module" and its annotation.

lib/rbs/definition_builder.rb
```diff
@@ -239,6 +239,19 @@ def build_singleton0(type_name)
             definition.class_variables.merge!(defn.class_variables)
           end
 
+          one_ancestors.each_included_module do |mod|
+            included_module = env.class_decls[mod.name] or raise "Unknown name for one_ancestors.included_module: #{type_name}"
+            included_module.decls.each do |decl|
+              decl.decl.annotations.each do |annotation|
+                if annotation.string.start_with? "autoextend:"
```
tk0miya, line 246: Should we support type arguments for modules? If so, this code should be updated more.

lib/rbs/definition_builder/ancestor_builder.rb
```diff
@@ -111,7 +111,7 @@ def self.singleton(type_name:, super_class:)
             params: nil,
             super_class: super_class,
             self_types: nil,
-            included_modules: nil,
+            included_modules: [],
```
tk0miya, line 114: I'm not sure this change does not give side effects to other processes.

lib/rbs/definition_builder/ancestor_builder.rb
```diff
@@ -111,7 +111,7 @@ def self.singleton(type_name:, super_class:)
...
             included_interfaces: nil,
```
tk0miya, line 115: I think there is no need for auto-extending interfaces. So I kept this as is.

test/rbs/ancestor_graph_test.rb
```diff
@@ -83,7 +83,7 @@ module M
           graph.each_ancestor(InstanceNode("::M")).to_set
         )
         assert_equal(
-          Set[InstanceNode("::B"), InstanceNode("::C")],
+          Set[InstanceNode("::B"), InstanceNode("::C"), SingletonNode("::B"), SingletonNode("::C")],
```
tk0miya, line 86: I think this is a side effect of changing `AncestorBuilder`. But I don't understand if this is a problem or not.

lib/rbs/definition_builder.rb
soutaro: Forget the generics for now.
It shouldn't happen frequently.

lib/rbs/definition_builder.rb
```diff
@@ -239,6 +239,19 @@ def build_singleton0(type_name)
...
+                  mod_name = TypeName("::" + annotation.string.split(":", 2)[1])
```
soutaro, line 247: Use `Resolver::TypeNameResolver` for name resolution.

lib/rbs/definition_builder.rb
```diff
@@ -239,6 +239,19 @@ def build_singleton0(type_name)
             definition.class_variables.merge!(defn.class_variables)
           end
 
+          one_ancestors.each_included_module do |mod|
```
soutaro, line 242: How about using `#instance_ancestors` explicitly, instead of having `included_modules` for _singleton_ ancestors?

lib/rbs/definition_builder.rb
```diff
@@ -239,6 +239,19 @@ def build_singleton0(type_name)
...
+            included_module = env.class_decls[mod.name] or raise "Unknown name for one_ancestors.included_module: #{type_name}"
+            included_module.decls.each do |decl|
+              decl.decl.annotations.each do |annotation|
+                if annotation.string.start_with? "autoextend:"
+                  mod_name = TypeName("::" + annotation.string.split(":", 2)[1])
+                  subst = tapp_subst(mod_name, [])
+                  define_instance(definition, mod_name, subst)
```
soutaro, line 249: We should dedup the explicitly extended modules and automatically extended modules.

Issue #1475, opened 2023-08-28
Many Kernel methods are defined as Object methods


ksss:
I think these should be `Kernel` methods, not `Object`.

```rb
require 'rbs'

env = RBS::Environment.from_loader(RBS::EnvironmentLoader.new).resolve_type_names
defined_kernel_methods = env.class_decls[TypeName("::Kernel")].primary.decl.members.grep(RBS::AST::Members::MethodDefinition).map(&:name)
defined_object_methods = env.class_decls[TypeName("::Object")].primary.decl.members.grep(RBS::AST::Members::MethodDefinition).map(&:name)
((Kernel.instance_methods.sort - defined_kernel_methods) & defined_object_methods).each do |sym|
  puts "- [ ] #{sym}"
end
```

- [ ] !~
- [ ] <=>
- [ ] ===
- [x] class
    - already defined
- [ ] clone
- [ ] define_singleton_method
- [ ] display
- [ ] dup
- [ ] enum_for
- [ ] eql?
- [ ] extend
- [ ] freeze
- [ ] frozen?
- [ ] hash
- [ ] inspect
- [ ] instance_of?
- [ ] instance_variable_defined?
- [ ] instance_variable_get
- [ ] instance_variable_set
- [ ] instance_variables
- [ ] is_a?
- [ ] itself
- [ ] method
- [ ] methods
- [ ] nil?
- [ ] object_id
- [ ] private_methods
- [ ] protected_methods
- [ ] public_method
- [ ] public_methods
- [ ] public_send
- [ ] remove_instance_variable
- [ ] respond_to?
- [ ] send
- [ ] singleton_class
- [ ] singleton_method
- [ ] singleton_methods
- [ ] tap
- [ ] to_s
- [ ] yield_self

ParadoxV5:
That’s a Ruby implementation problem ¯\\\_(ツ)_/¯
```ruby
Object.instance_methods(false) #=> []
```
Everything is actually all defined in Kernel, with Object merely BasicObject + Kernel.
The API docs put instance methods (as opposed to utility functions like `print`) on the Object page for our convenience (façade). The RBSs just followed suit.

ksss:
I found this PR. https://github.com/ruby/rbs/pull/86

@soutaro 
Do you still think the same way?
I am concerned about the meaning of RBS as a document.

soutaro:
+1 for moving those object to `::Kernel`. (But when?)

[...]

PR #1504, opened 2023-09-05
Move methods from Object to Kernel

ksss:
Fix https://github.com/ruby/rbs/issues/1475

- Move methods to Kernel
    - Keep to use `Object::name`
- Move testing code to Kernel
    - Remove and merge duplicated testing
    - Drop Object_test
- Fix cli test


soutaro:
I believe all of the changes by #1499 are picked but open another one if something is left...

PR #1507, opened 2023-09-08
Updated binding to use modern kernel tests

sampersand:
This updates `Binding`'s tests to use the modern version of unit tests.

Also, it updates `Binding#clone` and `Binding#dup` to return `instance`, not `self`.

### DIFF
--- a/core/binding.rbs
+++ b/core/binding.rbs
@@ -30,9 +30,9 @@
 # Binding objects have no class-specific methods.
 #
 class Binding
-  def clone: () -> instance
+  def clone: () -> self
 
-  def dup: () -> instance
+  def dup: () -> self
 
   # <!--
   #   rdoc-file=proc.c
--- a/core/binding.rbs
+++ b/core/binding.rbs
@@ -30,9 +30,9 @@
 # Binding objects have no class-specific methods.
 #
 class Binding
-  def clone: () -> self
+  def clone: () -> instance
 
-  def dup: () -> self
+  def dup: () -> instance
 
   # <!--
   #   rdoc-file=proc.c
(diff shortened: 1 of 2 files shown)


core/binding.rbs
```diff
@@ -30,9 +30,9 @@
 # Binding objects have no class-specific methods.
 #
 class Binding
-  def clone: () -> self
+  def clone: () -> instance
```
soutaro, line 33: Do you have any specific reason to use `instance` for those methods?
I think using `self` is enough and better here. It always returns a `Binding` object.

test/stdlib/Binding_test.rb
```diff
@@ -1,53 +1,72 @@
...
-class BindingTest < StdlibTest
-  target Binding
+class BindingInstanceTest < Test::Unit::TestCase
+  include TypeAssertions
+
+  testing 'Binding'
 
   def test_clone
-    binding.clone
+    assert_send_type  '() -> instance',
```
soutaro, line 9: It's better to test it with `Binding` instead of `instance` (or `self`).

core/binding.rbs
sampersand: I was thinking `self` means "the exact same object" (i.e. with `def foo: () -> self`, `self.foo.equal?(self)` wouldalways be true), whereas `instance` means an instance of `Binding`.

soutaro: `self` is the exactly same type of the receiver.
It's different from `instance` if the type is generic. `instance` type fills the generic parameters by `untyped`, so `Array[String]` will be `Array[untyped]`.

They are identical because `Binding` is not generic, but I think `self` is more preferable.

PR #1509, opened 2023-09-08
Updated ThreadGroup

sampersand:
This updates the signatures for `ThreadGroup`, and adds tests for it.

More specifically, the following changes were made.
- Removed redundant `< Object` from `ThreadGroup`
- Moved `Default` inside `ThreadGroup`
- `ThreadGroup#add` : return `self` instead of `ThreadGroup`
- `ThreadGroup#list`: removed `::` from `::Array[Thread]`

### DIFF
--- a/core/thread_group.rbs
+++ b/core/thread_group.rbs
@@ -8,7 +8,13 @@
 # Newly created threads belong to the same group as the thread from which they
 # were created.
 #
-class ThreadGroup < Object
+class ThreadGroup
+  # <!-- rdoc-file=thread.c -->
+  # The default ThreadGroup created when Ruby starts; all Threads belong to it by
+  # default.
+  #
+  Default: ThreadGroup
+
   # <!--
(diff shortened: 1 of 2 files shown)


core/thread_group.rbs
```diff
@@ -8,7 +8,13 @@
 # Newly created threads belong to the same group as the thread from which they
 # were created.
 #
-class ThreadGroup < Object
+class ThreadGroup
+  # <!-- rdoc-file=thread.c -->
+  # The default ThreadGroup created when Ruby starts; all Threads belong to it by
+  # default.
+  #
+  Default: instance
```
soutaro, line 16: Please use `ThreadGroup` instead of `instance`.
(I need to prohibit writing `instance` type in constants, type aliases, and anything it's not clear.)

soutaro (approved):
Thanks! 🙏

Issue #1471, opened 2023-08-27
Allow Type Variable constraints to be arbitrary types


ParadoxV5:
Currently, the `RHS` of `[T < RHS]` must be an interface, class or module, at least what RubyMine IDE identifies. Wynaut allow it to be anything? For the minumum, it can self-document when arg and return are the same constrained object/type: `def []=: [E < Elem?] (index name_or_position, E) -> E`

*Originally posted in https://github.com/ruby/rbs/pull/1456#discussion_r1306579971; c.c. @sampersand*

soutaro:
+1 for this.

PR #1510, opened 2023-09-08
Updated Encoding, its Errors, and its tests

sampersand:
This updates the signatures and tests of `Encoding` and its errors (`Encoding::CompatibilityError`, `Encoding::ConverterNotFoundError`, `Encoding::InvalidByteSequenceError`, and `Encoding::UndefinedConversionError`).

Notably, this does _not_ update `Encoding::Converter`, which will be a separate PR.

More specifically, this PR updates the following signatures: 
- `Encoding._load`: Added, but without documentation (as it doesn't have any).
- `Encoding.locale_charmap`: Added
- `Encoding.default_external=`, and `Encoding.default_internal=`: Updated to match `_ToStr`. This signature could be improved when #1471 merges.
- `Encoding.find`: Use `encoding` instead of `String | Encoding`.
- `Encoding.aliases`, `Encoding.list`, `Encoding.name_list`, `Encoding#names`: Remove leading `::` from signatures
- `Encoding#replicate`: **REMOVED ENTIRELY**. See https://github.com/ruby/ruby/pull/7079 for details.
- `Encoding#to_s`: Now an alias of `Encoding.name`
- All the `Encoding`s and errors are now underneath `Encoding` instead of at toplevel.

It makes the other following changes:
- Moves `Encoding_InvalidByteSequenceError_test.rb` into `Encoding_test.rb`
- Moves `Encoding_UndefinedConversionError_test.rb.rb` into `Encoding_test.rb`
- Creates `Encoding_CompatibilityErrorInstanceTest` and `Encoding_ConverterNotFoundErrorInstanceTest` test classes.
- Updates `Encoding_test.rb` to use the modern testing suite, testing the new new version of signatures.


soutaro (approved):
Looks good! Fixing the CI errors.

PR #1511, opened 2023-09-10
Updated UnboundMethod, added Method::param_types

sampersand:
This updates the type definitions for `UnboundMethod`, as well as introduces the `Method::param_types` alias.

More specifically, this:
- Updates `UnboundMethod_test.rb` to use the new testing harness, and conform to new signatures.
- Adds in `Method::param_types`, and changes `Method#parameters` to return it
- `UnboundMethod#eql?`: added as an alias for `UnboundMethod#==` 
- `UnboundMethod#hash`: added in 
- `UnboundMethod#clone`: Now returns `instance` not `self`
- `UnboundMethod#to_s`: added as an alias for `UnboundMethod#inspect`
- `UnboundMethod#owner`: The return type is now `(Class | Module)` instead of `Module`. While it's technically redundant, it's not common knowledge (especially amount new rubyists) that `Class`es are `Module`s, so this is a bit clearer IMO.
- `UnboundMethod#parameters`: Now returns `Method::param_types`
- `UnboundMethod#bind_call`: Added in `**untyped`; while technically redundant, I like making it explicit that kwargs are also allowed.

### DIFF
--- a/core/unbound_method.rbs
+++ b/core/unbound_method.rbs
@@ -60,7 +60,29 @@ class UnboundMethod
   #     Array.instance_method(:sum) == Enumerable.instance_method(:sum)
   #     #=> false, Array redefines the method for efficiency
   #
-  def ==: (untyped) -> bool
+  def ==: (untyped other) -> bool
+
+  # <!-- rdoc-file=proc.c -->
+  # Two unbound method objects are equal if they refer to the same method
+  # definition.
+  #
+  #     Array.instance_method(:each_slice) == Enumerable.instance_method(:each_slice)
+  #     #=> true
+  #
+  #     Array.instance_method(:sum) == Enumerable.instance_method(:sum)
+  #     #=> false, Array redefines the method for efficiency
+  #
+  alias eql? ==
+
+  # <!--
+  #   rdoc-file=proc.c
+  #   - meth.hash   -> integer
+  # -->
+  # Returns a hash value corresponding to the method object.
+  #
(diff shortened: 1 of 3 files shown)


core/unbound_method.rbs
```diff
@@ -213,7 +263,7 @@ class UnboundMethod
   #
   #     (1..3).method(:map).owner #=> Enumerable
   #
-  def owner: () -> Module
+  def owner: () -> (Class | Module)
```
soutaro, line 266: Do we need `Class` here? `Class` is a subclass of `Module`.

core/unbound_method.rbs
```diff
@@ -78,7 +100,7 @@ class UnboundMethod
   #     m.call # => "bar"
   #     n = m.clone.call # => "bar"
   #
-  def clone: () -> self
+  def clone: () -> instance
```
soutaro, line 103: `self`?

sampersand: I was thinking `self` means "the exact same object" (i.e. with `def foo: () -> self`, `self.foo.equal?(self)` wouldalways be true), whereas `instance` means an instance of `Binding`.

core/unbound_method.rbs
sampersand: Not technically, but I thought it was a bit clearer, as most new rubyists don't know `Class` is an instance of `Module`.

soutaro (approved):
👍

PR #1512, opened 2023-09-10
Updated Signal

sampersand:
This updates signatures for `Signal` as well as added `Signal_test.rb`.

More specifically, this makes the following changes:
- All three methods in `Signal` are now `self?`, as they're `module_function`s.
- `Signal.list`: Removed leading `::` in `::Hash[String, Integer]`
- `Signal.signame`: Use `int` instead of `Integer`  for `signo`
- `Signal.trap`: The signal is now `Integer | interned`, and use `Signal::trap_command` for the `command` type and return type.
- Added `Signal::trap_command`, which is `string | bool | nil | _TrapCommand` (where `_TrapCommand` is an interface with just `def call: (Integer signal) -> void`
   - Technically, `trap` can accept `Integer`s and `Float`s as well, but that's just an implementation bug (And, in fact, can cause a segfault if ruby is compiled without optimizations.)


core/signal.rbs
```diff
@@ -89,6 +89,12 @@ module Signal
   #     Child died
   #     Terminating: 27460
   #
-  def self.trap: (Integer | String | Symbol signal, ?untyped command) -> (String | Proc)
-               | (Integer | String | Symbol signal) { (Integer arg0) -> untyped } -> (String | Proc)
+  def self?.trap: (Integer | interned signal) { (Integer) -> void } -> trap_command
```
soutaro, line 92: Does it always return `Proc`?

sampersand: It returns whatever the previously registered `trap` command was.

soutaro (approved):
👍

PR #1516, opened 2023-09-11
core: Add Enumerator.produce

tk0miya:
Add the signature for Enumerator.produce
https://bugs.ruby-lang.org/issues/14781

### DIFF
--- a/core/enumerator.rbs
+++ b/core/enumerator.rbs
@@ -158,7 +158,8 @@ class Enumerator[unchecked out Elem, out Return] < Object
   #     Enumerator.produce { scanner.scan(PATTERN) }.slice_after { scanner.eos? }.first
   #     # => ["7", "+", "38", "/", "6"]
   #
-  def self.produce: (?untyped initial) { (untyped prev) -> untyped } -> Enumerator[untyped, bot]
+  def self.produce: [T] () { (T? prev) -> T } -> Enumerator[T, nil]
+                  | [T] (T initial) { (T prev) -> T } -> Enumerator[T, nil]
 
   # <!--
   #   rdoc-file=enumerator.c
--- a/core/enumerator.rbs
+++ b/core/enumerator.rbs
@@ -126,6 +126,42 @@ class Enumerator[unchecked out Elem, out Return] < Object
 
   # <!--
   #   rdoc-file=enumerator.c
+  #   - Enumerator.produce(initial = nil) { |prev| block } -> enumerator
+  # -->
+  # Creates an infinite enumerator from any block, just called over and over.  The
+  # result of the previous iteration is passed to the next one. If `initial` is
+  # provided, it is passed to the first iteration, and becomes the first element
+  # of the enumerator; if it is not provided, the first iteration receives `nil`,
+  # and its result becomes the first element of the iterator.
+  #
+  # Raising StopIteration from the block stops an iteration.
+  #
+  #     Enumerator.produce(1, &:succ)   # => enumerator of 1, 2, 3, 4, ....
+  #
+  #     Enumerator.produce { rand(10) } # => infinite random number sequence
+  #
+  #     ancestors = Enumerator.produce(node) { |prev| node = prev.parent or raise StopIteration }
+  #     enclosing_section = ancestors.find { |n| n.type == :section }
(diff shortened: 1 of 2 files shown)


core/enumerator.rbs
```diff
@@ -124,6 +124,43 @@
 class Enumerator[unchecked out Elem, out Return] < Object
   include Enumerable[Elem]
 
+  # <!--
```
tk0miya, line 127: I copied this document from enumerator.c. Is that alright?
https://github.com/ruby/ruby/blob/master/enumerator.c#L3052-L3086

core/enumerator.rbs
```diff
@@ -124,6 +124,43 @@
...
+  #   require "date"
+  #   Enumerator.produce(Date.today, &:succ).detect(&:tuesday?)
+  #
+  #   # Simple lexer:
+  #   require "strscan"
+  #   scanner = StringScanner.new("7+38/6")
+  #   PATTERN = %r{\d+|[-/+*]}
+  #   Enumerator.produce { scanner.scan(PATTERN) }.slice_after { scanner.eos? }.first
+  #   # => ["7", "+", "38", "/", "6"]
+  def self.produce: (?untyped initial) { (untyped prev) -> untyped } -> self
```
tk0miya, line 162: At first, I thought it would be better to use a type parameter for this method (ex. `[T] (T initial) { (T prev) -> T } -> self[T, untyped]`).
But developers can use the different types between `initial` and `prev` (the return type of the block). So it's difficult to give a better type to this method to me. Therefore, I used `untyped` for this.

core/enumerator.rbs
tk0miya: I updated this comment by the result of CI (`rake annotate`?): https://github.com/ruby/rbs/actions/runs/6149970207/job/16686945264?pr=1516

test/stdlib/Enumerator_test.rb
```diff
@@ -25,6 +25,13 @@ class EnumeratorSingletonTest < Test::Unit::TestCase
 
   testing "singleton(::Enumerator)"
 
+  def test_produce
+    assert_send_type(
+      "(Integer initial) { (Integer) -> Integer } -> Enumerator[Integer, untyped]",
+      Enumerator, :produce, 1, &:succ
+    )
+  end
```
tk0miya, line 33: Sorry, my PR is broken now. By mistake, I ran `rake test` on my local (not `rake stdlib_test`)... I'll check what happens tomorrow.

tk0miya: Fixed on my local. But CI has still been broken by SEGV. I guess it's not related to this change.

core/enumerator.rbs
```diff
@@ -124,6 +124,42 @@
...
+  #     Enumerator.produce(Date.today, &:succ).detect(&:tuesday?)
+  #
+  #     # Simple lexer:
+  #     require "strscan"
+  #     scanner = StringScanner.new("7+38/6")
+  #     PATTERN = %r{\d+|[-/+*]}
+  #     Enumerator.produce { scanner.scan(PATTERN) }.slice_after { scanner.eos? }.first
+  #     # => ["7", "+", "38", "/", "6"]
+  #
+  def self.produce: (?untyped initial) { (untyped prev) -> untyped } -> Enumerator[untyped, bot]
```
ParadoxV5, line 161: > At first, I thought it would be better to use a type parameter for this method (ex. `[T] (T initial) { (T prev) -> T } -> self[T, untyped]`). But developers can use the different types between `initial` and `prev` (the return type of the block). So it's difficult to give a better type to this method to me. Therefore, I used `untyped` for this.
> ⸺ https://github.com/ruby/rbs/pull/1516#discussion_r1321888385

`[T]` would be the union type of `initial`, `prev` and the block return.
```suggestion
  def self.produce: [T] () { (T? prev) -> T } -> Enumerator[T, nil]
                  | [T] (T initial) { (T prev) -> T } -> Enumerator[T, nil]
```
(seems that `Enumerator.produce(…).each {}` is always `nil`)

soutaro (approved):
Thank you, @tk0miya and @ParadoxV5 ! 🎉

PR #1518, opened 2023-09-15
add `Kernel#respond_to_missing?`

ParadoxV5:
https://github.com/ruby/rbs/blob/fce99b9700f5b9c9749e7a8c65852dc4b6b40cc7/test/test_helper.rb#L112-L113

My current project overrides `respond_to_missing?` and fails Steep on the courtesy `super` calls.

This originally based on #1504 because they conflict (this’d be `Object#respond_to_missing?` otherwise). Since the prerequisite is now (forced-pushed and) merged, this PR now bases on `master` itself.

### DIFF
--- a/test/stdlib/Kernel_test.rb
+++ b/test/stdlib/Kernel_test.rb
@@ -941,4 +941,15 @@ class KernelInstanceTest < Test::Unit::TestCase
       Object.instance_method(:to_s)
     )
   end
+  
+  def test_respond_to_missing?
+    obj = Object.new
+    
+    # The default implementation always returns `false` regardless of the args,
+    # let alone their types; though overrides only have to support Symbol + bool
+    assert_send_type(
+      "(::Symbol, bool) -> bool",
+      obj, :respond_to_missing?, :to_s, true
+    )
+  end
 end
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -2806,6 +2806,8 @@ module Kernel : BasicObject
   #
   def respond_to?: (interned name, ?boolish include_all) -> bool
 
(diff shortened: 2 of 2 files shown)


test/stdlib/Kernel_test.rb
```diff
@@ -888,6 +888,17 @@ def test_respond_to?
...
   end
 
+  def test_respond_to_missing?
+    obj = Object.new
+    
+    # The default implementation always returns `false` regardless of the args,
+    # let alone their types; though overrides only have to support Symbol + bool
+    assert_send_type(
+      "(Symbol, bool) -> bool",
+      obj, :respond_to_missing?, :to_s, true
```
ParadoxV5, line 898: I see other `(bool)` tests simply use `bool` for the type and `true` for the arg, instead of separate tests for true and false. I suppose I’ll match.

test/stdlib/Kernel_test.rb
```diff
@@ -888,6 +888,17 @@ def test_respond_to?
     obj.respond_to?('to_s', true)
```
ParadoxV5, line 888: It’s quite odd dropping a modern assertion among a sea of old-style tests ._.

ParadoxV5 (commented):
I wrote my first RBS test! 😄

test/stdlib/Kernel_test.rb
ParadoxV5: … and yeah indeed odd, the CI rejects my new-style assertions and I’m too novice to know why.

ParadoxV5: … nvmd, figured it out.

soutaro (approved):
Thanks!

PR #1520, opened 2023-09-15
Add `__todo__` type

soutaro:
`__todo__` is a type that is equivalent to `untyped`, but can be used to mark something should be done later.

### DIFF
--- a/ext/rbs_extension/lexer.h
+++ b/ext/rbs_extension/lexer.h
@@ -60,7 +60,7 @@ enum TokenType {
   kVOID,            /* void */
   kUSE,             /* use */
   kAS,              /* as */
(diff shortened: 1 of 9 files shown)


ext/rbs_extension/lexer.h
```diff
@@ -60,6 +60,7 @@ enum TokenType {
   kVOID,            /* void */
   kUSE,             /* use */
   kAS,              /* as */
+  k__TODO__,         /* __todo__ */
```
sampersand, line 63: Indentation is slightly off

soutaro:
```suggestion
  k__TODO__,        /* __todo__ */
```

PR #1524, opened 2023-09-19
Update Marshal

sampersand:
This updates `Marshal`'s definition, as well as its tests.

More specifically this has the following changes:
- `Marshal::MAJOR_VERSION`: Moved definition inside the module
- `Marshal::MINOR_VERSION`: Moved definition inside the module
- `Marshal.dump`: It's now a `self?.`, `port` is now a `_Writer` (and is the return value), and`limit` is now `int?` not `Integer`.
- `Marshal.load`: `source` is now `string | _Source`, and the `proc` field is no longer a literal proc (but rather a `interface _Proc[A]; def call: (untyped) -> A end`)


soutaro (approved):
💯

PR #1529, opened 2023-09-19
Update Exception

sampersand:
This updates the `Exception` type definitions and its associated unit tests.

More specifically, this:
- Removes `< Object` from `Exception`
- `Exception.exception`: Now accepts `string | _ToS` (instead of just `String`) and returns `instance`.
- `Exception#==`: Updated parameter name to `obj` instead of `arg0`.
- `Exception#backtrace` and `Exception#backtrace_locations`: Removed leading `::`
- `Exception#detailed_message`: `highlight` is now `bool?` instead of `boolish`, and the `**untyped` param is now named.
- `Exception#exception`: added `?self` to the no-argument case; `message` (renamed from `arg0`) now accepts `string | _ToS` and returns `instance`
- `Exception#initialize`: `arg0` renamed to `message` and now accepts `string | _ToS`. Also returns `self`.
- `Exception#set_backtrace`: Removed leading `::`s, renamed `arg0` to `backtrace`
- `Exception#full_message`: `highlight` and `order` now accept `nil`, `order` now also accepts `string` (which, should be `"top"` or `"bottom"`.


soutaro (approved):
👍

PR #1530, opened 2023-09-20
Updated GC signatures

sampersand:
This pr updates GC's signatures, unit tests, and adds in the `Writer` test type (which implements `_Writer`).

Two methods, `GC.latest_gc_info` and `GC.stat` technically can accept any hash, and just add their respective keys/values (e.g. `[K,V] (Hash[K, V]) -> Hash[K | Symbol, V | Integer]`). However, this is currently not testable, and the added complication of supporting it isn't worth the very niche—but technically correct—signature.

In addition to removing the occasional leading `::`, this has the following changes:
- `GC::compact_info` type alias was added for the return value of `GC.compact` and other methods
- `GC::{INTERNAL_CONSTANTS,OPTS}`: Moved inside the class
- `GC.start`: Removed duplicate argument names.
- `GC.stat`: Gave actual names to arguments, Hash variant is now nilable.
- `GC.stress`: `bool` not `TrueClass | FalseClass`
- `GC.stress=`: Split `Integer` and `bool` cases apart to fix return value.
- `GC.verify_compaction_references`: Added optional `toward`, `double_heap` and `expand_heap` keyword arguments
- `GC.latest_gc_info`: `nil` is now accepted, removed `K` from the `Hash[K | Symbol, unty]ed` function.
- `GC.garbage_collect`: Removed duplicate argument names.
- `GC::Profiler` was moved inside the class, and `auto_compact`-related functions were added.

### DIFF
--- a/core/gc.rbs
+++ b/core/gc.rbs
@@ -7,6 +7,151 @@
 # You may obtain information about the operation of the GC through GC::Profiler.
 #
 module GC
+  # <!-- rdoc-file=gc.c -->
+  # The GC profiler provides access to information on GC runs including time,
+  # length and object space size.
+  #
+  # Example:
+  #
+  #     GC::Profiler.enable
+  #
+  #     require 'rdoc/rdoc'
+  #
+  #     GC::Profiler.report
+  #
+  #     GC::Profiler.disable
+  #
+  # See also GC.count, GC.malloc_allocated_size and GC.malloc_allocations
+  #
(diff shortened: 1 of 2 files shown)


sampersand:
Oh lovely, ruby segfault: https://github.com/ruby/rbs/actions/runs/6256735966/job/16988021332?pr=1530

soutaro (approved):
👍

PR #1533, opened 2023-09-20
Updated Math

sampersand:
This updates `Math`, and converts its unit tests into the new version.

The only contentious thing in the update is the name of the `Numeric & _ToF` type, within `Math` which I named `double` (as that's what the source code refers to them as). I considered alternatives such as `number` or `numeric_float`, but I wasn't a fan of either. `double` is a bit undescriptive though, so suggestions welcome.

More specifically, the following changes were made:
- Tests in `Math_test.rb` were overhauled into the new version. A `refute` case to each was added to ensure that the `Numeric` bound was required.
- `Math::PI`, `Math::E`, and `Math::DomainError`: Moved within `Math`
- `Math::double`: Added (equiv to `Numeric & _ToF`)
- `Math::{a?(cosh?|sinh?|tanh?),cbrt,erfc?,frexp,gamma,log(10|2),sqrt}`: Changed the `Numeric` argument to `double`
- `Math::{atan2,hypot,log}`: Changed the `Numeric` arguments to `double`s.
- `Math::ldexp`: Change the `fraction` to `double` and the `exponent` to `int`
- `Math::lgamma`: Changed the `Integer` return value to `-1 | 1`.

### DIFF
--- a/core/math.rbs
+++ b/core/math.rbs
@@ -127,6 +127,37 @@
 # *   ::hypot: Returns `sqrt(a**2 + b**2)` for the given `a` and `b`.
 #
 module Math
+  # <!-- rdoc-file=math.c -->
+  # Definition of the mathematical constant E for Euler's number (e) as a Float
+  # number.
+  #
+  E: Float
+
+  # <!-- rdoc-file=math.c -->
+  # Definition of the mathematical constant PI as a Float number.
+  #
+  PI: Float
+
+  # <!-- rdoc-file=math.c -->
+  # Raised when a mathematical function is evaluated outside of its domain of
+  # definition.
+  #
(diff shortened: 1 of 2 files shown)


core/math.rbs
```diff
@@ -127,6 +127,37 @@
...
+  # *produces:*
+  #
+  #     Math::DomainError: Numerical argument is out of domain - "acos"
+  #
+  class DomainError < StandardError
+  end
+
+  # A type that's passable to `Math` functions: A `Numeric` type that defines `.to_f`.
+  #
+  type double = Numeric & _ToF
```
soutaro, line 159: 👍

soutaro (approved):
🎉

PR #1539, opened 2023-09-22
[prototype runtime --todo] Consider accessibility mismatch

ksss:
For example, the current `Kernel#initialize_copy` is defined as a public method in RBS, but it's actually a private method. This fact was discovered through the implementation of this PR.
Such accessibility inconsistencies should also be detected as part of the "todo" items.


soutaro (approved):
👍

PR #1540, opened 2023-09-22
Add private instance methods of Kernel `#initialize_*`

ksss:
- Fix visibility for `Kernel#initialize_copy`
- Add `Kernel#initialize_clone`
- Add `Kernel#initialize_dup`

All of these should be defined as private instance methods within `Kernel`.

Note: These methods were discovered by https://github.com/ruby/rbs/pull/1539


soutaro (approved):
👍

PR #1541, opened 2023-09-22
Remove all non-required methods from testing types

sampersand:
# What
This PR fixes all the non-JSON "test helper types" (`ToStr`, `ToInt`, etc.) by removing all non-essential methods (i.e. everything but `__id__`, `__send__` and whatever the type defines itself are undefined). It also updates a few signatures and tests to work with this new scheme.

# Why
The point of an interface is to denote the _bare minimum_ constraints for a type: If a function accepts a `_ToStr` type, the only thing it should do with that is call `.to_str`. If it also calls `.inspect`, then it should also have an `& _Inspect` constraint.

The problem is that our test types currently inherit from `Object`, which has a lot of extra chafe defined on it (`.method`, `.instance_exec`, `.tap`, etc.)—things that _aren't_ described by the interface. This means that if a function that only accepts a `_ToStr` type _relies_ on one of these "`Object`/`Kernel` methods," then our signature is incorrect. However, we have no way of testing that right now, as all types inherit from `Object`.

By inheriting from `BasicObjet` (and `undef`ing every non-essential method), we can now guarantee that if a test with our `ToStr` type passes, we know that only `.to_str` was accessed, and not the chafe like `.respond_to?` or friends.

## Caveat
One small caveat: By not defining `.is_a?` or `.inspect`, the testing harness actually fails because it's unable to check types/print output messages. The solution to this is the `Guaranteed` module, which supplies methods that're guaranteed to work for any type, including these "blank slates" from the standard library.

# `JSON` (and `Gem::Version`)
This "Blank Slate" scheme works well for C-originated methods (which is where most of the implicit conversion methods come from anyways). The problem is that for Ruby-originated methods, more often than not they're implicitly assuming `Object` descendants. The solution to this is to slap on `& Object` to their constraints. While we could enumerate exactly all the methods that're used, it gets a bit unwieldy. For example, `Gem::Version.create` would be:
```rbs
    def self.create: ((_ToS & _NilP & _Hash  & _Eql & _RespondTo) | Version input) -> instance
                   | (nil input) -> nil
```
Not exactly an easy read. `_ToS & Object | Version` seems better.

### DIFF
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -170,6 +170,10 @@ interface _Inspect
   def inspect: () -> String
 end
 
+interface _IsA
+  def is_a?: (Class cls) -> boolish
+end
+
 interface _Each[out A]
   def each: () { (A) -> void } -> void
 end
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -170,10 +170,6 @@ interface _Inspect
   def inspect: () -> String
 end
 
-interface _IsA
-  def is_a?: (Class cls) -> boolish
-end
-
 interface _Each[out A]
   def each: () { (A) -> void } -> void
 end
--- a/core/rubygems/version.rbs
+++ b/core/rubygems/version.rbs
@@ -167,7 +167,7 @@ module Gem
     # -->
     # True if the `version` string matches RubyGems' requirements.
     #
-    def self.correct?: (_ToS version) -> bool
+    def self.correct?: (_ToS & Object version) -> bool
 
     # <!--
     #   rdoc-file=lib/rubygems/version.rb
@@ -180,7 +180,7 @@ module Gem
     #     ver2 = Version.create(ver1)       # -> (ver1)
     #     ver3 = Version.create(nil)        # -> nil
     #
-    def self.create: (_ToS | Version input) -> instance
+    def self.create: (_ToS & Object | Version input) -> instance
                    | (nil input) -> nil
 
     # <!--
@@ -190,7 +190,7 @@ module Gem
     # Constructs a Version from the `version` string.  A version string is a series
     # of digits or ASCII letters separated by dots.
     #
-    def initialize: (_ToS version) -> void
+    def initialize: (_ToS & Object version) -> void
 
     # <!--
     #   rdoc-file=lib/rubygems/version.rb
--- a/core/rubygems/version.rbs
+++ b/core/rubygems/version.rbs
@@ -167,7 +167,7 @@ module Gem
     # -->
     # True if the `version` string matches RubyGems' requirements.
     #
-    def self.correct?: (_ToS & Kernel version) -> bool
+    def self.correct?: (_ToS version) -> bool
 
     # <!--
     #   rdoc-file=lib/rubygems/version.rb
@@ -180,7 +180,7 @@ module Gem
     #     ver2 = Version.create(ver1)       # -> (ver1)
     #     ver3 = Version.create(nil)        # -> nil
     #
-    def self.create: (_ToS & Kernel | Version input) -> instance
+    def self.create: (_ToS | Version input) -> instance
                    | (nil input) -> nil
 
     # <!--
@@ -190,7 +190,7 @@ module Gem
     # Constructs a Version from the `version` string.  A version string is a series
     # of digits or ASCII letters separated by dots.
     #
-    def initialize: (_ToS & Kernel version) -> void
+    def initialize: (_ToS version) -> void
 
     # <!--
     #   rdoc-file=lib/rubygems/version.rb
--- a/core/rubygems/version.rbs
+++ b/core/rubygems/version.rbs
@@ -167,7 +167,7 @@ module Gem
     # -->
     # True if the `version` string matches RubyGems' requirements.
     #
-    def self.correct?: (_ToS & Object version) -> bool
+    def self.correct?: (_ToS & Kernel version) -> bool
 
     # <!--
     #   rdoc-file=lib/rubygems/version.rb
@@ -180,7 +180,7 @@ module Gem
     #     ver2 = Version.create(ver1)       # -> (ver1)
     #     ver3 = Version.create(nil)        # -> nil
     #
-    def self.create: (_ToS & Object | Version input) -> instance
+    def self.create: (_ToS & Kernel | Version input) -> instance
                    | (nil input) -> nil
 
     # <!--
@@ -190,7 +190,7 @@ module Gem
     # Constructs a Version from the `version` string.  A version string is a series
     # of digits or ASCII letters separated by dots.
     #
-    def initialize: (_ToS & Object version) -> void
+    def initialize: (_ToS & Kernel version) -> void
 
     # <!--
     #   rdoc-file=lib/rubygems/version.rb
--- a/lib/rbs/test.rb
+++ b/lib/rbs/test.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "securerandom"
+require "rbs/test/guaranteed"
 require "rbs/test/observer"
 require "rbs/test/spy"
 require "rbs/test/errors"
@@ -23,6 +24,8 @@ module RBS
     METHODS = Kernel.instance_method(:methods)
 
     class ArgumentsReturn
+      include Guaranteed::Inspect
+
       attr_reader :arguments
       attr_reader :exit_value
       attr_reader :exit_type
--- a/lib/rbs/test/guaranteed.rb
+++ b/lib/rbs/test/guaranteed.rb
@@ -1,0 +1,36 @@
+module RBS
+  module Test
+    module Guaranteed
+      IS_AP = ::Kernel.instance_method(:is_a?)
+      private_constant :IS_AP
+
+      def self.value_is_a?(value, kind)
+        IS_AP.bind_call(value, kind)
+      end
+
+      module Inspect
+        EQUAL = ::BasicObject.instance_method(:equal?)
+        INSPECT = ::Kernel.instance_method(:inspect)
+        private_constant :EQUAL, :INSPECT
+
+        module_function def guaranteed_inspect(obj)
+          obj.inspect
+        rescue NoMethodError => err
+          raise unless err.name == :inspect && EQUAL.bind_call(obj, err.receiver)
+          GUARANTEED_INSPECT.bind_call(obj)
+        end
+
+        def inspect
+          string = "<#{self.class.name}:"
+
+          instance_variables.each_with_index do |variable, index|
+            string.concat ', ' unless index.zero?
+            string.concat "#{variable}: #{guaranteed_inspect(variable)}"
+          end
+
+          string.concat '>'
+        end
+      end
+    end
+  end
+end
--- a/lib/rbs/test/guaranteed.rb
+++ b/lib/rbs/test/guaranteed.rb
@@ -12,7 +12,7 @@ module RBS
           obj.inspect
         rescue NoMethodError => err
           raise unless err.name == :inspect && EQUAL.bind_call(obj, err.receiver)
-          GUARANTEED_INSPECT.bind_call(obj)
+          INSPECT.bind_call(obj)
         end
 
         def inspect
--- a/lib/rbs/test/guaranteed.rb
+++ b/lib/rbs/test/guaranteed.rb
@@ -1,13 +1,6 @@
 module RBS
   module Test
     module Guaranteed
-      IS_AP = ::Kernel.instance_method(:is_a?)
-      private_constant :IS_AP
-
-      def self.value_is_a?(value, kind)
-        IS_AP.bind_call(value, kind)
-      end
-
       module Inspect
         EQUAL = ::BasicObject.instance_method(:equal?)
         INSPECT = ::Kernel.instance_method(:inspect)
--- a/lib/rbs/test/guaranteed.rb
+++ b/lib/rbs/test/guaranteed.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module RBS
   module Test
     module Guaranteed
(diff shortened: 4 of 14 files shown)


test/stdlib/Warning_test.rb
sampersand, line 1: I want to start moving away from explicitly mentioning the test types, so file has changes converting from `ToSym` into `_ToSym`.

core/builtin.rbs
```diff
@@ -170,6 +170,10 @@ interface _Inspect
   def inspect: () -> String
 end
 
+interface _IsA
```
sampersand, line 173: Needed for `Pathname#relative_path_from`, and I think would be common enough that putting it in `builtin.rbs` is warranted. I could be persuaded though and put it in `Pathname`

core/rubygems/version.rbs
```diff
@@ -167,7 +167,7 @@ module Gem
     # -->
     # True if the `version` string matches RubyGems' requirements.
     #
-    def self.correct?: (_ToS version) -> bool
+    def self.correct?: (_ToS & Object version) -> bool
```
sampersand, line 170: Technically `Object` also has a `.to_s` defined on it. But I think having `_ToS & Object` makes it clear that the `version` object should have a custom `.to_s` defined.

lib/rbs/test.rb
```diff
@@ -23,6 +24,8 @@ module Test
     METHODS = Kernel.instance_method(:methods)
 
     class ArgumentsReturn
+      include Guaranteed::Inspect
```
sampersand, line 27: Without this, you can't call `.inspect` on `ArgumentsReturn` of test types

lib/rbs/test/guaranteed.rb
```diff
@@ -0,0 +1,36 @@
...
+        private_constant :EQUAL, :INSPECT
+
+        module_function def guaranteed_inspect(obj)
+          obj.inspect
+        rescue NoMethodError => err
+          raise unless err.name == :inspect && EQUAL.bind_call(obj, err.receiver)
+          GUARANTEED_INSPECT.bind_call(obj)
+        end
+
+        def inspect
```
sampersand, line 23: Custom inspect; We could update it to mamke it prettier, but I think this is good enough for now.

stdlib/pathname/0/pathname.rbs
```diff
@@ -1084,7 +1084,7 @@ class Pathname
   # Note that this method does not handle situations where the case sensitivity of
   # the filesystem in use differs from the operating system default.
   #
-  def relative_path_from: (Pathname | string base_directory) -> Pathname
+  def relative_path_from: (Pathname | (string & _IsA) base_directory) -> Pathname
```
sampersand, line 1087: Interestingly enough, this is the only function (outside of `JSON` and `Gem::Version`) that needed updating.

test/stdlib/Warning_test.rb
sampersand, line 1: I don't like referencing the test types in unit tests, I prefer to use the interfaces. This is part of my gradual plan to update them over time

test/stdlib/json/JSON_test.rb
```diff
@@ -1,6 +1,27 @@
 require_relative "../test_helper"
 require "json"
 
+
+class JsonToStr
```
sampersand, line 5: the JSON library uses a lot of `Object` methods, so it's a lot easier to just make new, custom `.to_str` and `.to_s` classes. When I get around to `JSON` i'll rework a lot of the logic.

sampersand: Defined these two classes in here instead of `test_helper`. In the future the rest of the json test helper classes should be moved over, and more generally helper classes should be local to their files IMO

core/rubygems/version.rbs
ParadoxV5: type alias `str` for `String | _ToS` /halfjoke

anyways, would it be
```suggestion
    def self.correct?: (_ToS & Kernel version) -> bool
```
rather?

lib/rbs/test.rb
ParadoxV5:
```suggestion
      include _Inspect
```
(yes you can do this)

lib/rbs/test/guaranteed.rb
ParadoxV5:
```ruby
irb(main):001> BasicObject.new
=> #<BasicObject:0x00007f63cfad6260>
irb(main):002> _.inspect
(irb):2:in `<main>': undefined method `inspect' for #<BasicObject:0x00007f63cfad6260> (NoMethodError)
        from /usr/local/lib/ruby/gems/3.2.0/gems/irb-1.8.1/exe/irb:9:in `<top (required)>'
        from /usr/local/bin/irb:25:in `load'
        from /usr/local/bin/irb:25:in `<main>'
```
I wonder how does IRB do it

Can also try binding `Kernel#inspect` on it

P.S. that could be how IRB does it
```
irb(main):001> class B < BasicObject; attr_accessor :b end
=> [:b, :b=]
irb(main):002> b = B.new
=> #<B:0x00007f18d3b123d8>
irb(main):003> b.b = 1
=> 1
irb(main):004> b
=> #<B:0x00007f18d3b123d8 @b=1>
irb(main):005> b.b = b
An error occurred when inspecting the object: #<NoMethodError: undefined method `inspect' for #<B:0x00007f18d3b123d8>>
An error occurred when running Kernel#inspect: #<NoMethodError: undefined method `inspect' for #<B:0x00007f18d3b123d8>>
…
```

lib/rbs/test/guaranteed.rb
```diff
@@ -0,0 +1,36 @@
+module RBS
+  module Test
+    module Guaranteed
```
ParadoxV5, line 3: Can we have a more sounding name?

ParadoxV5 (commented):
Could there be an object that isn’t a `BasicObject`? 🤔

lib/rbs/test/guaranteed.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        def inspect
+          string = "<#{self.class.name}:"
+
+          instance_variables.each_with_index do |variable, index|
+            string.concat ', ' unless index.zero?
+            string.concat "#{variable}: #{guaranteed_inspect(variable)}"
+          end
+
+          string.concat '>'
```
ParadoxV5, line 31: Been in RBS land for too long, resigning Ruby roots?
```suggestion
          "<#{self.class} #{instance_variables.map {|variable| "#{variable}: #{
            guaranteed_inspect(instance_variable_get(variable))
          }" }.join(", ")}>"
```
Although yea it’s kinda long

Though I still feel that there’s a more direct approach to the problem…

lib/rbs/test/type_check.rb
```diff
@@ -136,7 +136,7 @@ def zip_keyword_args(hash, fun)
       end
 
       def keyword?(value)
-        value.is_a?(Hash) && value.keys.all? {|key| key.is_a?(Symbol) }
+        Guaranteed.value_is_a?(value, Hash) && value.keys.all? {|key| Guaranteed.value_is_a?(key, Symbol) }
```
ParadoxV5, line 139:
```ruby
BasicObject.new.is_a? BasicObject #=! undefined method `is_a?' (NoMethodError)
BasicObject === BasicObject.new #=> true
```

```suggestion
        Hash === value and value.each_key.all?(Symbol)
```

lib/rbs/test/guaranteed.rb
sampersand: wdym more sounding name?

core/rubygems/version.rbs
sampersand: Yeah that's probably better.

lib/rbs/test/guaranteed.rb
ParadoxV5: > Although yea it’s kinda long

Attempt 2 (didn’t follow any particular code style)
```suggestion
12345678901234567890123456789012345678901234567890123456789012345678901234567890
          i_var_list = instance_variables.map do |var|
            "#{var}: #{guaranteed_inspect(instance_variable_get(var))}"
           end.join(', ')
          "<#{self.class} #{i_var_list}>"
```

ParadoxV5: > Though I still feel that there’s a more direct approach to the problem…

Kinda wann just
```ruby
class BlankSlate
  def inspect …
```
right before trying to inspect the `ArgumentReturn`

lib/rbs/test/type_check.rb
```diff
@@ -136,7 +136,7 @@ def zip_keyword_args(hash, fun)
       end
 
       def keyword?(value)
-        value.is_a?(Hash) && value.keys.all? {|key| key.is_a?(Symbol) }
+        Hash === value and value.each_key.all?(Symbol)
```
soutaro, line 139: Please use `&&` here, because `===` is not a control flow operator.

soutaro:
Thank you for opening this PR @sampersand,

I understand the problem of `ToS`, `ToI`, and others inheriting from `Object`. 👍 for inheriting from `BasicObject`.
Thinking of if intersection types with `& Kernel` is necessary or not.

stdlib/json/0/json.rbs
```diff
@@ -969,7 +969,7 @@ module JSON
   #        #<Admin:0x00000000064c41f8
   #        @attributes={"type"=>"Admin", "password"=>"0wn3d"}>}
   #
-  def self?.load: (string | _JsonReadableIO | _JsonRead source, ?Proc proc, ?json_options options) -> untyped
+  def self?.load: ((string | _JsonReadableIO | _JsonRead) & Kernel source, ?Proc proc, ?json_options options) -> untyped
```
soutaro, line 972: Is the `& Kernel` for `respond_to?`?

https://github.com/ruby/ruby/blob/master/ext/json/lib/json/common.rb#L537-L552

stdlib/pathname/0/pathname.rbs
soutaro: https://github.com/ruby/ruby/blob/9ee58b2054c1bbe722ae5a2a4ec6a750ee583220/ext/pathname/lib/pathname.rb#L515-L551

core/rubygems/version.rbs
```diff
@@ -167,7 +167,7 @@ module Gem
     # -->
     # True if the `version` string matches RubyGems' requirements.
     #
-    def self.correct?: (_ToS version) -> bool
+    def self.correct?: (_ToS & Kernel version) -> bool
```
soutaro, line 170: https://github.com/ruby/ruby/blob/9ee58b2054c1bbe722ae5a2a4ec6a750ee583220/lib/rubygems/version.rb#L174

`nil?` is used.

core/rubygems/version.rbs
```diff
@@ -180,7 +180,7 @@ module Gem
     #     ver2 = Version.create(ver1)       # -> (ver1)
     #     ver3 = Version.create(nil)        # -> nil
     #
-    def self.create: (_ToS | Version input) -> instance
+    def self.create: (_ToS & Kernel | Version input) -> instance
```
soutaro, line 183: `nil?` https://github.com/ruby/ruby/blob/9ee58b2054c1bbe722ae5a2a4ec6a750ee583220/lib/rubygems/version.rb#L190

core/rubygems/version.rbs
```diff
@@ -190,7 +190,7 @@ module Gem
     # Constructs a Version from the `version` string.  A version string is a series
     # of digits or ASCII letters separated by dots.
     #
-    def initialize: (_ToS version) -> void
+    def initialize: (_ToS & Kernel version) -> void
```
soutaro, line 193: `nil?` and `is_a?` (https://github.com/ruby/ruby/blob/9ee58b2054c1bbe722ae5a2a4ec6a750ee583220/lib/rubygems/version.rb#L221)

soutaro:
@sampersand 

Reviewed the `& Kernel` and `& _IsA` usages. My suggestion is to keep them without the annotations.

They are needed because `respond_to?`, `nil?`, and `is_a?`, and it seems like explicitly having the types is too much, because they are defined in most of the Ruby objects. (We can easily remove `nil?` and `is_a?` uses from the library code, but not sure if we should do it.) So, having the intersection type only makes sense if we pass a class inheriting from `BasicObject`, which would rarely happen. I don't think it's worth the extra type complexity.

For testing, I guess we can add some utility to import some of the methods from `Kernel`, like:

```
def test_correct?
  assert_send_type  "(String) -> bool",
                    Gem::Version, :correct?, "1.2.3"
  assert_send_type  "(_ToS) -> bool",
                    Gem::Version, :correct?, ToS.new(with_kernel_methods: :nil?)
end
```

To avoid having unintended successful method calls and the hidden requirements are explicit.

sampersand:
@soutaro I'm not entirely sure how we should keep those. Take `Pathname#relative_path_from` for example:
```ruby
class Pathname
  # ... 
  def relative_path_from(base_directory)
    base_directory = Pathname.new(base_directory) unless base_directory.is_a? Pathname
    # ... 
  end
  # ... 
end
```

How do you think we should annotate this? `base_directory` must respond to `.is_a?`. So an annotation of `def relative_path_from: (base_directory: _ToStr) -> ...` wouldn't be enough, as an object that only responds to `.to_str` wouldn't be permitted.

soutaro:
@sampersand I mean just ignore the case. Most Ruby objects has `is_a?` method because it's defined in `Object` (technically `Kernel`).

ParadoxV5:
> I mean just ignore the case. Most Ruby objects has `is_a?` method because it's defined in `Object` (technically `Kernel`).

Many rubyists are simply unaware that `BasicObject` is a thing (us three are among the few who know about it), and I suspect that there’re almost no Ruby projects that actively keep BasicObject support in mind (CRuby is one of the few that do, so is us RBS est. this PR). In summary, when someöne mentions their *say* `#to_str` support, chances are that they actually mean `_ToStr & Kernel`.

ParadoxV5:
> `BasicObject` support

* `self.is_a? mod` :arrow_right: `mod === self`
* `self.nil?` :arrow_right: `nil.equal? self`
* etc.

PR #1545, opened 2023-09-27
Add pp in stdlib

ksss:
I added the type for the standard library `pp` since it was missing.

I am performing the following tasks also:

- Modifying the test code generation template.
- Fix `Kernel#pp`.
    - ref: https://github.com/ruby/rbs/pull/1342
- Fix `RBS::Types::Optional#to_s` with Proc


ksss:
It seems there might be an issue with the `RBS::Writer`.

```diff
-    def seplist: (untyped list, ?(^() -> void)? sep, ?interned iter_method) { (*untyped, **untyped) -> void } -> void
+    def seplist: (untyped list, ?^() -> void? sep, ?interned iter_method) { (*untyped, **untyped) -> void } -> void
```

core/kernel.rbs
```diff
@@ -1269,7 +1269,7 @@ module Kernel : BasicObject
   # pp returns argument(s).
   #
   def self?.pp: [T] (T arg0) -> T
-              | (untyped, untyped, *untyped) -> Array[untyped]
+              | (untyped, *untyped) -> Array[untyped]
```
soutaro, line 1272: The second `untyped` was actually to skip the first overload if one argument is given.
Not sure if removing it causes another type error or not. (It won't on Steep.)

soutaro (approved):
👍

core/kernel.rbs
ksss: Oh, I see, that's what it meant.

The original definition is as follows:

```rb
module Kernel
  def pp(*objs)
    objs.each {|obj|
      PP.pp(obj)
    }
    objs.size <= 1 ? objs.first : objs
  end
  module_function :pp
end
```

I became concerned because Steep reported the following:

```
The method parameter has different kind from the declaration `[T(353)(354)] (?untyped, ?untyped, *untyped) -> (T(353)(354) | ::Array[untyped] | nil)`(Ruby::DifferentMethodParameterKind)
```

soutaro: @ksss I mean the use-site. It depends on how the overloading is resolved. (Steep uses the first matching one, so passing one argument to `pp` resolves to the first overload anyways.)

ksss: OK. I'll revert this commit later.

soutaro:
@ksss I found that `pp` is automatically loaded today. Moving to `core` would make more sense?

ksss:
@soutaro Thank you. Certainly, that's true.
I believe `pp` should be treated the same way as `set`.

I will move it to the core soon.

ksss:
@soutaro 
Oh wait.

If we move `pp` to the core, then its dependency `pretty_print` must also be moved to the core.
Looking at https://github.com/ruby/ruby/blob/5c1b7633fcd9c7c43974b72bd36cab1f9e9bd186/prelude.rb, we can see that while `Set` is autoloaded, but `PP` and `PrettyPrint` are not.
This might be because `PP` and `PrettyPrint` are used less frequently than `Set`.

It seems that `pp` is not exactly treated the same as `set`.

Moreover, the signature for `pp` is already included in the core.
https://github.com/ruby/rbs/blob/32cd6c91f0b44e509e4bad8be61b0a130690d48c/core/kernel.rbs#L1262-L1272

I believe it might be better not to add `pp` to the core.

soutaro:
@ksss Got it. 

`#pp` (method) can be used without explicitly loading it. `PP` (the library) should be loaded manually. It looks reasonable.

Clicking the merge button. 👍

PR #1556, opened 2023-10-01
doc: Add architecture note for DefinitionBuilder and AncestorBuilder

tk0miya:
Originally, this note was posted to #types channel in the ruby-jp slack from @soutaro.


tk0miya:
The better version was added in #1697. withdrawing.

PR #1557, opened 2023-10-03
core: Update signature for Module.define_method

tk0miya:
Inside the block for `Module.define_method` is evaluated as a instance method.  So its self-types should be an instance of the Module.


soutaro (approved):
👍

PR #1558, opened 2023-10-05
[Collection] Simple colorize collection text like Bundler

ksss:
I propose adding color to the log outputs of the collection.
As the current logs generated by the collection aren't colored, there's a potential for misunderstanding whether a log is indicating an error or not.
It appears that the text is based on Bundler, so I've taken a cue from Bundler for the coloration. **"Installing"**, **"Updating"**, and the final **"It's done!"** are colored in green, while **"Using"** retains its current color. This makes the command's success more evident.

image


soutaro (approved):
Good to merge!
We may want a flag to disable colorization for testing, like `Rainbow.enabled = false`.

soutaro:
We may want to see `NO_COLOR`? https://no-color.org

ksss:
`NO_COLOR` looks good to me.

Bundler is using also.

https://github.com/rubygems/rubygems/blob/ed65279100234a17d65d71fe26de5083984ac5b8/bundler/lib/bundler/vendor/thor/lib/thor/shell/color.rb#L99-L109

I'll fix it.

ksss:
I have made a modification so that the color toggle can be switched with the environment variable `NO_COLOR`.

soutaro (approved):
🙏 💯

PR #1560, opened 2023-10-05
stdlib: Add types for net/smtp

tk0miya:


stdlib/net-smtp/0/net-smtp.rbs
```diff
@@ -0,0 +1,55 @@
+module Net
```
tk0miya, line 1: Is this a good place for the bundle gems? Please let me know if not good.

test/stdlib/Net_SMTP_test.rb
```diff
@@ -0,0 +1,14 @@
+require_relative "test_helper"
+require "net/smtp"
+
+class NetSingletonTest < Test::Unit::TestCase
+  include TypeAssertions
+
+  library "net-smtp"
+  testing "singleton(::Net::SMTP)"
+
+  def test_new
```
tk0miya, line 10: I can't find the way to add test cases for `Net::SMTP.start` and `Net::SMTP#send_message` because they have side effects. Is there a better way to do that?

soutaro: Let's skip testing the methods. Accessing to external SMTP server doesn't look very good and safe...

(We will migrate to running servers locally to run tests in the future though...)

PR #1566, opened 2023-10-13
Add another type syntax validation

soutaro:
We have additional hidden rules on type syntax:

* `self`, `class`, and `instance` are only valid in class/module context (or interface context for `self`)
* `void` is only allowed at the return type or a generic parameter

This PR is to add validations and let `rbs validate` print error messages on that.


ParadoxV5:
TODO: Explicitly note the restrictions in [`syntax.md`](https://github.com/ruby/rbs/blob/master/docs/syntax.md)

ParadoxV5:
> * `void` is only allowed at the return type or a generic parameter

So should `top`, which `void` is an alias of. E.g., neither `top` nor `void` make sense as method args (see also soutaro/steep#918).

Conversely, `bot` only make sense as an arg type or a generic.

soutaro:
`top` and `bot` are different from `void`. `void` is a special type that only makes sense at return type. (with additional exception at generics parameter.)

PR #1568, opened 2023-10-15
Update MatchData

sampersand:
This PR updates `MatchData` and its associated tests, as well as adds `String::_Match`.

More specifically, this makes the following semantic updates:
- `MatchData::backref`: Added, equal to `String | Symbol | int`
- `MatchData#[]`: Updated to use `backref`, implicit `int`s, and `range[int?]`
- `MatchData#{begin,end,match,match_length,offset}`: Updated to use `backref`
- `MatchData#deconstruct_keys`: Named argument `keys`
- `MatchData#eql?`: Now an alias for `==`
- `MatchData#length`: Now an alias for `size`
- `MatchData#named_captures`: Added `symbolize_names` and annotation for >= 3.3.0
- `MatchData#values_at`: Now uses `backref` and `range[int?]`
- `MatchData#initialize_copy`: Takes an `instance` and returns `void`.

TODO: Add tests for `[]`, `values_at`, and `initialize_copy`.

### DIFF
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -229,7 +229,7 @@ type float = Float | _ToF
 
 # Represents a `Range[T]`, or a type that acts like it (via `.begin`, `.end`, and `.exclude_end?`).
 #
-type range[T] = Range[T] | _Range[T]
+type range[T] = Range[T & Range::_Spaceship] | _Range[T]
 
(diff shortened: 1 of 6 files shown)


sampersand:
Looks like this is already in master

PR #1573, opened 2023-10-17
Add `Abbrev` to runtime dependency

ima1zumi:
This PR adds `Abbrev` to runtime dependency to suppress the following Ruby 3.3's warning:

```console
$ ruby -ve 'require "rbs/cli"'
ruby 3.3.0preview2 (2023-09-14 master e50fcca9a7) [x86_64-darwin21]
/Users/mi/.asdf/installs/ruby/3.3.0-preview2/lib/ruby/gems/3.3.0+0/gems/rbs-3.2.2/lib/rbs/cli.rb:6: warning: abbrev which will be not part of the default gems since Ruby 3.4.0
```

ref: https://bugs.ruby-lang.org/issues/19351

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,6 +2,7 @@ PATH
   remote: .
   specs:
     rbs (3.2.0)
+      abbrev
 
 PATH
   remote: test/assets/test-gem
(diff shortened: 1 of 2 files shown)


soutaro (approved):
🙏

PR #1575, opened 2023-10-18
Add Kernel#`j`, `jj` and `JSON` with stdlib/json

ksss:
https://github.com/flori/json/blob/5c9d9f54bedc4c3711678816e126fdc603a68d19/lib/json/common.rb#L638-L672


soutaro (approved):
👍

PR #1577, opened 2023-10-19
Use Bundler.ui instead of ColoredIO

ksss:
The `ColoredIO` introduced in https://github.com/ruby/rbs/pull/1558 is based on `Bundler`'s implementation.
The goal was to aim for a minimal implementation independent of bundler, but I realized that rbs does require bundler.
I wonder if it would be more efficient to use `Bundler.ui` directly to avoid duplicate implementations and also utilize its rich features.

What do you think?

---

I think we should stop passing `stdout` as an argument.
It would be more consistent and simplify the code to always use `$stdout`, allowing it to be replaced from the outside, which also makes it easier to integrate with other tools like 'Bundler.ui'.
On the other hand, using it as an argument complicates the API and integration with other tools, making the code more complex.


pocke:
I am against this proposal because RBS is not Bundler. `Bundler.ui` looks like an internal API to implement Bundler's UI, so other programs should not use this API.

ksss:
Indeed, you are right.
It seems I was under the impression that Bundler.ui was a public API.

PR #1579, opened 2023-10-22
version.rb - update from 3.2.0 to 3.2.2

MSP-Greg:
This is causing the RBS gem to not be installed in Ruby master.  I believe [gems/bundled_gems](https://github.com/ruby/ruby/blob/f717faac632dd8664d0967f8e639b84d5d032854/gems/bundled_gems) in ruby/ruby will also need an update...

### DIFF
--- a/lib/rbs/version.rb
+++ b/lib/rbs/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module RBS
-  VERSION = "3.2.0"
+  VERSION = "3.2.2"
 end


MSP-Greg:
Windows ruby-loco builds and a local Ubuntu build log the following:
```
installing bundled gems:            /usr/local/lib/ruby/gems/3.3.0+0
                                    minitest 5.20.0
                                    power_assert 2.0.3
                                    rake 13.0.6
                                    test-unit 3.6.1
                                    rexml 3.2.6
                                    rss 0.3.0
                                    net-ftp 0.2.0
                                    net-imap 0.4.1
                                    net-pop 0.1.2
                                    net-smtp 0.4.0
                                    matrix 0.4.2
                                    prime 0.1.2
                                    typeprof 0.21.8
                                    debug 1.8.0
                                    racc 1.7.1
installing bundled gem cache:       /usr/local/lib/ruby/gems/3.3.0+0/cache
skipped bundled gems:
    rbs-3.2.0.gem                   not found in bundled_gems
```

soutaro (approved):
Thanks! 🙇‍♂️

ksss:
@MSP-Greg 
@soutaro 

Shouldn't we have made it `3.3.0` that the next release version?

When installing the `master` branch locally, I believe it can be confusing to determine whether `3.2.2` is the local version or the released version of `3.2.2`.

From my understanding, RBS branches just like CRuby. 
How about setting the version of the `master` branch to the next release miner version?
This way, when we install the `master` branch locally, we won't confuse it with the released version of `3.2.2`.

PR #1580, opened 2023-10-22
stdlib: Update type for URI.join

tk0miya:
URI.join takes not only strings, but also URI objects.  Therefore it would be better to update its signature.

Internally, arguments will be converted to URI objects before concatenating.
https://github.com/ruby/ruby/blob/v3_2_2/lib/uri/rfc3986_parser.rb#L107-L116


soutaro (approved):
👍

PR #1581, opened 2023-10-24
Add a guide about using `Data` and `Struct`

soutaro:


ParadoxV5:
Not everything in [`docs/`](https://github.com/ruby/rbs/tree/695890e6d2c034375286cfb2b36407d0a263a98e/docs) is indexed in [`README.md#guides`](https://github.com/ruby/rbs/blob/695890e6d2c034375286cfb2b36407d0a263a98e/README.md#guides). Intended?

https://github.com/ruby/rbs/blob/695890e6d2c034375286cfb2b36407d0a263a98e/README.md#L128-L135

soutaro:
It's not intentional, while I'm not very sure if listing everything here is good...

PR #1582, opened 2023-10-24
core: Add _Top interface to support inverting boolish

tk0miya:
To support inverting boolish values, adds a new interface `_Top` that provides `!` operator.

refs: https://github.com/soutaro/steep/issues/939

### DIFF
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -275,3 +275,9 @@ type interned = Symbol | string
 # Any Ruby object can have `boolish` type.
 #
 type boolish = top
+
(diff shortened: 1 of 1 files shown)


soutaro:
The behavior is only for Steep (at least for now), so I didn't plan to add the interface to RBS...
I'm thinking if it makes more sense.

tk0miya:
Agreed. But I can't find a better place to add such a type in the Steep repo.
Is there any good place?

Anyway, this is not a good choice. So I'm withdrawing now.

PR #1586, opened 2023-10-29
Add comments in the case of Struct

yui-knk:


soutaro (approved):
👍

soutaro:
I also want to mention having `attr_reader` for `Struct` is fine when we are using the `Struct` classes immutable.

PR #1589, opened 2023-10-31
Colored diff

ksss:
I added color by default to the `diff` format of the `rbs diff` command.

```rbs
# t.rbs
class Foo
  def a: () -> void
  def c: () -> void
end
```

```rbs
# tt.rbs
class Foo
  def a: () -> void
  def b: () -> void
  def c: () -> void
end
```

image

image

# Refactoring

### Rename from `RBS::Diff::CLI` to `RBS::CLI::Diff`

Previously, we had a namespace for the `Diff` feature, but there were two more general `CLI` classes, making it confusing.
We've changed the name and placed the class for the `Diff` feature under the `CLI` class, renaming it to `CLI::Diff`.

### Rename from `RBS::Collection::ColoredIO` to `RBS::CLI::ColoredIO`

We wanted to use `ColoredIO` within the `CLI` layer, but using `Collection::ColoredIO` at the `CLI` layer felt odd.
Moving `ColoredIO` to the `CLI` layer still felt natural within the `Collection` layer, so we made the shift.


soutaro (approved):
👍

PR #1590, opened 2023-11-01
[prototype runtime] Fix error if direct inherited

ksss:
rbs: 3.3.0.pre.1

When trying to analyze the [faraday gem](https://github.com/lostisland/faraday) using `rbs prototype runtime`, an error occurs.
This is due to `StructGenerator` not considering patterns like `class Option < Struct`.
In this case, I made a modification to exclude it from the generator target.

repro

```
$ bundle exec rbs prototype runtime -r faraday Faraday::Options
```

before

```
bundler: failed to load command: rbs (/bundle/ruby/3.2.0/bin/rbs)
/bundle/ruby/3.2.0/gems/rbs-3.3.0.pre.1/lib/rbs/prototype/runtime/value_object_generator.rb:107:in `block in build_s_new': undefined method `keyword_init?' for Faraday::Options:Class (NoMethodError)

            if CAN_CALL_KEYWORD_INIT_P ? (@target_class.keyword_init? == false || @target_class.keyword_init? == nil) : true
                                                       ^^^^^^^^^^^^^^
```

after

```
module Faraday
  class Options < ::Struct[untyped]
  end
end
```

### DIFF
--- a/lib/rbs/prototype/runtime.rb
+++ b/lib/rbs/prototype/runtime.rb
@@ -493,9 +493,9 @@ module RBS
         end #: AST::Declarations::Class?
 
         unless decl
-          if mod < Struct
+          if StructGenerator.generatable?(mod)
             decl = StructGenerator.new(mod).build_decl
-          elsif RUBY_VERSION >= '3.2' && mod < Data
+          elsif DataGenerator.generatable?(mod)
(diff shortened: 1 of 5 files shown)


soutaro (approved):
Thanks! 👍

PR #1591, opened 2023-11-01
Support cases where `Module#constants` has been redefined

ksss:
rbs 3.3.0.pre.1

When analyzing the gem `rouge` with `prototype runtime`, an error occurs.
The cause is that the `.constants` method is overridden in `rouge`.

I fixed the issue by using the method of calling the `Module#constants` UnboundMethod.

repro

```
$ bundle exec rbs prototype runtime -r rouge Rouge::Lexers::Actionscript
```

before

```
/ruby/3.2.0/gems/rouge-4.1.1/lib/rouge/lexers/actionscript.rb:96:in `constants': wrong number of arguments (given 1, expected 0) (ArgumentError)
```

after

```
module Rouge
  module Lexers
    class Actionscript < ::Rouge::RegexLexer
      def self.builtins: () -> untyped

      def self.constants: () -> untyped

      def self.declarations: () -> untyped

      def self.keywords: () -> untyped

      def self.reserved: () -> untyped
    end
  end
end
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,6 @@
 require: rubocop-rubycw
 AllCops:
-  TargetRubyVersion: 2.7
+  TargetRubyVersion: 3.0
   DisabledByDefault: true
   Exclude:
     - 'vendor/bundle/**/*'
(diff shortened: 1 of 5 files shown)


soutaro (approved):
👍

PR #1601, opened 2023-11-08
[rbs diff] Resolve constants name

ksss:
`rbs diff` command compares methods with their definitions resolved, but for constants, the comparison is a simple decl.

In this PR, constants are now compared as references from the specified type name by name resolution.


soutaro (approved):
👍

PR #1602, opened 2023-11-08
[rbs diff] Load dependencies from manifest.yaml

ksss:
The `rbs diff` command selects the environment to compare, but since manifest.yml is not read, there are cases where name resolution is not possible.

Fixed to read manifest.yml.


soutaro (approved):
👍

PR #1612, opened 2023-11-10
Load Bundler lazily

pocke:
Loading Bundler at the top-level causes a problem on testing for ruby/ruby. So this PR makes it lazy.

It will also improve the performance of launching `rbs` command (except `rbs collection`). 

cc/ @hsbt


soutaro (approved):
👍

hsbt (approved):
Thanks ❤️

PR #1618, opened 2023-11-14
Should escape if param name include not simple-word

ksss:
Fixes a syntax error in RBS output by `rbs prototype runtime` when names that cannot be used as variable names are used as members of Struct/Data, so that they are escaped.
The root cause of this problem is that parameter names are not escaped in `Writer`, or are missed even if escaped.

Real example:
https://github.com/rails/rails/blob/5b45a58cc8ad1809ee2732aa994fdb421a09cd94/activerecord/lib/active_record/relation/predicate_builder/range_handler.rb#L6

### DIFF
--- a/lib/rbs/types.rb
+++ b/lib/rbs/types.rb
@@ -883,7 +883,11 @@ module RBS
 
         def to_s
           if name
-            "#{type} #{name}"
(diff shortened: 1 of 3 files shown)


soutaro (approved):
👍

PR #1620, opened 2023-11-15
Fix `IO#gets` type

kateinoigakukun:
- Can take a string as a `nil` separator
- Can take limit integer as a first argument without a separator.
- Can take an optional keyword argument `chomp`

### DIFF
--- a/core/io.rbs
+++ b/core/io.rbs
@@ -1160,7 +1160,8 @@ class IO < Object
   #     f.gets(chomp: true) # => nil
   #     f.close
   #
-  def gets: (?String sep, ?Integer limit) -> String?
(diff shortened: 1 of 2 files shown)


core/io.rbs
```diff
@@ -1160,7 +1160,8 @@ class IO < Object
   #     f.gets(chomp: true) # => nil
   #     f.close
   #
-  def gets: (?String sep, ?Integer limit) -> String?
+  def gets: (String? sep, ?Integer limit, ?chomp: boolish) -> String?
```
sampersand, line 1163: Slight nitpick: `gets` allows for implicit conversions, so we should be using `string?` and `int`.

soutaro (approved):
👏

PR #1621, opened 2023-11-16
Update Struct

sampersand:
This PR updates `Struct`'s definitions, as well as adds in the `Hash::_Key` interface (needed for `Struct#deconstruct_keys`), which will be used when I work on `Hash`.

There's one small caveat: There's no real way to annotate `<subclass of Struct>.new`. After talking it over with @soutaro, we've decided it's fine and users can just add their own `new`s type definitions if needed. e.g.:
```ruby
Point = Struct.new(:x, :y) do def to_s = "(#{x}, #{y})" end
```
would become
```rbs
class Point < Struct[Numeric]
  attr_reader x: Numeric

  attr_reader y: Numeric
  
  def initialize: (Numeric x, Numeric y) -> void
  
  def to_s: () ->  String
end
```

### DIFF
--- a/core/hash.rbs
+++ b/core/hash.rbs
@@ -499,6 +499,12 @@
 class Hash[unchecked out K, unchecked out V] < Object
   include Enumerable[[ K, V ]]
 
+  interface _Key
+    def hash: () -> Integer
+
+    def eql?: (untyped rhs) -> boolish
+  end
+
   # <!--
   #   rdoc-file=hash.c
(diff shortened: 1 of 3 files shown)


soutaro:
It looks like the `hash.rbs` update contains unexpected indentation?

soutaro (approved):
👍 

I will try pushing commits with docs generated by `bin/generate_docs` and see what happens.

Issue #1448, opened 2023-08-21
Feature proposal: `rbs diff` command


ksss:
# Problem

## Differential check of generator output

When an RBS is automatically generated by a generator, it is difficult to track exactly what changes have been made to the RBS because of its large size and the order in which it is generated.

- Update of the generated collection.
- Comparison between generated RBS.

## Review RBS

If the RBS included in the PR is a complex change, it is difficult for the reviewer to make a decision based on the PR alone, and for the reviewee to explain the change.

## View differences by gem version

If you are managing multiple versions of RBS with gem_rbs_collection, etc., you can check the differences by version by checking the RBS file directly or writing a small script.

# Feature proposal: `rbs diff` command

Implement and provide new commands such as the `rbs diff` command.

```
$ cat dir1/t.rbs
class Foo
  def bar: () -> void
  def self.baz: () -> (Integer | String)
  def qux: (untyped) -> untyped
end

$ cat dir2/t.rbs
module Bar
  def bar: () -> void
end

module Baz
  def baz: (Integer) -> Integer?
end

class Foo
  include Bar
  extend Baz
end
```

```
$ bundle exec rbs diff --format markdown --type-name Foo --with-defined-in --before dir1 --after dir2
| before | after |
| --- | --- |
| `(::Foo) def bar: () -> void` | `(::Bar) def bar: () -> void` |
| `(::Foo) def qux: (untyped) -> untyped` | `-` |
| `(::Foo) def self.baz: () -> (::Integer \| ::String)` | `(::Baz) def self.baz: (::Integer) -> ::Integer?` |
```

### **format**

Required.
Specify output format.

markdown:

| before | after |
| --- | --- |
| `(::Foo) def bar: () -> void` | `(::Bar) def bar: () -> void` |
| `(::Foo) def qux: (untyped) -> untyped` | `-` |
| `(::Foo) def self.baz: () -> (::Integer \| ::String)` | `(::Baz) def self.baz: (::Integer) -> ::Integer?` |

diff:

```diff
- (::Foo) def bar: () -> void
+ (::Bar) def bar: () -> void

- (::Foo) def qux: (untyped) -> untyped
+ -

- (::Foo) def self.baz: () -> (::Integer | ::String)
+ (::Baz) def self.baz: (::Integer) -> ::Integer?
```

### **type-name**

Required.
Specify only one type name.
Type definitions are vast, but the types we want to know about are often specific.

### **before**, **after**

Required.
Specify one or more directories or files.

[...]

PR #1623, opened 2023-11-17
Add `--detail` option to rbs diff

ksss:
The `rbs diff` command has been released, but the initially planned feature to display definition class information and accessibility information was dropped. In this PR, I've made it so that adding an option allows for displaying and comparing more detailed information.

proposal: https://github.com/ruby/rbs/issues/1448


soutaro (approved):
👍

PR #1638, opened 2023-11-22
Add in ::Ops interfaces

sampersand:
This adds in the `Ops` module, which contains definitions for all the overloadable Ruby operators, but three (`[]`, `[]=`, and `.()` are missing, and'll be added when https://github.com/ruby/rbs/issues/1639 gets resolved.)

I think it makes sense to add these, for multiple reasons:
- All the operators have fixed signatures (these in particular taking one or no additional arguments, no blocks, and always return one type.)
- There's a limited amount of operators, with no plans to add support for custom operators.
- Having a centralized place for operator names makes it easy for end-users to reference them without having to create their own
	- prevent proliferation of many slightly-incompatible signatures (eg assuming `==` always returns `bool` and not `boolish` sometimes, etc)
- While a few of the operators could have homes (`<=>` in `Comparable`, `!` in `BasicObject`, etc.) a vast majority of them don't have a clearly defined location (`+`, `<<`, `=~`, just to name a few).
	- Having them all in one location is a lot more consistent, and makes finding signatures a lot easier (what if you didn't know that `!` was defined in `BasicObject`?)
	
Ruby's intended for programmer happiness after all


ParadoxV5:
> a vast majority of them don't have a clearly defined location (`+`, `<<`, `=~`, just to name a few).

Most of which (e.g., not `+`: it’s used by `Enumerable#sum`) also don’t have a *practical* reason to be included, as in, will people actually use those to describe all the duck-typing implementation details as much as possible?

ParadoxV5:
> `module Ops`

There is no module `Ops` in Ruby. The namespace should be under say `RBS` instead. Or maybe leave them toplevel like what [`builtin.rbs`](https://github.com/ruby/rbs/blob/master/core/builtin.rbs) does.

ParadoxV5:
#### Tracker: What methods use which operators

* [x] +: Enumerable#sum
* [ ] -:
* [ ] *: Enumerable#product in MyActiveSupport?
* [ ] /:
* [ ] %:
* [ ] **:
* [x] =~: Object#!~
* [x] <=>: Comparable
* [ ] <<:
* [ ] >>:
* [ ] &:
* [ ] |:
* [ ] ^:
* [ ] ~:
* [ ] -@:
* [ ] +@:

> (what if you didn't know that `!` was defined in `BasicObject`?)

Folks would take the fact that these things are always defined for granted. Thus they’d happily write `Object` or `untyped` if not `BasicObject` or `top`. (Why `BasicObject` and `top` aren’t synonymous is perhaps a different topic for a different thread.)

* BasicObject
  * ==
  * !=
  * !
* Object
  * !~
  * ===
* Comparable
  * <
  * >
  * <=
  * >=

soutaro:
@sampersand 

Generally, I agree with @ParadoxV5. It's not very clear if we have practical reasons for these operators. We don't need a _standard_ set of those operators technically, because interface types are structural.

A few more thoughts:

1. Is it better to give more abstract names to the interfaces, like `_Plus` instead of `_Add`?
2. The surrounding module should be under `::RBS` module, instead of `::Ops`

PR #1641, opened 2023-11-23
Update `String`'s definitions

sampersand:
**IMPORTANT: The `gsub!` function doesn't have tests (but has been tested manually) because of a bug in the test harness**

Test harness updates:
- `TypeAssertions::RUBY_EXE` added `RUBY_EXE`, which is the path to the ruby interpreter—Ruby's the executable guaranteed to be installed on any system we're testing (eg for `Kernel#system`)
- `TypeAssertions#pass`: Added `pass`, which is used in a few functions to indicate a test passes without using the type-checking assertions directly.

`String` updates (bolds are significant changes imo):
- `String#{*,+,==,byteindex,center,concat,include?,{r,}index,{l,r}just,match{,?},{r,}partition,scan,sum,to_i,upto}`: Updated variable names
- `String#{capitalize{,!},chop!}`: Added a TODO to add a variant for when arbitrary constraints come out.
- `String#{bytes,chars,codepoints}`: Returns `self` in the block variant
- `String#{capitalize,downcase}!`: All variants now return `self?`
- `String#{grapheme_clusters,lines}`: Added the block variant
- `String#{clear,insert,prepend,replace,succ}`: Returns `self` now

- **`String::selector`: an alias for `string` used to indicate methods take a string whose contents have a special meaning**
- **`String#{count,{delete,squeeze,tr{,_s}}{,!}}`: Now uses `String::selector`, updated variable names**
- `String.try_convert`: Added `(String) -> String | (_ToStr) -> String | (untyped) -> nil`
- `String#initialize{,_copy}`: Moved to the top of the class; `initialize`'s return type was updated.
- `String#%`: The variants can be implicit
- `String#{+,-}@`: Return `self | instance` now
- `String#<<`: Separated variants out for the variable names
- `String#<=>`: Annotated the previously `untyped` variant, now `untyped` returns `nil`.
- `String#===`: Now an alias for `String#==`
- `String#=~`: Added the variant for when the `other` is not a `Regexp` but defines `=~`
- `String#[]{,=}`: Collapsed variants, added implicits, the regexp variants now uses `MatchData::capture`
- `String#byteslice`: Collapsed `Range` variants, and made implicit
- **`String#bytesplice`: Added updates from Ruby 3.3**
- **`String#casecmp{,?}`: Split out `string` vs `untyped` variants; `string` variant in `casecmp` variant now returns `(-1 | 0 | 1)?`**
- `String#chomp`: Separator can now be nil
- `String#chomp!`: Now returns `self?`, and `nil` separator always returns `nil`.
- `String#delete_{prefix,suffix}!`: Now returns `self?` not `String?`
- `String#each_{byte,char,codepoint,grapheme_cluster,line}`: `Enumerator` variant is now first, to be consistent with other core lib definitions.
- **`String#encode{,!}`: All keyword args are now exhaustively annotated. The technically correct `**untyped` is not supplied.**
- `String::_EncodeFallbackAref`: Added for use with `String#encode{,!}`
- `String#force_encoding`: Now uses `encoding`
- `String#gsub{,!}`: The `Hash` variant is now implicit, with a `_ToS` value; Enumerator is now above the block variant; **!! `gsub!` isn't testable because of a bug in the test harness !!**
- `String#next{,!}`: Now aliases for `String#succ{,!}`
- `String#scrub:` Added `?nil` to the block variant
- `String#setbyte`: Now returns second variable, which is now `_ToInt`.
- `String#slice!`: Updated variable names, uses `MatchData:capture`, added implicits.
- `String#split`: Added `nil` value for pattern
- `String#sub{,!}`: The `Hash` variant is now implicit, with a `_ToS` value; `self` is now returned in the block form
- `String#to_s`: Now returns `self | String`
- `String#to_str`: An alias for `String#to_s`
- `String#to_sym`: An alias for `String#intern`
- `String#unicode_normalize`: Can now return `self`, named the variable
- `String#unicode_normalize!`: Always returns `self`, named the variable
- `String#unpack`: Implcits, and added the block variant
- `String#unpack1`: Implcits, added `?offset: int` parameter

### DIFF
--- a/core/string.rbs
+++ b/core/string.rbs
@@ -629,6 +629,9 @@
 class String
   include Comparable
 
+  # A `selector` is a special type of string, used within methods like `String#tr`.
+  type selector = string
+
   # <!--
   #   rdoc-file=string.c
   #   - String.try_convert(object) -> object, new_string, or nil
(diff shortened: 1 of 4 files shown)


sampersand:
<todo, clean this up, collapse commits down>

core/match_data.rbs
```diff
@@ -45,6 +45,7 @@
 # See also "Special global variables" section in Regexp documentation.
 #
 class MatchData
+  type capture = String | Symbol | int
```
ParadoxV5, line 48: Can they accept `_ToStr`?
```suggestion
  type capture = interned | int
```

sampersand: Sadly, no :-( It goes through `match_backref_number` which is (comments mine):
```c

static int
match_backref_number(VALUE match, VALUE backref)
{
    const char *name;
    int num;

    struct re_registers *regs = RMATCH_REGS(match);
    VALUE regexp = RMATCH(match)->regexp;

    match_check(match);
    if (SYMBOL_P(backref)) { // if backref is a `Symbol` (and not a subclass):
	backref = rb_sym2str(backref); // convert the symbol to a string
    }
    else if (!RB_TYPE_P(backref, T_STRING)) { // if backref is not a `String`
	return NUM2INT(backref); // call `.to_int` on it
    }
    name = StringValueCStr(backref); // convert the String to a null-terminated string

    num = name_to_backref_number(regs, regexp, name, name + RSTRING_LEN(backref));

    if (num < 1) {
        name_to_backref_error(backref);
    }

    return num;
}
```

sampersand:
This PR is waiting for https://github.com/ruby/rbs/pull/1643 to be merged

core/string.rbs
```diff
@@ -765,7 +765,7 @@ class String
   #
   # Otherwise returns `self.dup`, which is not frozen.
   #
-  def +@: () -> (self | instance)
```
soutaro: `self | instance` type doesn't make much sense.
It will be a `String` instance or it's subclass, so `self` is good here.

core/string.rbs
```diff
@@ -798,8 +798,7 @@ class String
   #
   # Related: String#concat, which takes multiple arguments.
   #
-  def <<: (Integer codepoint) -> self
-        | (string other_string) -> self
+  def <<: (string | Integer str_or_codepoint) -> self
```
soutaro, line 801: Combine the two overloads for the case passing a value of `String | Integer`.
(This can be improved with future version of Steep, but currently it doesn't work for this.)

core/string.rbs
```diff
@@ -822,19 +821,9 @@ class String
   #     'foo' <=> 'FOO' # => 1
   #     'foo' <=> 1 # => nil
   #
-  def <=>: (string | _Spaceship[self] other) -> (-1 | 0 | 1)?
+  def <=>: (string) -> (-1 | 0 | 1)
```
soutaro, line 824: As we discussed at Discord?

core/string.rbs
```diff
@@ -3436,8 +3415,7 @@ class String
   #
   # Related: String#unicode_normalize!, String#unicode_normalized?.
   #
-  def unicode_normalize: (?:nfc | :nfd | :nfkc | :nfkd form) -> (self | String)
-  # NOTE: ^ technically can return `self` if `self.encoding` is ascii.
+  def unicode_normalize: (?:nfc | :nfd | :nfkc | :nfkd form) -> self
```
soutaro, line 3418: There may be a case that returns exact `String`, not `self`?

core/string.rbs
ParadoxV5: Did you mean? `instance`

PR #1642, opened 2023-11-24
Small bugfix to have guaranteed print out instance variables

sampersand:


soutaro (approved):
Oh no, I have implemented this myself in #1660...

Issue #1543, opened 2023-09-24
Feature proposal: Multiple error on `rbs validate`


ksss:
The `rbs validate` command only outputs a single error at a time. However, when there are multiple issues, we have to go through a cycle of resolving errors, running `rbs validate`, resolving errors again, and running `rbs validate` repeatedly.

This issue is especially prevalent when auto-generating RBS, and it slows down the improvement cycle for addressing problems.

I propose resolving this issue by enhancing the `rbs validate` command to provide as much error information as possible when executed.

### Expected behavior.

```
$ cat t.rbs
class Foo
  def foo: () -> voids # typo
  def bar: -> Bar      # no type found
  def bar: -> Bar      # duplicated method
end

$ bundle exec rbs -I t.rbs validate --silent
t.rbs:2:17...2:22: Could not find voids (RBS::NoTypeFoundError)

    def foo: () -> voids
                   ^^^^^

t.rbs:3:14...3:17: Could not find Bar (RBS::NoTypeFoundError)

    def bar: -> Bar
                ^^^

t.rbs:3:2...3:17: ::Foo#bar has duplicated definitions in t.rbs:4:2...4:17 (RBS::DuplicatedMethodDefinitionError)

    def bar: -> Bar
    ^^^^^^^^^^^^^^^

```

soutaro:
Sounds good. 👍

PR #1648, opened 2023-11-26
Implement multiple validate

ksss:
fix https://github.com/ruby/rbs/issues/1543

# Canges of API

## Add `RBS::CLI::Validate`

I consolidated the behavior for validation into a single class, making it easier to divide the code using methods and such.

## Add `RBS::WillSyntaxError`

By transitioning from printing strings to using error classes, I've made it possible to display parts of the code using `detailed_message`.

# Changes of CLI

## Report by logger

By changing from direct display to stdout to using a logger, I've made it possible to differentiate displays by adjusting log levels.

## `--silent` is deprecated

With the introduction of log levels, the `--silent` option has become redundant, so I have marked it for deprecation.

### no option

It will report as many validation errors as possible. Warnings and errors are separated by log levels.
If there is even one error, the process will end with `exit 1`.

### --fail-fast

If validation fails, it will be reported immediately and the process will end. This is similar to the traditional individual validation.

### --exit-error-on-syntax-error

It will report all errors, including syntax errors, but if there is even one syntax error, the process will terminate with `exit 1`.

### --fail-fast --exit-error-on-syntax-error

Even syntax errors, which would not normally be considered failures, will be treated as errors and the process will immediately terminate with `exit 1`.


soutaro (approved):
@ksss Thanks! Added two commits for minor fix.

PR #1650, opened 2023-11-26
Bugfix; `.respond_to` is no longer used in `Rbs::Test::Errors#inspect_`

sampersand:
This fixes a bug in the testing suite where `.respond_to?` is being used on `BlankSlate` subclass instances.


soutaro (approved):
👏

PR #1652, opened 2023-11-27
Should raise ParsingError instead of ArgumentError

ksss:
I noticed that there are cases where obvious syntax errors are being reported as ArgumentError, so I have corrected this issue.

actual

```
$ bundle exec rbs parse --method-type -e '(foo` untyped) -> void'
bundler: failed to load command: rbs (/Users/ksss/.rbenv/versions/3.1.3/lib/ruby/gems/3.1.0/bin/rbs)
/Users/ksss/src/github.com/ksss/rbs/lib/rbs/parser_aux.rb:12:in `_parse_method_type': negative string size (or size too big) (ArgumentError)
```

expect

```
$ bundle exec rbs parse --method-type -e '(foo` untyped) -> void'
-e:1:4...1:5: Syntax error: unexpected token for function parameter name, token=``` (tOPERATOR) (RBS::ParsingError)

  (foo` untyped) -> void
      ^
```

### DIFF
--- a/ext/rbs_extension/parser.c
+++ b/ext/rbs_extension/parser.c
@@ -276,6 +276,14 @@ static VALUE parse_function_param(parserstate *state) {
     param_range.start = type_range.start;
     param_range.end = name_range.end;
 
+    if (!is_keyword_token(state->current_token.type)) {
+      raise_syntax_error(
+        state,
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍

PR #1653, opened 2023-11-28
change Date.parse signature to match behavior

emirikol:
Date.parse returns a a value when called with no argument.  
When called with too many arguments the ArgumentError also states the expected number is (0..3)


sampersand:
Could you add a test to `test/stdlib/Date_test.rb` for this case?

sampersand:
Woah, I never knew this was a thing. TIL! Could you add a test to `test/stdlib/Date_test.rb` for this case?

Issue #1542, opened 2023-09-24
Proposal: `Module[I]`


ParadoxV5:
Currently, we model module/class themselves (as opposed to `instance`s) as `singleton(Klass)`, whereas the `Module`/`Class` classes can only represent the module/class of `untyped`.

I suggest combining those two by giving `Module` and `Class` a type variable representing their instances. For example, `Class[String]` (and `Module[String]` too to under covariance) is equivlaent to `singleton(String)`.
For backward compatibility, `singleton(I)` can be a alternate (read: legacy) syntax for `Module[I]`.

The immediate benefit is `Class[I]` RBS finally able to write `def new: (…) -> I` and `def allocate: () -> I` (rather than expecting type checkers to infer from `singleton(I)`).

Speaking of `singleton(I)` —
https://github.com/ruby/rbs/blob/6e5a2893b4c2782cc9e71f76e193e9592b0c1ebf/docs/syntax.md?plain=1#L9
`[I < Bound] … singleton(I)` is invalid. `[I < singleton(Bound)] -> I` works the limitation around, but what if `I` has other duties?
```rbs
# Modules can’t inherit Classes, so I’m stuck with this design.
class MyClassWrapper[I < MyObject] < MyComponent
  def initialize: (singleton(I) klass) -> void #FIXME: Class[I] when
  def customized_new: (*args) -> I
end
```

This is also [how Java does it (as early as their type args were born)](https://docs.oracle.com/javase/1.5.0/docs/api/java/lang/Class.html).

soutaro:
`Class[String]` cannot be equivalent to `singleton(String)`, because `singleton(String)` has singleton methods.

I'm not sure if adding a generic parameter to `Class` makes sense, because it can only be used to `allocate` method. (RBS generates `new` for each classes with types.)

I'm curious if there are more use cases.

ParadoxV5:
> `Class[String]` cannot be equivalent to `singleton(String)`, because `singleton(String)` has singleton methods.

You are correct, the `singleton(String)` syntax enables special handling of singleton methods on the `String` class.
But also note that, anything can have singleton methods, not just classes (`Class` instances). And unfortunately, there is no RBS equivalent for:
```ruby
class << SINGLETON = Superclass.new
  def my_singleton_method …
```
and the current workaround is:
```rbs
SINGLETON: Superclass & _SINGLETON
interface _SINGLETON
  def my_singleton_method …
```

[...]

PR #1655, opened 2023-11-29
stdlib/resolv: Resolv::DNS::Resource::Generic.create returns a class

hanazuki:
`Resolv::DNS::Resource::Generic.create` is not a data constructor but defines a new subclass of `Resolv::DNS::Resource::Generic` and returns it.

```
% irb -rresolv
irb(main):001> rrtype = Resolv::DNS::Resource::Generic.create(65280, Resolv::DNS::Resource::IN::ClassValue)
=> Resolv::DNS::Resource::Generic::Type65280_Class1
irb(main):002> rrtype.class
=> Class
irb(main):003> rrtype.superclass
=> Resolv::DNS::Resource::Generic
```

The result type is some subclass of `Generic`, but, IIUC, RBS currently has no way to express the upper bound for a `Class` (ref. https://github.com/ruby/rbs/issues/1542).

### DIFF
--- a/stdlib/resolv/0/resolv.rbs
+++ b/stdlib/resolv/0/resolv.rbs
@@ -921,7 +921,7 @@ end
 # A generic resource abstract class.
 #
 class Resolv::DNS::Resource::Generic < Resolv::DNS::Resource
-  def self.create: (Integer type_value, Integer class_value) -> instance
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍

PR #1672, opened 2023-12-11
A tiny followup for #1641’s oopsies

ParadoxV5:
* ~~`chomp!`: The `?` suffix in `(?string? separator)` is already covered by `(nil)`.~~
* **`scrub`: Just like `scrub!`, `scrub(nil, &nil)` is a thing.**

---

Question: What shall the convension around `?Regexp | string | nil` vs. `?Regexp | string?` be?

### DIFF
--- a/core/string.rbs
+++ b/core/string.rbs
@@ -1349,7 +1349,7 @@ class String
   # modification made, `self` otherwise.
   #
   def chomp!: (nil) -> nil
-            | (?string separator) -> self?
+            | (?string? separator) -> self? # https://github.com/ruby/rbs/pull/1672#discussion_r1423324796
 
   # <!--
   #   rdoc-file=string.c
--- a/core/string.rbs
+++ b/core/string.rbs
@@ -1349,7 +1349,8 @@ class String
   # modification made, `self` otherwise.
   #
   def chomp!: (nil) -> nil
-            | (?string? separator) -> self? # https://github.com/ruby/rbs/pull/1672#discussion_r1423324796
+          # | (?string separator) -> self? # https://github.com/ruby/rbs/pull/1672#discussion_r1423324796
+            | (?string? separator) -> self?
 
   # <!--
   #   rdoc-file=string.c
--- a/core/string.rbs
+++ b/core/string.rbs
@@ -1349,7 +1349,7 @@ class String
   # modification made, `self` otherwise.
   #
   def chomp!: (nil) -> nil
-            | (?string? separator) -> self?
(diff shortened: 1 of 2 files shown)


ParadoxV5:
re-ping: #1641

core/string.rbs
```diff
@@ -1349,7 +1349,7 @@ class String
   # modification made, `self` otherwise.
   #
   def chomp!: (nil) -> nil
-            | (?string? separator) -> self?
```
soutaro, line 1352: This case is necessary.
Passing a node of type `String?` will be a type error without the `?`.

core/string.rbs
```diff
@@ -2741,7 +2741,7 @@ class String
   #     "\x81"
   #     "\x81"
   #
-  def scrub: (?string replacement) -> String
+  def scrub: (?string? replacement) -> String
```
soutaro, line 2744: 👍

soutaro (changes requested):
Revert the type of `chomp!`. 🙏

core/string.rbs
ParadoxV5: This is a problem with the type checkers unable to break `string` and `nil` from `string?` to different paths and then merge them back to `self? | nil` i.e. `self?`

soutaro: It may be true. Type checkers can do more about this.
But at least for now, RBS is for static type checking and having `string?` is more common for static type checking (I believe.)

ParadoxV5: > But at least for now

How about this: I will revert it **for now**, but we should also remind type checker maintainers about these sort of cases.

ParadoxV5: soutaro/steep#333

ParadoxV5:
No comments? Then how did L648 pass CI
https://github.com/ruby/rbs/blob/a4681af46b58f77c82c15021388e52f7cdf09e6b/core/string.rbs#L648

soutaro:
@ParadoxV5 You uncovered another bug. 🐞 

https://github.com/ruby/rbs/blob/master/lib/rbs/writer.rb#L224 writes lines, from the beginning of `end` to the end of the method type of the last overload. The comment here is after the end of the method type of the last overload.

Having a comment line between the two overloads would pass CI.

```rbs
  def chomp!: (nil) -> nil
            # | (?string separator) -> self? # https://github.com/ruby/rbs/pull/1672#discussion_r1423324796
            | (?string? separator) -> self?

```

soutaro (approved):
👍

PR #1675, opened 2023-12-14
Should raise ParsingError instead of NoMethodError

ksss:
I found a record type pattern that resulted in a `NoMethodError` instead of a `ParsingError`, so I fixed it.

```rb
RBS::Parser.parse_type("{1?: 1}")
#=> /Users/ksss/.rbenv/versions/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/rbs-3.3.2/lib/rbs/parser_aux.rb:7:in `_parse_type': undefined method `literal' for an instance of RBS::Types::Optional (NoMethodError)

      _parse_type(buf, range.begin || 0, range.end || buf.last_position, variables, require_eof)
      ^^^^^^^^^^^
```

It seems that if the record key starts with Symbol, String, Integer, true, or false, and is not a literal type, the `literal` method cannot be called, resulting in a NoMethodError.

Although it is not impossible to optimize the code for each token in detail, I think that using `parse_simple` is a reasonable fix.

### DIFF
--- a/ext/rbs_extension/parser.c
+++ b/ext/rbs_extension/parser.c
@@ -723,7 +723,7 @@ VALUE parse_record_attributes(parserstate *state) {
       case tINTEGER:
       case kTRUE:
       case kFALSE:
-        key = rb_funcall(parse_type(state), rb_intern("literal"), 0);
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks! 🙏

PR #1677, opened 2023-12-15
Add `delegate` library

### LINKED ISSUE #1474: `SimpleDelegator` is missing
Reported in https://github.com/soutaro/steep/issues/862

```
sig/magic/permanent.rbs:39:4: [error] Cannot find type `SimpleDelegator`
│ Diagnostic ID: RBS::UnknownTypeName
│
└     class Protections < SimpleDelegator
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

soutaro:
Closes #1474 

We are adding the type definition of the library to `rbs` repo, not to the repository of `delegate` gem. This is because it's a *default* gem, which is more tightly coupled to ruby distribution than *bundled* gems.

### DIFF
--- a/stdlib/delegate/0/delegator.rbs
+++ b/stdlib/delegate/0/delegator.rbs
@@ -1,0 +1,186 @@
+# <!-- rdoc-file=lib/delegate.rb -->
+# This library provides three different ways to delegate method calls to an
+# object.  The easiest to use is SimpleDelegator.  Pass an object to the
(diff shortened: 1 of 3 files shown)


soutaro:
I don't think so. This allows generated RBS works that inherits from `SimpleDelegator`, but I'm not sure if it works for other situations...

PR #1678, opened 2023-12-15
Add the _Duration type

sampersand:
This adds the `_Duration` type, which is used in most places that need a timeout in the standard lib (eg `Kernel#sleep`, `IO#timeout=`, `Mutex#sleep`, `TCPSocket#initialize`'s `connect_timeout` field, etc.

I couldn't find a good place to put it (it doesn't make much sense under `Time`, because `Time` doesn't even use it), so I ended up putting it under `builtins.rbs`

### DIFF
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -219,6 +219,26 @@ interface _Exception
                | (String arg0) -> Exception
 end
 
+# A type that's used for durations.
+#
+# All numeric types implement this, but custom classes can also implement it if desired.
+#
+# Usage of `_Duration` is fairly common throughout the stdlib, such as in `Kernel#sleep`,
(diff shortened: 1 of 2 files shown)


soutaro:
@sampersand Thank you for opening this PR. I have a few thoughts.

1. `_Duration` might not be a very good name for this. It explains the technical details of the argument, and I don't think it can be called `_Duration` because it doesn't look a duration... (I understand finding a good name/place for this is not easy...)
2. I'm not sure if we need to add this interface. My assumption is:
  * When we try to add a new class that will be used for the duration, then
  * We won't try to implement these methods for this, but just implement a method to translate the duration to seconds

(I might be okay to merging this if you believe we really need it, after finding a good name/place.)

sampersand:
I'll be renaming it to `_Timeout` and adding tests

PR #1681, opened 2023-12-17
Make Pathame() argument type more accurate

KitaitiMakoto:
Currently, `Pathname()` argument's type is declared as `string | Pathname` but actually it accepts objects which responds to `#to_str`.

I agree for this patch to be distributed under RBS's license.


KitaitiMakoto:
Thanks!

PR #1682, opened 2023-12-17
Support Delegator object

ksss:
The Delegator class intentionally rewrites `#methods`, but the rewrite does not work in `Test.call` and the type using interface does not pass the test.

The `#respond_to?` is not rewritten, so the Delegator class is supported by using `#respond_to?` instead of `#methods`.


soutaro (approved):
Thanks!
We should have used `respond_to?`!

PR #1687, opened 2023-12-19
Add `with_***` helpers

soutaro:
This PR moves the helpers like `with_int` from `test/stdlib` to `lib/rbs/unit_test`, so that other gems can use them to test their type definitions.


lib/rbs/unit_test/with_aliases.rb
```diff
@@ -0,0 +1,143 @@
...
+      end
+
+      def with_array(*elements)
+        return WithEnum.new to_enum(__method__ || raise, *elements) unless block_given?
+
+        yield elements
+        yield ToArray.new(*elements)
+      end
+
+      def with_hash(hash = {})
```
ParadoxV5, line 71: `with_array(*elements)`, but not `with_hash(**elements)`?

Note: kwargs isn’t Symbols only and that – good?
```ruby
p('any key' => 'is valid', 420 => 69)
  #=> {"any key"=>"is valid", 420=>69}
```

test/stdlib/test_helper.rb
```diff
@@ -411,15 +106,33 @@ def each(&block)
...
+ToC = RBS::UnitTest::Convertibles::ToC
+ToStr = RBS::UnitTest::Convertibles::ToStr
+ToS = RBS::UnitTest::Convertibles::ToS
+ToSym = RBS::UnitTest::Convertibles::ToSym
+ToA = RBS::UnitTest::Convertibles::ToA
+ToArray = RBS::UnitTest::Convertibles::ToArray
+ToHash = RBS::UnitTest::Convertibles::ToHash
+ToPath = RBS::UnitTest::Convertibles::ToPath
+CustomRange = RBS::UnitTest::Convertibles::CustomRange
+Each = RBS::UnitTest::Convertibles::Each
```
ParadoxV5, line 130: `include RBS::UnitTest::Convertibles`?

soutaro: Ah this is in fact for RBS runtime type checker. I'm not sure if including works for it...

lib/rbs/unit_test/with_aliases.rb
soutaro: We can pass hash without curly brace with the current definition. So any other good reason to make it kwargs?

ParadoxV5: I just prefer consistency.

ksss:
I see potential for property based testing in this framework.
Do you have any plans for such an extension in the future?

```rb
with_int(count: 100) do |i|
  # Yields 100 times random integer value
end
```

lib/rbs/unit_test/convertibles.rb
```diff
@@ -0,0 +1,176 @@
+# frozen_string_literal: true
+
+module RBS
+  module UnitTest
+    module Convertibles
+      class BlankSlate < BasicObject
+        instance_methods.each do |im|
+          next if %i[__send__ __id__].include? im
```
sampersand, line 8: We might want to remove `__id__` from this list, as it doesnt warn when you `undef` it

lib/rbs/unit_test/with_aliases.rb
```diff
@@ -0,0 +1,143 @@
...
+
+      class WithEnum
+        include Enumerable
+
+        def initialize(enum) = @enum = enum
+
+        def each(&block) = @enum.each(&block)
+
+        def and_nil(&block)
+          self.and(nil, &_ = block)
```
sampersand, line 16: We should add `return WithEnum.new to_enum(__method__ || raise) unless block` here

sig/unit_test/convertibles.rbs
```diff
@@ -0,0 +1,154 @@
...
+      #
+      class BlankSlate < BasicObject
+        # Add Object methods to `self`
+        #
+        def __with_object_methods: (*Symbol methods) -> self
+      end
+
+      # An object with `#to_io` method
+      class ToIO < BlankSlate
+        @io: untyped
```
sampersand, line 16: Shouldn't these `untyped`s be given types? ditto for `initialize` and `to_xxx` for each of the `ToXXX` classes

sig/unit_test/with_aliases.rbs
```diff
@@ -0,0 +1,136 @@
...
+                    | (?Float value) -> WithEnum[float]
+
+      # Yields `::string` objects
+      #
+      def with_string: (?String value) { (string) -> void } -> void
+                     | (?String value) -> WithEnum[string]
+
+      # Yields `::array` objects
+      #
+      def with_array: (*untyped elements) { (array[untyped]) -> void } -> void
```
sampersand, line 88: We should give this a type, like `def with_array: [T] (*T elements)`. Same with `hash` and `range`

test/stdlib/File_test.rb
sampersand, line 1: unintentional inclusion?

soutaro: Yeah, it's unrelated, but I think it's acceptable to mix it.

lib/rbs/unit_test/with_aliases.rb
soutaro: and accepts no block call. So it should be ok.

lib/rbs/unit_test/convertibles.rb
soutaro: Hmm. Will try it.

sig/unit_test/convertibles.rbs
soutaro: What is the type of IO here? I think it's not very clear and decided to keep them untyped.

sig/unit_test/with_aliases.rbs
soutaro: Skip `range` method because it requires `<=>` method.

sig/unit_test/convertibles.rbs
sampersand: `@io` should be `::IO`; Same for `ToStr` and `String`, etc

PR #1688, opened 2023-12-19
Add more details error message for UnresolvedOverloadingError

ksss:
In `RBS::Test`, if an exception `UnresolvedOverloadingError` is reported, the following message is displayed, but we don't know the details of what did not match.

```
[Object#foo] UnresolvedOverloadingError: couldn't find a suitable overloading
```

I suggest that detailed information also be logged as follows.

```
W, [2023-12-19T13:35:17.268712 #21135]  WARN -- rbs: [Object#foo] UnresolvedOverloadingError method_type=`() -> ::String` details=["ArgumentError: expected method type () -> ::String", "ReturnTypeError: expected `::String` but returns `30`"], method_type=`(::Integer) -> ::String` details=["ReturnTypeError: expected `::String` but returns `30`"]
[Object#foo] UnresolvedOverloadingError: couldn't find a suitable overloading
```

I have made it so that the detailed information is output in the logger so that it can be turned off if not needed.

### DIFF
--- a/lib/rbs/test/type_check.rb
+++ b/lib/rbs/test/type_check.rb
@@ -35,6 +35,17 @@ module RBS
         if es.size == 1
           errors.push(*es[0])
         else
+          RBS.logger.warn do
+            message = +"[#{self_class}#{method_name}] UnresolvedOverloadingError "
+            message << method.method_types.zip(es).map do |method_type, es|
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👏

PR #1692, opened 2023-12-20
Update IO instance methods and IO.{new,open}

sampersand:
- Add the _Duration type
- Update IO and its tests
- more updates
- working on io.open
- use .allcoate not .new
- initialize now works
- more pudates
- more updates
- fixed
- revert changes to io class methods
- remove reopen and omits for ioctl
- updating tests


core/encoding.rbs
```diff
@@ -28,6 +28,13 @@
 # useful when you use other ASCII-compatible encodings.
 #
 class Encoding
+  # The `_EncodeFallbackAref` is used in functions such as `String::encode` to get the fallback
+  # value when a charcater can't be converted between encodings.
+  interface _EncodeFallbackAref
```
sampersand, line 33: This was moved from `String` to here

lib/rbs/unit_test/spy.rb
```diff
@@ -128,7 +128,7 @@ def wrapped_object
                 end
               end.ruby2_keywords
             )
-          end.new()
+          end.allocate()
```
sampersand, line 131: This is needed because `.new` implicitly calls `.initialize`; if `method_name` is `:initialize`, then this will accidentally call it with no arguments. Doing it this way prevents that (and since the parent object is `BasicObject`, `:initialize` is a no-op anyways)

ParadoxV5: BTW, why not [`MiniTest::Mock`](https://ruby-doc.org/current/gems/minitest/Minitest/Mock.html)?

test/stdlib/test_helper.rb
```diff
@@ -151,6 +151,13 @@ module TestHelper
   include RBS::UnitTest::WithAliases
   include VersionHelper
   include WithStdlibAliases
+<<<<<<< HEAD
+
+  # The current ruby executable path; used in some functions which spawn shells, as the only 
+  # executable we can be guaranteed to exist is ruby itself.
+  RUBY_EXE = ENV['RUBY'] || RbConfig.ruby
+=======
+>>>>>>> master
```
ParadoxV5, line 160:
```suggestion

  # The current ruby executable path; used in some functions which spawn shells, as the only 
  # executable we can be guaranteed to exist is ruby itself.
  RUBY_EXE = ENV['RUBY'] || RbConfig.ruby
```

ParadoxV5:
> * The `%a{ruby:since:3.2.0}` annotation has been added to methods introduced in ruby 3.2.0

We have Ruby 3.3 now (:
* #1674

PR #1694, opened 2023-12-20
Change return type of lifecycle hooks in `minitest` to be `void`

stevegeek:
Hi all, many thanks for all the work on RBS.

I am proposing a change to the Lifecycle hooks of `Minitest::Test` so that they are `void` instead of `nil`.

Otherwise any time we use the `setup` hook for example, we need to explicitly return `nil` but normally we don't care what we return from these methods. The return value is not used in `minitest` I believe (https://github.com/minitest/minitest/blob/master/lib/minitest/test.rb#L91)

I wasn't really sure about how to use the `assert_send_type` test helper but I tried to write a test anyway.

Let me know what you think.

### DIFF
--- a/stdlib/minitest/0/minitest/test/lifecycle_hooks.rbs
+++ b/stdlib/minitest/0/minitest/test/lifecycle_hooks.rbs
@@ -39,7 +39,7 @@ module Minitest::Test::LifecycleHooks
   #       include MyMinitestPlugin
   #     end
   #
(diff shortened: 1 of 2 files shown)


ksss:
I agree with you.
These types are generated by `rbs prototype rb`, so there is no special intention.
LGTM!

stevegeek:
Thanks @ksss !

PR #1695, opened 2023-12-20
Fix klass name if klass is singleton_class

ksss:
When I type-check the singleton method, there is no class/module name in the method name display.

```
[.foo] ReturnTypeError: expected `Integer` but returns `"a"`
```

This is due to the fact that the `#name` of the singleton class returns `nil`.

To solve the problem, I extracted the original class/module name from the string when `#inspect` is called.


soutaro (approved):
👍

PR #1698, opened 2023-12-25
Improve interface check by method parameters

ksss:
Regarding checking interfaces, in addition to checking with `#respond_to?`, I believe that comparing `Method#parameters` with types can provide a more accurate interface check.

As an implementation, I've made it so that a match only occurs if the method has exactly the same arguments as the type notation. 
Does this match with the philosophy of RBS? 
For example, the following code actually works. 

```rbs
# sample.rbs
class Sample
  def foo: (a: Integer, b: String) -> void
end
```

```rb
# sample.rb
class Sample
  def foo(a = {})
    p a # { :a => 1, :b => 'b' }
  end
end

Sample.new.foo(a: 1, b: 'b')
```

However, for instance, steep would report `Ruby::MethodArityMismatch` and `Ruby::DifferentMethodParameterKind`. 
What kind of checks should be attempted with RBS alone?

### DIFF
--- a/test/rbs/test/type_check_test.rb
+++ b/test/rbs/test/type_check_test.rb
@@ -198,9 +198,6 @@ EOF
           end
           assert yes
           assert typecheck.value(success, parse_type(interface))
-        rescue
-          p [:success, definer, interface]
-          raise
         end
 
         [
@@ -252,9 +249,6 @@ EOF
           end
           refute no
           refute typecheck.value(failure, parse_type(interface))
-        rescue
-          p [:failure, definer, interface]
-          raise
         end
       end
     end
--- a/test/rbs/test/type_check_test.rb
+++ b/test/rbs/test/type_check_test.rb
@@ -73,165 +73,188 @@ EOF
     SignatureManager.new do |manager|
       manager.files[Pathname("foo.rbs")] = <<EOF
 interface _NoArgs
-  def args: () -> void
+  def f: () -> void
 end
 interface _PosArgs
-  def args: (Integer) -> void
+  def f: (Integer) -> void
 end
 interface _OptArgs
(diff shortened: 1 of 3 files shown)


soutaro:
I think the parameter checking should be permissive. Having an optional argument in implementation while required argument in RBS should be allowed for example.

soutaro:
Steep validates the method implementation and reports `DifferentMethodParameterKind`. But I don't think RBS should do it on the implementation side.

ksss:
Thank you!
I think it seems good to use the criterion "if it can be called, it's OK"

ksss:
I have reimplemented code to allow for many method call patterns.
I also added an ArgumentChecker to generate arguments from RBS and test whether a method can actually be called.

test/rbs/test/type_check_test.rb
```diff
@@ -69,6 +69,197 @@ def to_int: () -> Integer
...
+          failure = Object.new
+          definer.call(failure)
+
+          no = ArgumentChecker.new(builder: builder, interface: interface).no_argument_error?(:f) do |args, kwargs, block|
+            failure.f(*args, **kwargs, &block)
+          end
+          refute no
+          refute typecheck.value(failure, parse_type(interface))
+        rescue
+          p [:failure, definer, interface]
```
soutaro, line 256: Is this intentionally print using `p`? (Not sure if it should be.)

soutaro (approved):
Didn't review the implementation very carefully, but give it a try! 💪

test/rbs/test/type_check_test.rb
ksss: When a test failed, it was intentionally included to indicate which line of the case had failed.
However, I agree that it is complicated and will remove it.

PR #1701, opened 2023-12-26
Refactor Git source by using `git?` method

pocke:
`git?` has been defined but not used. This method is useful for `need_to_fetch?` method.


soutaro (approved):
👍

PR #1703, opened 2023-12-27
Updated Regexp

sampersand:
This PR updates `Regexp` and related constants/classes, along with completely overhauling the tests. Changes in bold are a bit larger.

More specifically, this:
-  `Regexp::_ToRegexp`: Added, used in `Regexp.{try_convert,union}`.
- `Regexp::TimeoutError` Added
- `Regexp::{EXTENDED,FIXEDENCODING,IGNORECASE,MULTILINE,NOENCODING}`: Moved inside `Regexp`
- `Regexp.last_match`: Now uses `MatchData::capture`
- ** `Regexp.linear_time?`: Filled out the signature (the old one was entirely incorrect)**
- `Regexp.quote`: Now an alias for `Regexp.escape`
- `Regexp.try_convert`: Now has a case for `_ToRegexp`
- `Regexp.timeout=`: Now returns its argument, split out the `nil` case
- **`Regexp.union`: Now uses `_ToRegexp` and implicit `array`. Also added in the `Symbol` case (which is only used if the only argument is a symbol)**
- **`Regexp#initialize`: Updated to have `Regexp` case, along with implicit arguments**
- `Regexp#=~`: Now uses `interned`, split out the `nil` case
- `Regexp#eql?`: Now an alias for `==`
- `Regexp#match{,?}`: Now uses `interned`, split out the `nil` case, `offset` is now `int`

### DIFF
--- a/core/regexp.rbs
+++ b/core/regexp.rbs
@@ -1332,56 +1332,41 @@
 # *   [Rubular](https://rubular.com/).
 #
 class Regexp
-  # <!--
-  #   rdoc-file=re.c
-  #   - Regexp.new(string, options = 0, timeout: nil) -> regexp
-  #   - Regexp.new(regexp, timeout: nil) -> regexp
-  # -->
-  # With argument `string` given, returns a new regexp with the given string and
-  # options:
+  # Represents an object's ability to be converted to a `Regexp`.
   #
-  #     r = Regexp.new('foo') # => /foo/
-  #     r.source              # => "foo"
-  #     r.options             # => 0
+  # This is only used in `Regexp.try_convert` and `Regexp.union` within the standard library.
+  interface _ToRegexp
+    # Converts `self` to a `Regexp`.
+    def to_regexp: () -> Regexp
+  end
+
+  class TimeoutError < RegexpError
+  end
+
+  # <!-- rdoc-file=re.c -->
+  # see Regexp.options and Regexp.new
   #
-  # Optional argument `options` is one of the following:
+  EXTENDED: Integer
+
+  # <!-- rdoc-file=re.c -->
+  # see Regexp.options and Regexp.new
   #
-  # *   A String of options:
+  FIXEDENCODING: Integer
+
+  # <!-- rdoc-file=re.c -->
+  # see Regexp.options and Regexp.new
   #
-  #         Regexp.new('foo', 'i')  # => /foo/i
-  #         Regexp.new('foo', 'im') # => /foo/im
+  IGNORECASE: Integer
+
+  # <!-- rdoc-file=re.c -->
+  # see Regexp.options and Regexp.new
   #
-  # *   The bit-wise OR of one or more of the constants Regexp::EXTENDED,
-  #     Regexp::IGNORECASE, Regexp::MULTILINE, and Regexp::NOENCODING:
+  MULTILINE: Integer
+
+  # <!-- rdoc-file=re.c -->
+  # see Regexp.options and Regexp.new
   #
-  #         Regexp.new('foo', Regexp::IGNORECASE) # => /foo/i
-  #         Regexp.new('foo', Regexp::EXTENDED)   # => /foo/x
-  #         Regexp.new('foo', Regexp::MULTILINE)  # => /foo/m
-  #         Regexp.new('foo', Regexp::NOENCODING)  # => /foo/n
-  #         flags = Regexp::IGNORECASE | Regexp::EXTENDED |  Regexp::MULTILINE
-  #         Regexp.new('foo', flags)              # => /foo/mix
-  #
-  # *   `nil` or `false`, which is ignored.
-  # *   Any other truthy value, in which case the regexp will be case-insensitive.
-  #
-  #
-  # If optional keyword argument `timeout` is given, its float value overrides the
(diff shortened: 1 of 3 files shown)


core/regexp.rbs
```diff
@@ -1558,11 +1550,63 @@ class Regexp
   #
   # If any regexp pattern contains captures, the behavior is unspecified.
   #
-  def self.union: () -> Regexp
-                | (String | Regexp pat1, *String | Regexp pat2) -> Regexp
-                | (::Array[String | Regexp]) -> Regexp
+  def self.union: (*_ToRegexp | string patterns) -> Regexp
+                | (array[_ToRegexp | string] patterns) -> Regexp
+                | (Symbol | [Symbol] symbol_pattern) -> Regexp
```
sampersand, line 1555: Technically this `[Symbol]` should be `array[Symbol]` where the return value is an array of length 1, only containing a symbol, but I think that's overkill here.

lib/rbs/unit_test/spy.rb
```diff
@@ -128,7 +128,7 @@ def wrapped_object
                 end
               end.ruby2_keywords
             )
-          end.new()
+          end.allocate()
```
soutaro, line 131: @sampersand Can you explain why you use `allocate` instead of `new`?

sampersand: It's a minor bug. TL;DR: if you you pass `initialize` as the name of the method, `.new()` will call it with zero params. By usign `.allocate()`, we can safely mock `#initialize()` methods.

core/regexp.rbs
```diff
@@ -1524,7 +1510,8 @@ class Regexp
   #     Regexp.timeout = 1
   #     /^a*b?a*$/ =~ "a" * 100000 + "x" #=> regexp match timeout (RuntimeError)
   #
-  def self.timeout=: (Float?) -> Float?
+  def self.timeout=: [T < _ToF] (T timeout) -> T
+                   | (nil) -> nil
```
soutaro, line 1514: Looks like `nil` has `to_f`, so the `nil` case should be at the first line.

core/regexp.rbs
```diff
@@ -1648,7 +1686,8 @@ class Regexp
   #     /(?<foo>\w+)\s*=\s*#{r}/ =~ 'x = y'
   #     p foo # Undefined local variable
   #
-  def =~: (String? | Symbol | _ToStr str) -> Integer?
+  def =~: (interned str) -> Integer?
+        | (nil) -> nil
```
soutaro, line 1690: Need an optional case here. `(interned?) -> Integer?`.

core/regexp.rbs
```diff
@@ -1774,8 +1813,9 @@ class Regexp
   #      /(.)(.)(.)/.match("abc")[2] # => "b"
   #      /(.)(.)/.match("abc", 1)[2] # => "c"
   #
-  def match: (String? | Symbol | _ToStr str, ?Integer pos) -> MatchData?
-           | [T] (String? | Symbol | _ToStr str, ?Integer pos) { (MatchData) -> T } -> T?
+  def match: (interned str, ?int offset) -> MatchData?
+           | [T] (interned str, ?int offset) { (MatchData matchdata) -> T } -> T?
+           | (nil, ?int offset) ?{ (MatchData matchdata) -> void } -> nil
```
soutaro, line 1818: I think just having `[T] (interned? str, ?int offset) { (MatchData matchdata) -> T } -> T?` would make sense.

core/regexp.rbs
```diff
@@ -1791,7 +1831,8 @@ class Regexp
   #     /P.../.match?("Ruby")    # => false
   #     $&                       # => nil
   #
-  def match?: (String? | Symbol | _ToStr str, ?Integer pos) -> bool
+  def match?: (interned str, ?int offset) -> bool
+            | (nil, ?int offset) -> false
```
soutaro, line 1835: `interned?`

core/regexp.rbs
```diff
@@ -1501,7 +1486,8 @@ class Regexp
   #
   # Raises an exception unless `object.to_regexp` returns a regexp.
   #
-  def self.try_convert: (untyped obj) -> Regexp?
+  def self.try_convert: (_ToRegexp regexp_like) -> Regexp
```
soutaro, line 1489: We should have `(Regexp) -> Regexp`, since we don't have `Regexp#to_regexp`.

test/stdlib/Regexp_test.rb
```diff
@@ -1,181 +1,399 @@
...
+    end
+  end
+end
+
+class RegexpInstanceTest < Test::Unit::TestCase
+  include TestHelper
+
+  testing '::Regexp'
+
+  def test_initialize
```
soutaro, line 203: We usually test `#initialize` with `.new`, so I recommend testing with it if you don't have any specific reason.

lib/rbs/unit_test/spy.rb
soutaro: Got it. I'm not against for using `.allocate` here, but I also recommend using `.new` instead of `#initialize` to test.

soutaro (approved):
Thanks!

PR #1704, opened 2023-12-27
Fix type for Symbol#=~

ksss:
`untyped` is too broad.

```rb
:sym =~ 1
# => undefined method `=~' for an instance of Integer (NoMethodError)
```


soutaro (approved):
👏

PR #1705, opened 2023-12-28
Fix method name tag on assertion

ksss:
Fixed incorrect method tag display for failure indication when using `assert_send_type`.

e.g.)
```
<["[Integerpow] ArgumentTypeError: expected `Integer` but given `0.1`"]> was expected to be empty.
          ^^ missing `#`
```

I have also reorganized the code so that similar mistakes are less likely to occur.


soutaro (approved):
👍

PR #1709, opened 2024-01-05
Update Range#step to accept Numerics and int

sampersand:
Updated per a request. No tests added because `Range_test.rbs` is still on the old version and I didn't want to overhaul it for this pr

### DIFF
--- a/core/range.rbs
+++ b/core/range.rbs
@@ -961,8 +961,8 @@ class Range[out Elem] < Object
   #
   # Related: Range#%.
   #
-  def step: (?Integer n) { (Elem arg0) -> untyped } -> self
-          | (?Integer n) -> ::Enumerator[Elem, void]
+  def step: (?Numeric | int n) -> Enumerator[Elem, self]
+          | (?Numeric | int n) { (Elem element) -> void } -> self
 
   # <!--
   #   rdoc-file=range.c


soutaro (approved):
👍

soutaro:
I found the `Range#step` without block returns an `Enumerator[Elem, Enumerator[Elem, Enumerator[Elem, ...]]]`, not `Enumerator[Enum, self]`. Merged this PR because the infinite type cannot be written in RBS... (`Enumerator[Elem, untyped]` may be better, but not sure if it makes more sense...)

Adding a new syntax something like `Enumerator[Enum, T] as T` would make giving precise type to this, but...

ParadoxV5:
> Merged this PR because the infinite type cannot be written in RBS... (`Enumerator[Elem, untyped]` may be better, but not sure if it makes more sense...)
> 
> Adding a new syntax something like `Enumerator[Enum, T] as T` would help giving precise type to this, but...

@soutaro Type aliases can be recursive.

~~@sampersand already added a bunch of interfaces to describe the CRuby implementation details, another one-time-use type alias wouldn’t hurt…~~

soutaro:
@ParadoxV5 Yeah! True. We can define it by adding a helper recursive type alias!

```rbs
type recursive_enum[T] = Enumerator[T, recursive_enum[T]]
```

PR #1711, opened 2024-01-05
Add Numeric Node's check for parse rbs

S-H-GAMELINKS:
I'll introduce  Numeric Node's for the parser in https://github.com/ruby/ruby/pull/9417.

This changed the behavior of the rbs parse process and caused the test to fail.

This pull request has been modified to check Numeric Node's in parse process.

### DIFF
--- a/lib/rbs/prototype/rb.rb
+++ b/lib/rbs/prototype/rb.rb
@@ -584,6 +584,14 @@ module RBS
           Types::Literal.new(literal: false, location: nil)
         when :NIL
           Types::Bases::Nil.new(location: nil)
+        when :INTEGER
+          Types::Literal.new(literal: node.children[0], location: nil)
+        when :FLOAT
+          BuiltinNames::Float.instance_type
+        when :RATIONAL, :IMAGINARY
+          lit = node.children[0]
+          type_name = TypeName.new(name: lit.class.name.to_sym, namespace: Namespace.root)
+          Types::ClassInstance.new(name: type_name, args: [], location: nil)
         when :LIT
           lit = node.children[0]
           case lit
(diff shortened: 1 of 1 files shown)


lib/rbs/prototype/rb.rb
```diff
@@ -686,6 +694,12 @@ def range_element_type(types)
 
       def param_type(node, default: Types::Bases::Any.new(location: nil))
         case node.type
+        when :INTEGER
+          BuiltinNames::Integer.instance_type
+        when :FLOAT
+          BuiltinNames::Float.instance_type
+        when :RATIONAL, :IMAGINARY
+          BuiltinNames::Numeric.instance_type
```
ksss, line 702: How about to like this?

```suggestion
        when :RATIONAL
          Types::ClassInstance.new(name: TypeName("::Rational"), args: [], location: nil)
        when :IMAGINARY
          Types::ClassInstance.new(name: TypeName("::Complex"), args: [], location: nil)
```

lib/rbs/prototype/rb.rb
```diff
@@ -584,6 +584,14 @@ def literal_to_type(node)
           Types::Literal.new(literal: false, location: nil)
         when :NIL
           Types::Bases::Nil.new(location: nil)
+        when :INTEGER
+          Types::Literal.new(literal: node.children[0], location: nil)
+        when :FLOAT
+          BuiltinNames::Float.instance_type
+        when :RATIONAL, :IMAGINARY
+          lit = node.children[0]
```
ksss, line 592: Is this a "literal"?

ksss (commented):
Great works!

lib/rbs/prototype/rb.rb
S-H-GAMELINKS: Thanks for the review!
I've fixed it in the following commit.
https://github.com/ruby/rbs/pull/1711/commits/225ed2205df26f51cb4f7f24139ba34c8a1a0fcd

soutaro (approved):
Thanks!

Issue #504, opened 2020-12-05
Record type with optional keys


HoneyryderChuck:
The [record type](https://github.com/ruby/rbs/blob/master/docs/syntax.md#record-type) definition forces all keys to be passed, whereas in most cases, one already handles the case to a default when one of the record values isn't passed. Picking up the linked example:

```ruby
{ id: Integer, name: String }     # Hash object like `{ id: 31, name: "John" }`
# However, I want to handle
# Hash object like `{ name: "John" }` where id is attributed inside the function
# this isn't handled because I can't define
# { ?id: Integer, name: String } 
# like for kwargs
```

The only way to support this right now is to repeat all combinations in a type definition, which is a bit brittle.

WDYT?

soutaro:
In fact, the existence of the key is not checked in the runtime type checker and Steep also allows omitting optional record keys. (https://github.com/soutaro/steep/pull/256)

So, my intuition is that supporting optional keys is better, but we can postpone working for it after Ruby 3.0.

HoneyryderChuck:
@soutaro do you think it's time for devising a plan for this? Caught this again while using `steep` and stdlib's `URI::Generic.build` signature, which is currently incorrect, as it expects all options to be passed, whereas they're all optional.

ShalokShalom:
I would be also interested in this.

PR #1717, opened 2024-01-11
parser: enable record types with optional fields

HoneyryderChuck:
this enables richer definitions of record types, which can only be defined nowadays with Hash[Symbol, untyped].

To handle #504

### DIFF
--- a/ext/rbs_extension/parser.c
+++ b/ext/rbs_extension/parser.c
@@ -698,15 +698,22 @@ static VALUE parse_proc_type(parserstate *state) {
                      | {} literal_type `=>` <type>
 */
 VALUE parse_record_attributes(parserstate *state) {
-  VALUE hash = rb_hash_new();
+  VALUE fields = rb_hash_new();
 
   if (state->next_token.type == pRBRACE) {
-    return hash;
+    return fields;
   }
 
   while (true) {
-    VALUE key;
-    VALUE type;
+    VALUE key, type,
+          value = rb_ary_new(),
+          required = Qtrue;
+
+    if (state->next_token.type == pQUESTION) {
+      // { ?foo: type } syntax
+      required = Qfalse;
+      parser_advance(state);
+    }
 
     if (is_keyword(state)) {
       // { foo: type } syntax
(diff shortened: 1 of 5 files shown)


HoneyryderChuck:
@soutaro this is a draft proposal, as in, the optional fields are parsed and stored, but not used yet. If you agree with this, I can proceed and use it in inferences (may need some help with that eventually as well).

soutaro:
@HoneyryderChuck Thanks! Let's go with the syntax.

Having different hash tables for required and optional fields would result breaking the order of the records, for `Writer` or rewriting features. I'd like to add `#all_fields` attribute and let `#fields` and `#optional_fields` fields filtered from `#all_fields`.

```rbs
    class Record
      attr_reader fields: Hash[Symbol, [t, bool]]

      type loc = Location[bot, bot]

      def initialize: (fields: Hash[Symbol, t], location: loc?) -> void      # For compatibility with 3.4
                          | (all_fields: Hash[Symbol, [t, bool]], location: loc?) -> void

      include _TypeBase

      def fields: () -> Hash[Symbol, t]
      def optional_fields: () -> Hash[Symbol, t]?
    end
```

HoneyryderChuck:
@soutaro can you check the changes? I believe this is what you were aiming for?

soutaro:
Thanks @HoneyryderChuck! Looks great. Do you think it's ready to merge?

ext/rbs_extension/parser.c
```diff
@@ -52,6 +52,11 @@ typedef struct {
   VALUE rest_keywords;
 } method_params;
 
+typedef struct {
```
soutaro, line 55: Looks like the struct is no longer used.

sig/types.rbs
```diff
@@ -301,16 +301,21 @@ module RBS
...
       def initialize: (fields: Hash[Symbol, t], location: loc?) -> void
+                    | (all_fields: Hash[Symbol, [t, bool]], location: loc?) -> void
 
       include _TypeBase
 
       attr_reader location: loc?
 
+      def fields: () -> Hash[Symbol, t]
+
+      def optional_fields: () -> Hash[Symbol, t]?
```
soutaro, line 317: Ah, I made a mistake to put `?` at the end of the method type. It should return an empty hash.

```suggestion
      def optional_fields: () -> Hash[Symbol, t]
```

HoneyryderChuck: if you look at the implementation, nothing is returned if all_fields is the same as fields. I think that's fine, i.e. we can avoid creating that extra hash when not necessary. Or do you see any use for it?

ext/rbs_extension/parser.c
HoneyryderChuck: thx, removed.

HoneyryderChuck:
@soutaro anything on the above?

soutaro (approved):
Thanks! 👍

PR #1726, opened 2024-01-19
Bckport #1724

soutaro:
Stop trace object allocations during test


soutaro:
Some tests are failing, but it's fine.

PR #1733, opened 2024-01-24
Clarify deprecation state of `--silent` option of `rbs validate`

pocke:
Currently `--silent` option just displays ``"`--silent` option is deprecated." ``. The user cannot decide what they should do from this message. So I updated the message to describe that the silent mode is default and they can use `--log-level` option also.

ref: https://github.com/ruby/rbs/pull/1648

cc/@ksss


ksss (commented):
👌

PR #1738, opened 2024-01-28
Update type for CSV.foreach

m11o:
This Pull Request updates the type signatures of `CSV.foreach` in RBS. 
When CSV.foreach is called with the headers keyword argument, the block parameter or the elements of the returned Enumerator are of type `CSV::Row`. If the headers argument is not provided, the elements of the Enumerator will be arrays with elements of type String. 

This change accurately reflects the behavior of CSV.foreach depending on the presence of the headers keyword argument.

This Pull Request does not include changes for other methods that accept the headers keyword argument and have behavior variations based on this argument. After this Pull Request is merged, I plan to address these methods in a separate Pull Request.

Thanks.

CSV.foreach: https://github.com/ruby/csv/blob/0cba3e766d34373043b4ca0bbae93c8de12014f0/lib/csv.rb#L1332

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -18,6 +18,7 @@ GEM
     ast (2.4.2)
     base64 (0.2.0)
     bigdecimal (3.1.6)
+    csv (3.2.8)
     dbm (1.1.0)
     diff-lcs (1.5.0)
     digest (3.1.1)
@@ -102,6 +103,7 @@ DEPENDENCIES
   abbrev
   base64
   bigdecimal
+  csv
   dbm
   digest
   fileutils
--- a/stdlib/csv/0/csv.rbs
+++ b/stdlib/csv/0/csv.rbs
@@ -1722,7 +1722,10 @@ class CSV < Object
   #     would read `UTF-32BE` data from the file but transcode it to `UTF-8`
   #     before parsing.
   #
-  def self.foreach: [U] (String | IO | StringIO path, ?::Hash[Symbol, U] options) { (::Array[String?] arg0) -> void } -> void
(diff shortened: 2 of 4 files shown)


stdlib/csv/0/csv.rbs
```diff
@@ -1722,7 +1722,10 @@ class CSV < Object
   #     would read `UTF-32BE` data from the file but transcode it to `UTF-8`
   #     before parsing.
   #
-  def self.foreach: [U] (String | IO | StringIO path, ?::Hash[Symbol, U] options) { (::Array[String?] arg0) -> void } -> void
+  def self.foreach: (String | IO path, ?String mode, headers: true, **untyped options) { (::CSV::Row arg0) -> void } -> void
```
m11o, line 1725: Accurately, the `headers` keyword argument can't be only received true, but also truthy value, for example, String and Symbol, etc.

However, As headers argument is often provided true, I defined true only in this methods.
Please tell me the way of definitions much better than this if you know. Thanks.

Gemfile.lock
```diff
@@ -18,6 +18,7 @@ GEM
     ast (2.4.2)
     base64 (0.2.0)
     bigdecimal (3.1.6)
+    csv (3.2.8)
```
m11o, line 21: Add csv gem to Gemfile because it is changed from default gem to bundled gem in ruby 3.4.

soutaro (approved):
Thanks!

Test is failing but unrelated to this change.

soutaro:
Technically, the type definition will result in an incorrect result:

```rbs
bool = true
CSV.foreach(..., headers: bool) do |row|
  # `bool` is `bool` type, not `true`, and `row` is `String`.
end
```

I think it's not too weird to assume we pass a `true` literal typically, as `headers: true`.

PR #1745, opened 2024-02-05
WIP update Float

sampersand:
- Moved float constants inside float
- updating float


soutaro:
The changes look good, but it doesn't work well with the current version of Steep. It is a kind of bug of Steep, so I will fix the problem and then merge this PR, to avoid unnecessary type checking issues at the end users side.

PR #1749, opened 2024-02-07
uri: Use `instance` instead of static type as return type

ThisGuyCodes:
In the three methods I've modified the return type is always an instance of the class the method is called on, however they were hardcoded to a specific class.

I ran into `URI::HTTPS.Build` thinking it returns a `URI::HTTP`, which prompted this PR, but I decided to go up the chain and saw `URI::Generic` did the same thing (`URI::Generic` returns the result of `self.new`, so `instance` is a more true-to-code return type).

### DIFF
--- a/stdlib/uri/0/generic.rbs
+++ b/stdlib/uri/0/generic.rbs
@@ -154,7 +154,7 @@ module URI
     # then it does URI::Escape.escape all URI components and tries again.
     #
     def self.build2: (Array[nil | String | Integer]) -> URI::Generic
-                   | ({ scheme: String?, userinfo: String?, host: String?, port: Integer?, registry: String?, path: String?, opaque: String?, query: String?, fragment: String? }) -> URI::Generic
+                   | ({ scheme: String?, userinfo: String?, host: String?, port: Integer?, registry: String?, path: String?, opaque: String?, query: String?, fragment: String? }) -> instance
 
     # <!--
     #   rdoc-file=lib/uri/generic.rb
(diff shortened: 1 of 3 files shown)


soutaro:
Thanks @ThisGuyCodes! 

I don't think making `URI::Generic.build` return `instance` is the best solution for the problem.
It looks like most of the subclasses of `URI::Generic` have their own `build` implementations. So, the correct fix would be adding `build` methods to the subclasses.

```
irb(main):003> URI::File.method(:build)
=> #<Method: URI::File.build(args) /Users/soutaro/.rbenv/versions/3.3.0/lib/ruby/3.3.0/uri/file.rb:53>
irb(main):004> URI::WS.method(:build)
=> #<Method: URI::WS.build(args) /Users/soutaro/.rbenv/versions/3.3.0/lib/ruby/3.3.0/uri/ws.rb:56>
```

:+1: for `HTTP` class, because it's subclass `HTTPS` doesn't have custom implementation of `build`.

ThisGuyCodes:
@soutaro I agree that the correct fix is to add `build` methods to the subclasses, but since the parent class here returns "whatever class I am" wouldn't we want the "default" behavior to match this? Which would then be overridden by subclasses that implement their own version (unless I'm not understanding inheritance right, which is possible).

soutaro:
@ThisGuyCodes Got it. It returns `self.new`, so `instance` would be better.

https://github.com/ruby/uri/blob/master/lib/uri/generic.rb#L136

I was confused because the doc says it returns *a new URI::Generic instance*. (It is incorrect...)

soutaro (approved):
👍

PR #1750, opened 2024-02-07
net-http: Allow symbol based hashes for headers

ThisGuyCodes:
These methods work fine with headers given as `Hash[Symbol, untyped]`. I noticed when I got some squiggly lines in some of my methods that I already know work. I also defined `headers` as a type alias rather than copy-pasting the same definition everywhere (I can undo this if not desired).

I did `(Hash[String, untyped] | Hash[Symbol, untyped])` rather than `Hash[String | Symbol, untyped]` because hashes with mixed key types can allow name collisions which gives unexpected behavior.

I also feel like we can do better than `untyped` for the values, but leaving it for now.


soutaro (approved):
Thanks!

PR #1762, opened 2024-03-10
[rbs diff] Fix error when empty manifest.yaml

ksss:
I have found a problem with an error when manifest.yaml exists but is all comments.

```
rbs/lib/rbs/diff.rb:86:in `block in build_env': undefined method `[]' for nil (NoMethodError)

          if manifest['dependencies']
                     ^^^^^^^^^^^^^^^^
```


soutaro (approved):
Thanks! 👍

PR #1763, opened 2024-03-11
Should return void at `#initialize`

ksss:
There is a problem if the `#initialize` type is fixed.

The user must match or override the return type of `#initialize` in the inherited class.
The `#initialize` does not need to care about the return type, and it should be possible that it may change in the future.

This PR has been modified to unify all `#initialize` method return types with void.


sw-square:
I'm not a huge fan of this, as I can see situations where people might want to do custom overrides of `initialize`. But I dont have a big enough reason other than "this seems unnecessarily restrictive" to reject it

ksss:
@sw-square Thank you for reviewing.

An example I can think of is as follows.

```rb
# sample/cli.rb
module Sample
  class CLI
    class << self
      attr_accessor :logger
    end

    # Cannot allow method body have type `::Logger` because declared as type `nil`
    #   ::Logger <: nil(Ruby::MethodBodyTypeMismatch)
    def initialize
      CLI.logger = Logger.new($stdout)
    end
  end
end
```

When I used steep, `Sample::CLI#initialize` gave me a type error.

There are two ways to resolve this type error.
Either put `nil` at the end of `#initialize` or write the type of `Sample::CLI#initialize`.

```rb
# sample/cli.rb
module Sample
  class CLI
    def initialize
      CLI.logger = Logger.new($stdout)
      nil
    end
  end
end
```

or

```rbs
# sample/cli.rbs
module Sample
  class CLI
    def initialize: () -> void
  end
end
```

Neither of those methods seems like the best way to me.

PR #1767, opened 2024-03-15
Make the test suite compatible with `--enable-frozen-string-literal`

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/20205
Ref: https://github.com/voxpupuli/json-schema/pull/508

Since `rbs` is tested as part of ruby-core CI, it needs to be compatible with the `--enable-frozen-string-literal` option.

~~NB: The `json-schema` fix was merged, but not sure when it will be released. I don't know if using a git gem is a blocker?~~

### DIFF
--- a/test/stdlib/String_test.rb
+++ b/test/stdlib/String_test.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: false
 require_relative 'test_helper'
 
 # TODO: encode, encode!, byteslice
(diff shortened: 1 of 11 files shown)


test/stdlib/String_test.rb
```diff
@@ -1,3 +1,4 @@
+# frozen_string_literal: false
```
casperisfine, line 1: Lots of tested methods mutate the string, so it's simpler to globally use mutable strings in this file.

casperisfine:
> The json-schema fix was merged, but not sure when it will be released

It just got released 🎉 . So I update the PR to no longer use a git gem.

soutaro (approved):
Thanks!

PR #1768, opened 2024-03-18
String gsub!

ksss:
I found next pattern.

```rb
'a'.gsub!(/b/).each { }
#=> nil
```

And I also fixed the cause of the `test_gsub!` being omit.
Then I have also added more test cases.


ksss:
Oh, it needs a few more revisions.

ksss:
Ready to merge.

soutaro (approved):
Thanks! 👍

PR #1769, opened 2024-03-19
Requires at least one argument.

ksss:
Fixed to require at least one argument, since only the rest argument would also allow zero arguments.


soutaro (approved):
👍

PR #1770, opened 2024-03-19
Zone is nilable

ksss:
The `Time#zone` can be `nil`.

ruby/spec:
https://github.com/ruby/spec/blob/e2f39fe241d6e0443f3b9e9f989f4cc6a061d06c/core/time/zone_spec.rb#L14-L16


soutaro (approved):
👍

PR #1773, opened 2024-03-22
Avoid string allocation on unquoting string

pocke:
TODO: write doc comment
TOOD: performance testing


pocke:
Hmm, it improves only a few performances. So I'll close it.

PR #1774, opened 2024-03-25
Always be private

ksss:
Instance methods of `#initialize`, `#initialize_copy`, `#initialize_clone`, `#initialize_dup` and `#respond_to_missing?` are always private in CRuby.

https://github.com/ruby/ruby/blob/69579ed57a2aa1c3ad739417db70564d570bf2c1/vm_method.c#L966-L977

I think it would be difficult to always mandate this behavior in the RBS signature, so I suggest that it be automatically set to private.


soutaro (approved):
Thanks! 👍

PR #1776, opened 2024-03-26
Modify template to use TestHelper

ksss:
I believe it might have been overlooked in https://github.com/ruby/rbs/pull/1660


soutaro (approved):
👍

PR #1779, opened 2024-03-27
Refine Digest

ksss:
- Should be return bot for methods of abstruct class. Use interface instead.
- The number of arguments of Digest::Class methods cannot be determined because they are delegated to the inherited initialize.
- Fix some types.

### Good points

- Become a more accurate signatures.

### Bad points

- Comments are effectively lost because annotations cannot be automatically written on the interface.

### DIFF
--- a/stdlib/digest/0/digest.rbs
+++ b/stdlib/digest/0/digest.rbs
@@ -74,7 +74,7 @@ module Digest
   # -->
   # Returns a BubbleBabble encoded version of a given *string*.
   #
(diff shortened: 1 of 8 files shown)


soutaro:
Having `-> bot` types doesn't look a good idea, especially for `Digest::Base`. We may want to have `Digest::Base` type which might be one of the sub classes instances, and it all method calls will result in `bot`.

ksss:
I understand. I would like to modify it with a different approach. I will close this PR.

PR #1783, opened 2024-04-03
Fix Kernel.rand signature

ksss:
`Kernel.rand` has many patterns.


soutaro (approved):
Having a overload `(int) -> Float` might confuse users... Yeah, it's correct, but...

PR #1787, opened 2024-04-04
Add location for attribute

ksss:
I noticed that location is not set for the attribute method. Is this intentional?

I actually have some problems with the `rbs method` command, such as not being able to find the location where the attribute method is defined.

I think it is better to have attribute location than not to have it, so I fixed it.

---

Before

```
$ bundle exec rbs -r rbs method --singleton RBS::Test suffix
::RBS::Test.suffix
  defined_in: ::RBS::Test
  implementation: ::RBS::Test
  accessibility: public
  types:
      () -> ::String   at
```

After

```
$ bundle exec rbs -r rbs method --singleton RBS::Test suffix
::RBS::Test.suffix
  defined_in: ::RBS::Test
  implementation: ::RBS::Test
  accessibility: public
  types:
      () -> ::String   at /Users/ksss/src/github.com/ksss/rbs/sig/test.rbs:76:4...76:37
```

### DIFF
--- a/lib/rbs/definition_builder.rb
+++ b/lib/rbs/definition_builder.rb
@@ -669,7 +669,7 @@ module RBS
                 ]
               ),
               block: nil,
-              location: nil
+              location: original.location
             )
           else
             # getter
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks!

PR #1788, opened 2024-04-05
Omit unnecessary field from location range

pocke:
This PR reduces memory usage for `RBS::Location` by omitting unnecessary fields.

Previously `RBS::Location` had byte pos, char pos, line, and column. But they are caches except for char pos. These caches are rarely accessed. So we can drop them from `RBS::Location`.

design doc: https://hackmd.io/QNkMU6roTqSn_iXb2ENhsA

## Benchmarking

I used `benchmark/memory_new_rails_env.rb` for the benchmark.
It decreases 15% memory.

```
# Before
Total retained:  52372539 bytes (459882 objects)

allocated memory by class
-----------------------------------
 152307312  Hash
  20146528  RBS::Location
  15744181  String
  15124312  Array


# After
Total retained:  44234355 bytes (459882 objects)

allocated memory by class
-----------------------------------
 152307312  Hash
  15744181  String
  15124312  Array
  12008296  RBS::Location
```

---

This PR also fixes two problems of the current implementations.

First, `Buffer#pos_to_loc` returns the wrong line and column for the EOS position of a string without the end of a newline. For example, the location for `abc|` (`|` is the cursor) should be `[1, 3]`, but `#pos_to_loc` unexpectedly returns `[2, 0]`.
This problem is fixed in https://github.com/ruby/rbs/pull/1788/commits/81b44e1de4fd1f9b737b8d8cd8aa660b17bb952c

Second, `#pos_to_loc` returns an unexpected value when the position is located EOF.
The cause is that the parser determines the length of EOF as 1. Then, the `RBS::Location` of the `pEOF` token contains an out-of-range position. Because an EOF has a length in C String, but it doesn't have a length in Ruby String.
I changed the length of EOF token to zero to fix this problem. 
https://github.com/ruby/rbs/pull/1788/commits/16006c2c8cce5d29995936c2ca507be0268aba97

### DIFF
--- a/ext/rbs_extension/lexer.c
+++ b/ext/rbs_extension/lexer.c
@@ -116,7 +116,7 @@ token rbsparser_next_token(lexstate *state) {
 yy1:
 	rbs_skip(state);
 #line 144 "ext/rbs_extension/lexer.re"
-	{ return next_token(state, pEOF); }
+	{ return next_eof_token(state); }
 #line 121 "ext/rbs_extension/lexer.c"
 yy2:
 	rbs_skip(state);
--- a/ext/rbs_extension/lexer.h
+++ b/ext/rbs_extension/lexer.h
@@ -167,6 +167,11 @@ void skipn(lexstate *state, size_t size);
  * */
 token next_token(lexstate *state, enum TokenType type);
 
+/**
+ * Return new token with EOF type.
+ * */
+token next_eof_token(lexstate *state);
+
 token rbsparser_next_token(lexstate *state);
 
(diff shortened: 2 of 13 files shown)


soutaro:
Great work! 👏 

`RBS::Location` now stores `rbs_loc_range`, a pair of start char pos and end char pos, but other C internal code keeps using `range`, with additional fields. Because the `range` is internal and not exposed as Ruby objects, this PR reduces the memory usage. 👍

soutaro (approved):
🎉

PR #1789, opened 2024-04-05
Should focus on `String.new` not `String#initialize`.

ksss:
There is a slight difference in the allowed types between `String#initialize` and `String#new`.

The current signature focuses on `String#initialize`, and includes cases where `String.new` would fail.

```rb
String.new(capacity: nil)
#=> no implicit conversion from nil to integer (TypeError)
```

There's an option to separate the signatures for `String.new` and `String#initialize`, but it's hard to see the benefit of allowing `capacity` and `encoding` to be nil.

In this revision, I've changed to a signature that only works with `String.new`.
I believe this is simpler and more practical.

### DIFF
--- a/core/string.rbs
+++ b/core/string.rbs
@@ -703,7 +703,7 @@ class String
   #
   #     String.new('hello', encoding: 'UTF-8', capacity: 25)
   #
-  def initialize: (?string source, ?encoding: encoding?, ?capacity: int?) -> self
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍 

I agree testing `#initialize` method by itself is unnecessary, while it's not too bad if `.new` is tested. (But it was missing for `String`...)

PR #1790, opened 2024-04-05
Fix SEGV with parse negative position

ksss:
I've found the following cases. It will raise SEGV.

```rb
RBS::Parser.parse_type("a", range: -2...-1)
```


soutaro (approved):
👍

PR #1796, opened 2024-04-09
Add infinity pattern

ksss:
Add a pattern to `Array#cycle` that returns an Enumerator that loops infinitely, as it is missing.

Whether it loops infinitely or not depends on the receiver and cannot be isolated by overloading.

```rb
[1,2,3].cycle.each {} # Return type should be `bot` since it will never return

[].cycle.each {} # Just return `nil`
```

### DIFF
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -1311,7 +1311,7 @@ class Array[unchecked out Elem] < Object
   #     [0, 1].cycle.first(5) # => [0, 1, 0, 1, 0]
   #
   def cycle: (?int? n) { (Elem) -> void } -> nil
-           | (?int? n) -> ::Enumerator[Elem, nil]
+           | (?int? n) -> (::Enumerator[Elem, nil] | ::Enumerator[Elem, bot])
 
   # <!--
   #   rdoc-file=array.c
--- a/test/stdlib/Array_test.rb
+++ b/test/stdlib/Array_test.rb
@@ -240,15 +240,54 @@ class ArrayInstanceTest < Test::Unit::TestCase
   end
 
   def test_cycle
+    yield_self do
+      # when empty receiver
+      assert_send_type(
+        "() -> Enumerator[Integer, nil]",
+        [], :cycle
+      )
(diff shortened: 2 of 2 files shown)


soutaro:
@ksss I'm not sure if adding the `bot` case here. We don't have `bot` type with `IO#gets` while it doesn't return when the stream is closed. (I'm curious if having it improves something, and open to have it if it does.)

ksss:
The key point of this issue is that it's not about `bot` itself, but about the type argument.
There is no way to write a test case for `[1,2,3].cycle` despite it having a return value.

```rb
assert_send_type(
  "() -> Enumerator[untyped, untyped]",
  [1,2,3], :cycle
)
# Failure: test_cycle(ArrayInstanceTest):
#   Call trace does not match with given method type: #<struct RBS::Test::CallTrace method_name=:cycle, method_call=<RBS::Test::ArgumentsReturn:@arguments: [], @exit_value: #<Enumerator: [1, 2, 3]:cycle>, @exit_type: :return>, block_calls=[], block_given=false>.
#   <["[Array#cycle] ReturnTypeError: expected `Enumerator[untyped, untyped]` but returns `#<Enumerator: [1, 2, 3]:cycle>`"]> was expected to be empty.
```

-----

Another perspective is that infinite loops can be a critical issue for applications. If programmers could determine the possibility of infinite loops from type information, they could avoid these problems. If it is known that there is a possibility of returning `bot`, programmers can avoid calling `each` from `Enumerator[untyped, bot]`.
However, whether this applies to `IO#gets` is unclear.

soutaro:
I still don't agree having the `Enumerator[.., bot]` case makes sense.

1. A method call *may* not return doesn't mean we should have `bot` return type.
2. We cannot use the two `Enumerator` union cases. We cannot decompose them with `is_a?`, and the `each` return type will be `nil | bot`, which is technically identical to `bot`.

ksss:
OK make sense!
Thank you for reviewing.

PR #1797, opened 2024-04-11
More details `Set.new`

ksss:
Since the implementation of Set is Hash, the key must have a `hash` method.

```
$ ruby -e 'Set.new([BasicObject.new])'
/Users/ksss/.rbenv/versions/3.3.0/lib/ruby/3.3.0/set.rb:509:in `add': undefined method `hash' for an instance of BasicObject (NoMethodError)

    @hash[o] = true
         ^^^^^
```

The type argument `A` for Set should strictly be `A < Hash::_Key`. However, making this change would break compatibility, so it cannot be done.
In practice, using BasicObject is rare, so there is no need to make this change.

### DIFF
--- a/lib/rbs/test.rb
+++ b/lib/rbs/test.rb
@@ -72,6 +72,7 @@ module RBS
     end
 
     CallTrace = Struct.new(:method_name, :method_call, :block_calls, :block_given, keyword_init: true)
+    CallTrace.include Guaranteed::Inspect
 
     class <<self
       attr_accessor :suffix
--- a/core/set.rbs
+++ b/core/set.rbs
@@ -197,9 +197,9 @@ class Set[unchecked out A]
   #     Set.new(1..5)                         #=> #<Set: {1, 2, 3, 4, 5}>
   #     Set.new([1, 2, 3]) { |x| x * x }      #=> #<Set: {1, 4, 9}>
   #
-  def initialize: (_Each[A]) -> untyped
-                | [X] (_Each[X]) { (X) -> A } -> untyped
-                | (?nil) -> untyped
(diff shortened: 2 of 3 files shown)


lib/rbs/test.rb
```diff
@@ -72,6 +72,7 @@ def break?
     end
 
     CallTrace = Struct.new(:method_name, :method_call, :block_calls, :block_given, keyword_init: true)
+    CallTrace.include Guaranteed::Inspect
```
ksss, line 75: Required for testing when a BasicObject is used.

If not...

```
Error: test_new(SetSingletonTest): NoMethodError: undefined method `inspect' for #<BasicObject:0x0000000128fd0560>
...snip...
  => 50:     assert_send_type "(Object & _Each[Hash::_Key]) -> Set[Hash::_Key]",
     51:                      Set, :new, o
```

soutaro:
@ksss 

> The type argument `A` for Set should strictly be `A < Hash::_Key`. However, making this change would break compatibility, so it cannot be done.

I'm fan of making the type parameter bounded, `A < Hash::_Key`. Did you see any specific compatibility problem with it?

ksss:
For `Set` only `json` has been found, but I have the following concern.

- Only [ruby/rbs](https://github.com/ruby/rbs) and [ruby/gem_rbs_collection](https://github.com/ruby/gem_rbs_collection) were checked, and I am not sure if they are also completely absent in the sig directory distributed in the rubygems.
- The same issue exists with `Hash`, which is used in activesupport.

ksss:
This PR has resolved by https://github.com/ruby/rbs/pull/1810

PR #1799, opened 2024-04-12
redefine superclass in Object as always returning a class

HoneyryderChuck:
the only instance of a nil superclass is basic object.

This is a proposal to make this definition more precise.


soutaro:
@HoneyryderChuck 

I don't think moving the `superclass` method definition to `Object` is a good idea. The method is defined in `Class` class.

Having a special definition in `BasicObject` might make sense, because we can skip testing `superclass` value every time, except it's `BasicObject`.

PR #1803, opened 2024-04-16
Add Timeout::ExitException 

ksss:
Timeout::ExitException is an internal error.
However, it can appear in Ruby code.

```rb
Timeout.timeout(3) do
  begin
    long_run
  rescue StandardError => e
    logger.error(e)
    raise
  rescue Timeout::ExitException
    # skip logging
    raise
  end
end
```


soutaro (approved):
👍

Issue #1622, opened 2023-11-17
Add in `&untyped`


sampersand:
Currently, `*untyped` and `**untyped` can be used to accept any amount of any type of positional and keyword arguments. However, there's no way to accept a block of any type:
```rbs
def foo: () { (*untyped, **untyped) -> untyped } -> void
```
indicates that the block doesn't accept block parameters itself. To fix this, you need to do
```rbs
def foo: () { (*untyped, **untyped) { (*untyped, **untyped) -> untyped } -> untyped } -> untyped
```
But _that_ block also needs a block itself, etc. ad nauseum. So, my proposal is to add in `&untyped` syntax, eg `def foo: (&untyped) -> void`. (This syntax also allows the possibility of naming the blocks—`&untyped foo`, and might allow for naming block variables themselves, such as: `def yield_ten: (&block) { (Integer) -> void } -> void`

ParadoxV5:
Add a `?` prefix to indicate optional block
```rbs
def foo: () ?{ (*untyped, **untyped) -> void } -> void
```

---

Possible alternative:
```rbs
def foo: () ?{ (...) -> void } -> void
```
Where `...` eats everything just like in
```rb
def example(...) = 42 # The entire `...` is unused
```

soutaro:
The proposal makes sense. Will compose a proposal for block syntax updates including this.

ParadoxV5:
```rbs
def foo: (&block_type blk) -> void
```
The `block_type` must be a `_ToProc` (Hah! #1223).

Issue #1335, opened 2023-05-16
Untyped block args


soutaro:
`AR::Base.scope` accepts lambdas, but the type of the lambdas is difficult to describe in RBS.

```rbs
class AR::Base
  def self.scope: (Symbol, ^(*untyped, **untyped) -> void) -> void
end
```

The proc type means *it can receive any argument and any keywords*, while what we want is *a proc is required but the type of its arguments is not checked*. So, we want to introduce a new syntax for *untyped args*.

Possible syntax:

```rbs
class AR::Base
  # ...
  def self.scope: (Symbol, ^(...) -> void) -> void
                | (Symbol) { (...) -> void } -> void

  # ?
  def self.scope: (Symbol, ^(?) -> void) -> void
                | (Symbol) { (?) -> void } -> void

  # untyped
  def self.scope: (Symbol, ^untyped -> void) -> void
                | (Symbol) { untyped -> void } -> void
end
```

Any other ones?

ParadoxV5:
Can simply use `Proc` for lambdas; this new syntax mainly benefits untyped blocks

> `untyped -> void`

This suggestion is uncomfortably similar to `(untyped) -> void`. In fact, it conflicts with the possibility of making arguments’ round parentheses optional (if there’s such a possibility).

soutaro:
I'm going to implement it in `(?) -> void` syntax. `...` would be confusing because Ruby has similar syntax with totally different semantics. `untyped -> void` is too similar to normal positional argument.

PR #1806, opened 2024-04-22
Function types with untyped parameter

soutaro:
This PR adds *untyped function parameters*, which means there is no static  type checking on parameters/blocks given to method calls and block yields.

```rbs
class Foo
  def foo: (?) -> void                     # #foo accepts any argument
  def bar: () { (?) -> Integer } -> void   # #bar accepts any block which returns Integer
end
```

See https://github.com/ruby/rbs/issues/1335

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -45,6 +45,7 @@ _literal_ ::= _string-literal_
             | `false`
 
 _proc_ ::= `^` _parameters?_ _self-type-binding?_ _block?_ `->` _type_
+         | `^` `(` `?` `)` _self-type-binding?_ _block?_ `->` _type_      # Proc type with untyped parameter
 ```
 
 ### Class instance type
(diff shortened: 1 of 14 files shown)


ParadoxV5:
The `syntax.md` update hints that this doesn’t resolve #1622

soutaro:
@ParadoxV5 I mean it accepts any blocks too. So `(?) -> T` is equivalent to `(*untyped, **untyped) ?{ (?) -> untyped } -> T` from caller side. I think it improves some problems mentioned at #1622.

ParadoxV5:
Oh, I see. Sorry, it’s late, I’m getting sleepy.

This is the change that confused me:
```bnf
_proc_ ::= `^` _parameters?_ _self-type-binding?_ _block?_ `->` _type_
         | `^` `(` `?` `)` _self-type-binding?_ _block?_ `->` _type_      # Proc type with untyped parameter
```
It appears as if `(?)` doesn’t count the block; but this is the only instance with both `(?)` and `_block?_`, so it’s more likely an oversight.

PR #1810, opened 2024-04-23
Introduce RaaP for testing of signature

ksss:
# Problem

Currently, signature testing is conducted by rbs/unit_test.
However, there are the following issues:

- Many methods exist that have not been tested.
- There is a pattern that has not been tested.
- It takes time to cover the pattern.

# Solution: RaaP

[raap](https://github.com/ksss/raap) is a testing tool for RBS signature.
RaaP can materialize objects and execute methods as long as there is an RBS signature, and it can verify whether the return type matches.

The effectiveness of RaaP can be explained by this PR.

# Take `Set` as an example

This PR covers `Set`.
Many of the `Set` signatures have not been tested and RaaP has been able to find some errors.
The ability to verify many test cases without writing test code is one of the advantages of using RaaP.

### DIFF
--- a/core/set.rbs
+++ b/core/set.rbs
@@ -415,7 +415,7 @@ class Set[unchecked out A]
   # Equivalent to Set#delete_if, but returns nil if no changes were
   # made. Returns an enumerator if no block is given.
   #
-  def reject!: () { (A) -> untyped } -> self
+  def reject!: () { (A) -> untyped } -> self?
 
   # <!--
   #   rdoc-file=lib/set.rb
@@ -437,7 +437,7 @@ class Set[unchecked out A]
   #     Set[1, 2, 3].disjoint? [3, 4]      #=> false
   #     Set[1, 2, 3].disjoint? 4..5        #=> true
   #
-  def disjoint?: (_Each[A]) -> bool
+  def disjoint?: (Set[A] | Enumerable[A]) -> bool
 
   # <!--
   #   rdoc-file=lib/set.rb
@@ -458,8 +458,8 @@ class Set[unchecked out A]
   #
   # Returns an enumerator if no block is given.
   #
-  def divide: () { (A, A) -> untyped } -> Set[self]
-            | () { (A) -> untyped } -> Set[self]
+  def divide: () { (A, A) -> Hash::_Key } -> Set[self]
+            | () { (A) -> Hash::_Key } -> Set[self]
 
   # <!--
   #   rdoc-file=lib/set.rb
@@ -500,7 +500,7 @@ class Set[unchecked out A]
   #     Set[1, 2, 3].intersect? 4..5        #=> false
   #     Set[1, 2, 3].intersect? [3, 4]      #=> true
   #
-  def intersect?: (_Each[A]) -> bool
+  def intersect?: (Set[A] | Enumerable[A]) -> bool
 
   # <!--
   #   rdoc-file=lib/set.rb
@@ -534,7 +534,7 @@ class Set[unchecked out A]
   # Merges the elements of the given enumerable objects to the set and
   # returns self.
   #
-  def merge: (_Each[A]) -> self
+  def merge: (*_Each[A]) -> self
 
   # <!--
   #   rdoc-file=lib/set.rb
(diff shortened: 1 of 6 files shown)


core/set.rbs
```diff
@@ -180,7 +180,7 @@
 #      Resets the internal state; useful if an object
 #      has been modified while an element in the set.
 #
-class Set[unchecked out A]
+class Set[unchecked out A < Object]
```
ParadoxV5, line 183: Does `Set` not support `BasicObject`s, or does RaaP not support them?

core/set.rbs
```diff
@@ -534,7 +534,7 @@ class Set[unchecked out A]
   # Merges the elements of the given enumerable objects to the set and
   # returns self.
   #
-  def merge: (_Each[A]) -> self
+  def merge: (*(Object & _Each[A])) -> self
```
ksss, line 537: Needs `instance_of?`

https://github.com/ruby/set/blob/4e95d55294f0e577529ca8702c4d2c3ae90247e7/lib/set.rb#L597

core/set.rbs
```diff
@@ -572,7 +578,7 @@ class Set[unchecked out A]
   #     set.replace([1, 2])               #=> #<Set: {1, 2}>
   #     set                               #=> #<Set: {1, 2}>
   #
-  def replace: (_Each[A]) -> self
+  def replace: (Object & _Each[A]) -> self
```
ksss, line 581: Needs `instance_of?`

https://github.com/ruby/set/blob/4e95d55294f0e577529ca8702c4d2c3ae90247e7/lib/set.rb#L328

test/raap.txt
```diff
@@ -0,0 +1,5 @@
+# Specify the class/module and method names to be executed by RaaP.
+# By prefixing with `!`, you can skip testing a method.
+# Lines beginning with `#` are comments.
+
+Set[Integer]
```
ksss, line 5: We can add more testing.

ParadoxV5: Ideally, RaaP should be able to automatically test with the given bound (`Object` at this writing), and maybe also a variety of other sample classes.
Then, this file would simply be
```suggestion
Set
```

core/set.rbs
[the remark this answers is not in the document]
ParadoxV5: One of us should rather send a PR to support `BasicObject`.
Else it’d be a sign that `Set` isn’t stable enough for the highly-requested Set literal.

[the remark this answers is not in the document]
ksss: Are you suggesting that we should make modifications to support `BasicObject` in ruby/set as much as possible?
I do not have the motivation to do so myself, but it would certainly be a contribution to the community.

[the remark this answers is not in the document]
ParadoxV5: FWIW… neither does `Array#inspect`
```ruby
[BasicObject.new].inspect #=! undefined method `inspect' for an instance of BasicObject (NoMethodError)
```
For this situation, IIRC @sampersand had the idea of adding a RBS syntax for additional fine-grained bounds on the method level, e.g.,
```rbs
class Set[unchecked out E] # & class Array[unchecked out E]
  …
  def inspect: [E < _Inspect] () -> String
  …
end
```

test/raap.txt
ksss: The minimal condition to satisfy the upper bound is `Object`, but using Object in Set makes it harder to visually interpret the reproducible code, so I use `Integer` instead.

ksss:
I removed `Object &` and made adjustments to minimize changes.

core/set.rbs
```diff
@@ -437,7 +437,7 @@ class Set[unchecked out A]
   #     Set[1, 2, 3].disjoint? [3, 4]      #=> false
   #     Set[1, 2, 3].disjoint? 4..5        #=> true
   #
-  def disjoint?: (_Each[A]) -> bool
+  def disjoint?: (Set[A] | Enumerable[A]) -> bool
```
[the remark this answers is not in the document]
ParadoxV5: Is `Set` `Enumerable`?
```suggestion
  def disjoint?: (Enumerable[A]) -> bool
```

ParadoxV5:
> I removed `Object &` and made adjustments to minimize changes.

🆙. Non critical changes to the RBS are better scoped for another PR (or after [some adjustments to the `.rb` itself](#discussion_r1578872643)

PR #1826, opened 2024-05-24
Use ShellSession syntax highlight in README

ybiquitous:
`sh-session` is supported by GitHub (Linguist):
https://github.com/github-linguist/linguist/blob/e2012cd67867e6287ae5ee96c8fe0b92f9a88efa/lib/linguist/languages.yml#L6615

Compare:

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ gem "rbs"
 
 The gem ships with the `rbs` command line tool to demonstrate what it can do and help develop RBS.
 
-```bash
+```sh-session
 $ rbs version
 $ rbs list
 $ rbs ancestors ::Object
@@ -98,7 +98,7 @@ end
 
 Running prototype on the above will automatically generate
 
-```
+```sh-session
 $ rbs prototype rb person.rb
 class Person
   @name: untyped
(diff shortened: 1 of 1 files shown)


ParadoxV5:
That’s an option? 😮 *takes notes*

ybiquitous:
@ParadoxV5 You mean `sh-session`, right? If so, yes. We can also choose `bash session` or `console` instead of `sh-session`. They are an alias:
https://github.com/github-linguist/linguist/blob/e2012cd67867e6287ae5ee96c8fe0b92f9a88efa/lib/linguist/languages.yml#L6619-L6621

ParadoxV5:
> We can also choose `bash session` or `console` instead of `sh-session`.

`console` could be better because this PR only highlights `rbs` commands and not stuff specific to a certain shell (*nix or not).

ybiquitous:
Okay, I changed it to `console` with e74b2a4367ba9b4ce3c3bf8fec59caf42ab24a1f since I have no strong opinion. 😃

soutaro (approved):
Thanks!

PR #1827, opened 2024-05-24
Add link to gem_rbs_collection to README

ybiquitous:
I tried adding a link to [ruby/gem_rbs_collection](https://github.com/ruby/gem_rbs_collection) to README so that more people can contribute to RBS. Any thoughts?

I think gem_rbs_collection is also an important part of the RBS community, but if you know a more appropriate location, let me know. 🙏🏼


ParadoxV5:
:up: for adding a link because it’s significant, but the reference needs to be worded with clarification.

The preference is that gem maintainers ship RBSs themselves in their own repositories, rather than leaving RBS users to rely on ruby/gem_rbs_collection which’s maintained by third-party volunteers. @pocke thoughts on this ambition?

soutaro:
Merging with minor rephrasing.

PR #1830, opened 2024-05-25
Fix memory leak when ParsingError

ksss:
I have discovered that a memory leak occurs when a `ParsingError` happens. This can occur in the following methods.

- `RBS::Parser.parse_type`
- `RBS::Parser.parse_method_type`
- `RBS::Parser.parse_signature`

If a global exit occurs before `free_parser` is called, `free_parser` may not be called at all.

I have fixed this issue by using `rb_ensure` to ensure that `free_parser` is called under all circumstances.

```console
$ cat ensure.rb
#! /usr/bin/env ruby

require 'rbs'

Thread.new {
  while true
    mem = `ps -o rss= -p #{Process.pid}`.to_i
    puts mem
    sleep 1
  end
}

while true
  begin
    RBS::Parser.parse_signature("c")
  rescue RBS::ParsingError
  end
end

# before
$ bundle exec ruby -I lib ensure.rb
37200
230096
421920
614544
806752
995488
1185664
1376944
1567616
1755488

# after
$ bundle exec ruby -I lib ensure.rb
33024
33152
33152
33184
33280
33328
33328
```

### DIFF
--- a/ext/rbs_extension/parser.c
+++ b/ext/rbs_extension/parser.c
@@ -2805,56 +2805,84 @@ VALUE parse_signature(parserstate *state) {
   return ret;
 }
 
+struct parse_type_arg {
+  parserstate *parser;
+  VALUE require_eof;
+};
+
+static VALUE
+ensure_free_parser(VALUE parser) {
+  free_parser((parserstate *)parser);
+  return Qnil;
+}
+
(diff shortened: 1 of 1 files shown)


soutaro (approved):
💯

PR #1833, opened 2024-05-30
Fixup dependabot

soutaro:


ParadoxV5:
I’m concerned that this will automatically accept breaking changes even if they fail CI.

Also, if the intent is to automatically bump teeny versions and transitive deps for `Gemfile.lock`, I recommend simply removing it from the repo, so a fresh checkout/clean (including CIs) will force `bundle` to look up the latest versions as part of `bundle install`.

soutaro:
No. This won't merge PRs when CI fails.

Dependabot won't change gemspec files. So the potential minor unstableness increase for the dev team is acceptable I think. 

I don't think removing Gemfile.lock from the repo is a good idea. My experience was it caused more problems -- CI starts failing randomly if a release of a gem causes an issue.

ParadoxV5:
> No. This won't merge PRs when CI fails.

I just noted that `rbs` uses Merge Queue, which can be configured to reject failing PRs. However, it doesn’t look like `steep` and `rbs-inline` are also using the queue.

> Dependabot won't change gemspec files.

Dependabot *can* update `.gemspec`s. E.g.,
* ParadoxV5/template-ruby-gem#3

> I don't think removing Gemfile.lock from the repo is a good idea. My experience was it caused more problems -- CI starts failing randomly if a release of a gem causes an issue.

We can always pin specific versions or adjust the bounds (lower or upper) to exclude problematic versions.
In fact, *if* we need to do that, we shouldn’t specify the constraints just in `Gemfile.lock`, but rather `.gemspec` so they’re included in releases.

soutaro:
Oops. My bad. We need a config not to update the .gemspec. 

It specifies --auto option with gh merge command. It enables auto-merge, and CI fails reject merging the PRs.

ParadoxV5:
> Oops. My bad. We need a config not to update the .gemspec.

Why 😕?

> It specifies --auto option with gh merge command. It enables auto-merge, and CI fails reject merging the PRs.

👍

soutaro:
> we shouldn’t specify the constraints just in Gemfile.lock

It may be true in theory, but practically I don't want to compose PRs to pin gems in Gemfile. If this is only for the developments, because this is a gem, I believe having the lockfile is a reasonable balance.

soutaro:
> Why 😕?

Hmm. What I want exactly is:

1. Merge automatically updating gems in the lockfile, which is used for developments.
2. Merge after reviewing if it changes .gemspec dependency, because it is an update for releases.

Maybe we can do it...

soutaro:
The gemspec doesn't have any dependency with version constraints!

ParadoxV5:
> 2. Merge after reviewing if it changes .gemspec dependency, because it is an update for releases.

With Dependabot PRs automatically merging, we can easily overlook them even if they affect the `.gemspec`, especially for `steep`’s large assortment of dependencies.
The thing is, isn’t this _a_ reason why Dependabot opens PRs – to give maintainers opportunity to review?

> The gemspec doesn't have any dependency with version constraints!

For `rbs`, at least. Could this also be intended? 🤔

ParadoxV5:
@pocke any comments?

PR #1857, opened 2024-06-02
stdlib: Add types for kconv

tk0miya:


stdlib/kconv/0/kconv.rbs
```diff
@@ -0,0 +1,166 @@
...
+  #
+  UTF8: Encoding
+
+  # <!--
+  #   rdoc-file=ext/nkf/lib/kconv.rb
+  #   - Kconv.guess(str)   => encoding
+  # -->
+  # Guess input encoding by NKF.guess
+  #
+  def self.guess: (String str) -> Encoding?
```
ksss, line 66: I couldn't find any cases where `::Kconv.guess` returns `nil`. Wouldn't `Encoding` be sufficient?

tk0miya: Sorry, I'm not sure. Japanese Ruby Reference manual says it may return nil.
https://docs.ruby-lang.org/ja/latest/class/Kconv.html#M_GUESS

But the English version says it always returns an Encoding object.
https://docs.ruby-lang.org/en/3.3/Kconv.html#method-i-guess

I'll take a look the implementation later.

tk0miya: Hmm... it seems `Kconv.guess` always returns an Eencoding object, right? I'll fix the definition and the documentation tomorrow.

https://github.com/ruby/nkf/blob/master/lib/kconv.rb#L141
https://github.com/ruby/nkf/blob/master/ext/nkf/nkf.c#L191

Thank you for reviewing!

tk0miya: Fixed.

soutaro (approved):
Thanks!

PR #1858, opened 2024-06-02
stdlib: Add types for URI::MailTo

tk0miya:


stdlib/uri/0/mailto.rbs
```diff
@@ -4,5 +4,89 @@ module URI
   # RFC6068, the mailto URL scheme.
   #
   class MailTo < Generic
+    EMAIL_REGEXP: Regexp
```
tk0miya, line 7: This constant is not documented in the official document.
https://docs.ruby-lang.org/en/3.3/URI/MailTo.html

But it's well used for verifying the format of email addresses.
For example, the Rails Guide mentions this constant in implementing validators on Rails.
https://guides.rubyonrails.org/active_record_validations.html#custom-validators

stdlib/uri/0/mailto.rbs
```diff
@@ -4,5 +4,89 @@ module URI
...
+    #     m1 = URI::MailTo.build(['joe@example.com', 'subject=Ruby'])
+    #     m1.to_s  # => "mailto:joe@example.com?subject=Ruby"
+    #
+    #     m2 = URI::MailTo.build(['john@example.com', [['Subject', 'Ruby'], ['Cc', 'jack@example.com']]])
+    #     m2.to_s  # => "mailto:john@example.com?Subject=Ruby&Cc=jack@example.com"
+    #
+    #     m3 = URI::MailTo.build({:to => 'listman@example.com', :headers => [['subject', 'subscribe']]})
+    #     m3.to_s  # => "mailto:listman@example.com?subject=subscribe"
+    #
+    def self.build: (Array[String]) -> instance
```
tk0miya, line 39: The official document says `URI::MailTo.build` takes a tuple of strings. So it would be correct and accurate to use `([String, String]) -> instance` instead of this line. But it's not good for use because Steep recognizes the type of tuple literal of strings as an `Array[String]`.

stdlib/uri/0/mailto.rbs
```diff
@@ -4,5 +4,89 @@ module URI
...
+                  | ([String, Array[Array[String]]]) -> instance
+                  | (Hash[Symbol, String | Array[Array[String]]]) -> instance
```
tk0miya, line 41: This would also be correct and accurate to use a record type `({ to: String, headers: String | Array[Array[String]] }) -> instance` instead. But it's not good for use too.

stdlib/uri/0/mailto.rbs
```diff
@@ -4,5 +4,89 @@ module URI
...
+
+    # <!-- rdoc-file=lib/uri/mailto.rb -->
+    # E-mail headers set by the URL, as an Array of Arrays.
+    #
+    def headers: () -> Array[[String, String]]
```
ksss, line 46: I found invalid pattern by raap.

``````
$ bundle exec raap -r uri --library uri URI::MailTo#headers
[RaaP] INFO: # URI::MailTo
[RaaP] INFO: ## def headers: () -> ::Array[[ ::String, ::String ]]
EEF

() -> Array[[String, String]]

[RaaP] INFO: success: 0, skip: 0, exception: 2, time: 15ms
# 1) Failure:

def headers: () -> ::Array[[ ::String, ::String ]]
  in /Users/ksss/src/github.com/ksss/rbs/stdlib/uri/0/mailto.rbs:46:17...46:46

## Reason

Failed in case of `#<URI::MailTo>.headers() -> [["ez"]][Array[Array[String]]]`

### Repro

```rb
uri_mailto = URI::MailTo.new(nil, nil, 'lm', -4, nil, nil, nil, 'ez', nil)
uri_mailto.headers()
```
``````

Comment (soutaro) [reply stdlib/uri/0/mailto.rbs:46]: Looks like the `query` (`?ez`) cannot be constructed with `URI::MailTo.build`. 

```
irb(main):020> URI::MailTo.build(["hoge@huga", "foo"]).headers
/Users/soutaro/.rbenv/versions/3.3.2/lib/ruby/3.3.0/uri/mailto.rb:212:in `check_headers': bad component(expected opaque component): foo (URI::InvalidComponentError)
```

I think we can have `Array[[String, String]]` here, even raap detects an error.

soutaro (approved):
Thanks!

tk0miya:
Thank you for your review and merge!

PR #1864, opened 2024-06-05
stdlib: the headers keyword for CSV.foreach can take String

tk0miya:
The headers keyword option of `CSV.foreach` can take String, Symbol and Array too.  Therefore, this adds them to the type definition.

refs:
* https://docs.ruby-lang.org/en/3.3/CSV.html#method-c-foreach
* https://docs.ruby-lang.org/en/3.3/CSV.html#class-CSV-label-Options+for+Parsing

Note:

In https://github.com/ruby/rbs/pull/1738/files#r1468802569, @m11o was aware of the lackness of the types, but they determined to support only "true" value at first.


soutaro (approved):
:+1:

PR #1866, opened 2024-06-05
Drop dependency on `abbrev`

Earlopain:
Pulling in something extra just for this seems pretty wasteful in my opinion

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,6 @@ PATH
   remote: .
   specs:
     rbs (3.5.0.pre.2)
-      abbrev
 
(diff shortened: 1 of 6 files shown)


soutaro:
I'm open to this change. It implements the same feature after dropping `abbrev`.
Just curious if there is some real problem installing the library?

Earlopain:
No real problem, I just like having a small dependency tree if possible. Abbrev isn't buildin anymore, if it did something more complicated there could be an argument for it but it really isn't doing much as it is used here.

soutaro (approved):
👍

PR #1867, opened 2024-06-05
Don't try to sign git commits when running tests

Earlopain:
I have git set up to always sign with a key that requires a password, this would require me to enter that password when running tests.

Git allows overwriting config for a single invocation, lets just do that.


soutaro:
@Earlopain Do you think moving the config to the newly created git repository makes more sense?

Earlopain:
Yes, that totally makes sense. I'll change this over in a bit

soutaro (approved):
Thanks! 👏

PR #1868, opened 2024-06-06
Revise `#divmod` method so that `sleep` can accept `Float` and `Rational`

soutaro:
This PR also adds `typecheck_test` rake task. Some complicated type definition causes unexpected type errors that cannot be detected using runtime tests.


soutaro:
The added `typecheck_test` job is not *required*, because there may be a situation that cannot be successful -- when `rbs` gem is updated and `steep` needs update too.

PR #1903, opened 2024-06-21
stdlib: Add type for CSV::Row#initialize

aki19035vc:
Add type to CSV::Row#initialize in stdlib because the type does not exist.

The content of the headers array is `untyped`, because any object can be specified as long as it is an Ruby object.

I placed the test files directly under the `stdlib` directory, but please let me know if you need to create more directories like `stdlib/csv`!


soutaro (approved):
👍

Issue #420, opened 2020-10-16
_Callable interface in builtin


HoneyryderChuck:
I was looking at https://github.com/ruby/rbs/blob/c64d1f5ba5a4fbc95ffb1d72d783ee39b17d60fe/stdlib/builtin/builtin.rbs, where a lot of common knownn interfaces are stored. I was wondering whether there should be one for "callable" objects, that is: implements `#call`. Objects implementing this interface can already be passed to places where are expected, for example.

```ruby
interface _Callable
  def call: (*untyped) -> void
  alias [] call
end

**Comment by marcandre** [comment]
Maybe, but very few builtin methods take such a parameter. I can think of `Enumerable#find` (which is currently badly defined)

Note that the return type should not be `void` but `top`. Also the interface should not depend on `[]`:

```sh
interface _Call
  def call: (*top) -> top
end
```

Still, it might be best not to, since the above interface is too lax, usually you will want to specify the type of inputs the `call` method should be expecting.

HoneyryderChuck:
In the benefit of understanding, what does "top" stand for? I struggled a bit with the meaning of it.

Perhaps you're right, but a Callable with a T reference seems more suitable to my example. All things considered I might backtrack a bit on this until I come up with compelling use cases.

soutaro:
If you have a specific method in your mind, like `Enumerable#find`, defining an interface would be better.

```rbs
interface _NotFound[A]
  def call: () -> A
end

module Enumerable[X, Y]
  def find: [T] (_NotFound[T]) -> (X | T)
  ...
end
```

HoneyryderChuck:
I had this in mind: https://docs.python.org/3/library/typing.html#callable

HoneyryderChuck:
After thinking more about this, I think that I do not necessarily want a Callable interface, rather an "inline" way of saying "something implementing #call", i.e. I think that rbs could have syntactic sugar for defining these 1 method-interfaces.

```
  def callables: () -> Array[Imp[call:(*untyped) -> top]]
```

How that would look like, I have no idea :)

PR #1904, opened 2024-06-21
[Enumerator#initialize] support `nil`, `#call` and `#to_int`

ksss:
Argument of `Enumerator#initialize` allows `nil`, `#call` and `#to_int` object.
Included `Float::INFINITY` in `#to_int`.

https://github.com/ruby/ruby/blob/ac9e84df3d0b06e62498aafbca99e8f8475ec142/enumerator.c#L437-L451

### DIFF
--- a/core/enumerator.rbs
+++ b/core/enumerator.rbs
@@ -269,6 +269,10 @@ class Enumerator[unchecked out Elem, out Return] < Object
   #
   def feed: (Elem arg0) -> NilClass
 
+  interface _Call
+    def call: () -> untyped
+  end
+
   # <!--
   #   rdoc-file=enumerator.c
   #   - Enumerator.new(size = nil) { |yielder| ... }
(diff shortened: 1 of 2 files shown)


core/enumerator.rbs
```diff
@@ -293,7 +297,7 @@ class Enumerator[unchecked out Elem, out Return] < Object
   # lazy fashion (see Enumerator#size). It can either be a value or a callable
   # object.
   #
-  def initialize: (?Integer arg0) { (Enumerator::Yielder arg0) -> Return } -> void
+  def initialize: (?(nil | _Call | ::_ToInt) arg0) { (Enumerator::Yielder arg0) -> Return } -> void
```
ParadoxV5, line 300:
```suggestion
  def initialize: (?(int? | _Call) size) { (Yielder yielder) -> Return } -> void
```

core/enumerator.rbs
```diff
@@ -269,6 +269,10 @@ class Enumerator[unchecked out Elem, out Return] < Object
   #
   def feed: (Elem arg0) -> NilClass
 
+  interface _Call
+    def call: () -> untyped
+  end
```
ParadoxV5, line 274:
```ruby
Enumerator.new(->{'hi'}) {}.size #=> "hi"
```
```suggestion
  interface _Size
    def call: () -> (Integer | Float)?
  end
```
Relevant: https://github.com/ruby/rbs/issues/420#issuecomment-710768398

PR #1909, opened 2024-06-25
Fix double-splat method

ksss:
Currently `sleep(2 ** 2)` results in a type error since `2 ** 2` returns `Numeric` type.
Because of these problems, it does not seem very convenient to set the return value to `Numeric` in numerical operations.

In this PR, I have made similar fixes to `#**` as those in [Integer#pow](https://github.com/ruby/rbs/pull/1706).

It has been exhaustively tested in raap because the combination is complicated.


ParadoxV5:
* Resolves https://ptb.discord.com/channels/518658712081268738/1143338778384146574/1252966446221492256

+1 for the good intention, but we should keep the `(Numeric) ->` branch as math operators accept any Numeric that `coerce` appropriately.

For the intention, we should go through __all__ files and check for `Numeric` usages and specify known cases as appropriate.
Summon: @sampersand: you had PRs on numerics and they include making the sigs more ~~pedantic~~ _exact_.

ksss:
@ParadoxV5 
Thanks for the review. Indeed, with this change `1 ** BigDecimal(1)` would result in a type error.
What about the idea of extending `Integer#**` from BigDecimal?

```rbs
# bigdecimal.rbs
class Integer
  def **: (BigDecimal) -> BigDecimal
        | ...
end
```

sampersand:
> * Resolves https://ptb.discord.com/channels/518658712081268738/1143338778384146574/1252966446221492256
> 
> 
> +1 for the good intention, but we should keep the `(Numeric) ->` branch as math operators accept any Numeric that `coerce` appropriately.
> 
> For the intention, we should go through **all** files and check for `Numeric` usages and specify known cases as appropriate. Summon: @sampersand: you had PRs on numerics and they include making the sigs more ~pedantic~ _exact_.

I hadn't gotten to around to doing `integer.rbs`, but maybe i should!

PR #1911, opened 2024-06-25
stdlib/zlib: Add types for singleton methods of GzipReader

aki19035vc:
Add types for singleton methods of GzipReader.

Also, fix the incorrect types.


soutaro (approved):
Thanks! 🙏

PR #1914, opened 2024-06-28
[prototype] Remove redundant access modifier

ksss:
Method visibility is public by default.
`public` is always redundant because the order in which methods appear is always guaranteed at `rbs prototype runtime`.


soutaro (approved):
👍

PR #1920, opened 2024-07-05
Update Integer

sampersand:
- part 1, finished allbut coercion
- i think i got em
- everything but two methods
- more tests
- more tests, now just for coerce and enumerators
- partx?
- x

### DIFF
--- a/core/ruby_vm.rbs
+++ b/core/ruby_vm.rbs
@@ -260,118 +260,118 @@ module RubyVM::AbstractSyntaxTree
     #
     def children: () -> Array[untyped]
   end
+end
 
-  # <!-- rdoc-file=yjit.rb -->
-  # This module allows for introspection of YJIT, CRuby's just-in-time compiler.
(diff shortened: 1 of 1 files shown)


ParadoxV5:
Welcome back, brother.

core/integer.rbs
```diff
@@ -1263,43 +1214,7 @@ class Integer < Numeric
...
-          | (27) -> String
-          | (28) -> String
-          | (29) -> String
-          | (30) -> String
-          | (31) -> String
-          | (32) -> String
-          | (33) -> String
-          | (34) -> String
-          | (35) -> String
-          | (36) -> String
```
ParadoxV5, line 1301: These were probably to represent that `integer.to_s(69)` is an error.
But this error is literally not a type error:
```ruby
42.to_s 'a' # TypeError
42.to_s 69  # ArgumentError
```
Not to mention how’d this error arise in production…

Issue #1822, opened 2024-05-20
Support type narrowing


mtsmfm:
Ref: https://github.com/soutaro/steep/issues/472

Currently RBS doesn't have any syntax to narrow types in control flow.

So there's no way to achieve the following:

```ruby
a = [1, nil].sample # a is `Integer | nil`
unless a.nil?
  # a is `Integer` here
  p a + 1
end
```

But actually, both Sorbet and Steep treat some special methods to narrow types

https://github.com/sorbet/sorbet/blob/718bc64f895abeeff88f56efbc92a3554ffaa4f2/infer/environment.cc#L536-L545

https://github.com/soutaro/steep/blob/a868762c2bd09f0954b05cdd9eab1819e14a51d3/lib/steep/interface/builder.rb#L709-L721

Ideally, I think we need to have a syntax considering Ruby gems which extend `Object` like ActiveSupport's `present?`.

I'm not very familiar with type system though, I guess the syntax would be similar with TypeScript's type predicates

https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

```rbs
class Object[T?]
  def nil?: () -> self is nil
end
```

```rbs
class Object[T?]
  def present?: () -> self is T
end
```

ParadoxV5:
:up: for *something* (not necessarily this)

From https://github.com/soutaro/steep/pull/905#issuecomment-1977213468, the current solution is
```rbs
class Object
  def nil?: () -> bool # cannot be assertively `false` or else `NilClass#nil?` breaks polymorphism of `NilClass < Object`
end
class NilClass
  def nil?: () -> true
end
```
Or, with https://github.com/ruby/rbs/discussions/1639#discussioncomment-7648121,
```rbs
class Object
  def nil?: [self < nil] () -> true
          | () -> bool
end
```
But in either case, the prickly part is that there’s no format to tell that non-`nil`s’ `#nil?` always return `false`. In pseudocode, we need:
```rbs
class Object
  def nil?: [self < nil] () -> true
          | [self !< nil] () -> false # ⬅ a way to declare a negative constraint
end
```
Alternatively, if the overload semantics changes so that earlier branches subtractively narrow latter branches,
```rbs
class Object
  def nil?: [self < nil] () -> true
          | () -> false # `else`
end
```

ParadoxV5:
> https://docs.python.org/3.13/library/typing.html#typing.TypeIs
[...]

PR #1921, opened 2024-07-05
Type signature updates for Enumerable, Module, Range, Array

hjwylde:
These are some type signature updates, mostly following on from the discussion here: https://github.com/ruby/rbs/pull/1226, but also some additional ones on top of it.

I've made the following changes:
* `Enumerable` - added a more specific type override for `enum_for`/`to_enum` for the default (`:each`/no-arg) case; this override means that we know the `Enumerator` returned will be iterating over elements of type `Elem`. I think that this is a sensible override to provide because of the type from `_Each`. I wasn't sure if the overrides should go here or in the `_Each` interface, but it felt strange putting more things into `_Each` when it's currently clean and simple.

* `Module` - improved the type signature for `class_exec` to be similar to `instance_exec`. Specifically it now shows that arguments passed are forwarded to the block, and that the block has a `self` type based on the receiver.

* `Range` - added a type signature for `%` as it was missing. This one is a copy/paste from `step`, minus the optional parameters.

* `Array` - added some more `%a{implicitly-returns-nil}` annotations. I've only added them in locations that looked similar to the original for `[]`, where they may return `nil` if an index is out of bounds or the array is empty, as this is typically knowledge that is checked beforehand. Happy to reduce the number added though if it is too many. I haven't added any to `Enumerable` as there is no size/length/empty methods, so `Hash#first` will not have this annotation.

Let me know if there's anything I missed or items you'd like me to change.

### DIFF
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -1082,7 +1082,7 @@ class Array[unchecked out Elem] < Object
   #     a.at(0) # => :foo
   #     a.at(2) # => 2
   #
-  def at: (int index) -> Elem?
+  def at: %a{implicitly-returns-nil} (int index) -> Elem
 
   # <!--
   #   rdoc-file=array.c
@@ -1378,7 +1378,7 @@ class Array[unchecked out Elem] < Object
   #
   # If `index` is too small (far from zero), returns nil.
   #
-  def delete_at: (int index) -> Elem?
+  def delete_at: %a{implicitly-returns-nil} (int index) -> Elem
 
   # <!--
   #   rdoc-file=array.c
@@ -1942,7 +1942,7 @@ class Array[unchecked out Elem] < Object
   #
   # Related: #last.
   #
-  def first: () -> Elem?
+  def first: %a{implicitly-returns-nil} () -> Elem
            | (int n) -> ::Array[Elem]
 
   # <!--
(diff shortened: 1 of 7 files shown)


ParadoxV5:
> I wasn't sure if the overrides should go here or in the `_Each` interface, but it felt strange putting more things into `_Each` when it's currently clean and simple.

The purpose of `_Each` (and #424 too) was to identify that a type has the method `#each`.
It is not its concern that `enum_for`/`to_enum` has a narrower type, but rather `Kernel`’s own responsibility to type-narrow (#1822).

hjwylde:
> The purpose of _Each (and https://github.com/ruby/rbs/issues/424 too) was to identify that a type has the method #each.

I agree, better to keep it as such.

> It is not its concern that enum_for/to_enum has a narrower type, but rather Kernel’s own responsibility to type-narrow (https://github.com/ruby/rbs/issues/1822).

I believe that method contracts do have a purpose in signatures and would be of help, but currently they do not exist and it looks like it's still under discussion. In the meantime, I believe this has value as it is still a standard practice for a subclass in OO languages to provide narrower types. The commit regarding `to_enum`/`enum_for` could always be reverted if #1822 is resolved.

As a side note, if there are only concerns with the `to_enum`/`enum_for` updates, I can drop that commit so that the PR can just address the other changes (to `Range`, `Array`, `Module`)?

soutaro (approved):
Thanks! Let's try with this version, which makes a deadline to support it in Steep. 😅

hjwylde:
Thanks! Yeah let's see how it goes :), no concerns from me if it needs to be changed/reverted at some point.

PR #1922, opened 2024-07-08
Updated Integer

sampersand:


core/integer.rbs
```diff
@@ -986,15 +967,13 @@ class Integer < Numeric
   #
   def nobits?: (int mask) -> bool
 
-  def nonzero?: () -> self?
```
sampersand, line 989: Oddly enough, `nonzero?` isn't defined on integers, but `zero?` is

core/integer.rbs
```diff
@@ -1047,28 +1019,14 @@ class Integer < Numeric
...
-  def rationalize: (?Numeric eps) -> Rational
-
-  def real: () -> self
-
-  def real?: () -> true
-
-  def rect: () -> [ Integer, Numeric ]
-
-  alias rectangular rect
+  def rationalize: (?untyped _ignored_eps) -> Rational
```
sampersand, line 1029: Wasn't sure if we wanted to keep this `Numeric _ignored_eps` or `untyped`.

core/integer.rbs
```diff
@@ -1263,43 +1207,7 @@ class Integer < Numeric
...
-          | (27) -> String
-          | (28) -> String
-          | (29) -> String
-          | (30) -> String
-          | (31) -> String
-          | (32) -> String
-          | (33) -> String
-          | (34) -> String
-          | (35) -> String
-          | (36) -> String
```
sampersand, line 1301: Like @ParadoxV5 said, presumably these cases were for documentation about which int literals can be given, but they're covered under `int`, as well arent even `TypeError`s

core/integer.rbs
```diff
@@ -1382,4 +1292,92 @@ class Integer < Numeric
...
+          | (to: Numeric, ?by: Integer) { (Integer) -> void } -> void
+          | (by: Numeric, ?to: Numeric) { (Numeric) -> void } -> void
+          | () -> Enumerator[Integer, bot]
+          | (Numeric limit, ?Integer step) -> Enumerator[Integer, void]
+          | (Numeric limit, ?Numeric step) -> Enumerator[Numeric, void]
+          | (to: Numeric, ?by: Integer) -> Enumerator[Integer, void]
+          | (by: Numeric, ?to: Numeric) -> Enumerator[Numeric, void]
+
+  %a{steep:deprecated}
+  def to_c: () -> Complex
```
sampersand, line 1382: None of these are defined on `Integer`, but instead on `Numeric`

test/stdlib/Integer_test.rb
```diff
@@ -1,398 +1,811 @@
...
+end
+
+class IntegerInstanceTest < Test::Unit::TestCase
+  include TestHelper
 
-  def test_angle
-    3.angle()
+  testing "::Integer"
+
+  def with_random_Integers
```
sampersand, line 33: Instead of using hardcoded integers, I figured generally using a range of integers would cover more cases.

core/integer.rbs
ParadoxV5: https://github.com/ruby/rbs/pull/1920#discussion_r1666266654

Issue #1875, opened 2024-06-07
Method overload narrowing


ParadoxV5:
I request a notation for method overload narrowing, such that subtype args narrow supertype args.
The intent is so the return type of supertype overloads not longer need to explicitly include those of narrower overloads.

E.g., `TrueClass#&` example from [prior brief discussion on Discord](https://ptb.discord.com/channels/518658712081268738/1143338778384146574/1185012559493992448):
```rbs
def &: (false? rhs) -> false
     | (untyped rhs) -> true
```
Contrast with our current compromise:
https://github.com/ruby/rbs/blob/45b34bd9bf9a2b6872ae832f6a344b23dc75ec3d/core/true_class.rbs#L28-L29
This new solution enables us to type the result for truthy args without needing a “difference type” `type truthy = top - false?`.
In fact, this is possibly the only practical use case for “difference types”.

I’ve considered moving this focus to Steep but decided that the entire RBS ecosystem could use this.
For example, https://github.com/ruby/rbs/issues/1874#issuecomment-2155458884 finally encouraged me to formally file this issue.

Though it is backward-incompatible to change the semantics of `|` set by yore, we can instead introduce an new, alternative syntax, such as the tokens:
* [ ] `&` (from #566)
* [ ] `:` (from the aformentioned Discord discussion)
* [ ] `||`

---

* Relates to, but does not resolve: #1822
* Summon: @sampersand (replied to me in the aformentioned Discord)

PR #1930, opened 2024-07-13
Update Symbol

sampersand:
This pr updates `Symbol`.

There wasn't actually a ton to do; it was mostly marking things as aliases, and copying signatures over from `String` (eg for `Symbol#[]`).

### DIFF
--- a/core/symbol.rbs
+++ b/core/symbol.rbs
@@ -151,7 +151,7 @@ class Symbol
   # Related: String#<=>.
   #
   def <=>: (Symbol object) -> (-1 | 0 | 1)
-         | (untyped) -> nil
+         | (untyped) -> (-1 | 0 | 1)?
 
   # <!--
   #   rdoc-file=string.c
@@ -236,8 +236,7 @@ class Symbol
   #
   # Related: Symbol#casecmp?, String#casecmp.
   #
-  def casecmp: (Symbol object) -> (-1 | 0 | 1)?
-             | (untyped) -> nil
+  def casecmp: (untyped object) -> (-1 | 0 | 1)?
 
   # <!--
   #   rdoc-file=string.c
@@ -270,8 +269,7 @@ class Symbol
   #
   # Related: Symbol#casecmp, String#casecmp?.
   #
-  def casecmp?: (Symbol object) -> bool?
-              | (untyped) -> nil
+  def casecmp?: (untyped object) -> bool?
 
   # <!--
   #   rdoc-file=string.c
--- a/core/symbol.rbs
+++ b/core/symbol.rbs
@@ -131,9 +131,7 @@ class Symbol
   #     Symbol.all_symbols.size    # => 9334
   #     Symbol.all_symbols.take(3) # => [:!, :"\"", :"#"]
   #
-  def self.all_symbols: () -> ::Array[Symbol]
-
-  public
+  def self.all_symbols: () -> Array[Symbol]
 
   # <!--
   #   rdoc-file=string.c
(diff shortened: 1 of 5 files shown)


core/string.rbs
```diff
@@ -897,10 +897,10 @@ class String
   #     number #=> "9"
   #
   def =~: (Regexp regex) -> Integer?
-        | [T] (_MatchAgainst[self, T] object) -> T
+        | [T] (_MatchAgainst[T] object) -> T
 
-  interface _MatchAgainst[O, T]
```
sampersand, line 902: It didnt make much sense to have an `O` field here, as it's only ever being matched against `String`s, so I've removed the field. However, it is a breaking change, so we might want to hold off on this

core/symbol.rbs
```diff
@@ -469,6 +469,4 @@ class Symbol
             | (:ascii | :lithuanian | :turkic) -> Symbol
             | (:lithuanian, :turkic) -> Symbol
             | (:turkic, :lithuanian) -> Symbol
-
-  def clone: (?freeze: true?) -> self
```
sampersand, line 473: Not actually defined in `Symbol`

core/string.rbs
ParadoxV5: Changing from `O = self` to `String` loses type narrowing.

Having a generic `#=~` (with both LHS and RHS type variables) is also fine for #1638.

sampersand: Resolved by reverting this

core/symbol.rbs
```diff
@@ -152,21 +150,21 @@ class Symbol
   #
   # Related: String#<=>.
   #
-  def <=>: (Symbol other) -> Integer
-         | (untyped other) -> Integer?
+  def <=>: (Symbol object) -> (-1 | 0 | 1)
+         | (untyped) -> nil
```
soutaro, line 154:
```suggestion
         | (untyped) -> Integer?
```

The overload may be applied to a `Symbol` object too.

core/symbol.rbs
```diff
@@ -240,7 +236,8 @@ class Symbol
   #
   # Related: Symbol#casecmp?, String#casecmp.
   #
-  def casecmp: (untyped other) -> Integer?
+  def casecmp: (Symbol object) -> (-1 | 0 | 1)?
+             | (untyped) -> nil
```
soutaro, line 240:
```suggestion
             | (untyped) -> Integer?
```

core/symbol.rbs
```diff
@@ -273,7 +270,8 @@ class Symbol
   #
   # Related: Symbol#casecmp, String#casecmp?.
   #
-  def casecmp?: (untyped other) -> bool?
+  def casecmp?: (Symbol object) -> bool?
+              | (untyped) -> nil
```
soutaro, line 274:
```suggestion
              | (untyped) -> bool?
```

soutaro: We can revert the change for this method because they are identical.

soutaro (changes requested):
The `untyped` version should return for the case of the argument is a `Symbol`.

```rb
a = [1, :foo, "bar"].sample     # => a: Integer | Symbol | String | nil

:foo <=> a         # untyped overload is selected, but the actual value may be a Symbol
```

core/symbol.rbs
```diff
@@ -176,7 +174,7 @@ class Symbol
   # variables; see String#=~.
   #
   def =~: (Regexp regex) -> Integer?
-        | [T] (String::_MatchAgainst[self, T] object) -> T
+        | [T] (String::_MatchAgainst[String, T] object) -> T
```
soutaro, line 177: [Optional] Wondering if you can add a test to confirm the `_MatchAgains` generic interface works correctly with Steep. (in `test/typecheck/_MatchAgainst` or somewhere.)

core/symbol.rbs
ParadoxV5: * #1875

core/symbol.rbs
sampersand: Do we have any of these as examples to base it off of?

soutaro (approved):
👍

soutaro:
@sampersand Added a few tests with 97e5fbc.

PR #1931, opened 2024-07-14
Update Refinement

sampersand:
This typechecks `Refinement`.

It only has two methods (`target` and a private `import_methods`; `refined_class` is deprecated in Ruby 3.4 and identical to `target`, so I've removed it). `target` is easy to typecheck, but due to a bug in ruby, `import_methods` has to be handled specially (calling it outside of a `refine` block, via `send`, segfaults).


soutaro (approved):
Thanks! 🎉

PR #1939, opened 2024-07-26
Update Method

sampersand:
This PR updates `Method`.

Nothing too crazy going on here.

### DIFF
--- a/core/method.rbs
+++ b/core/method.rbs
@@ -22,8 +22,99 @@
 #     %w[2017-03-01 2017-03-02].collect(&Date.method(:parse))
 #     #=> [#<Date: 2017-03-01 ((2457814j,0s,0n),+0s,2299161j)>, #<Date: 2017-03-02 ((2457815j,0s,0n),+0s,2299161j)>]
 #
-class Method < Object
+class Method
+  # The return type from `#parameters` methods (such as those defined on `Method`, `Proc`, and `UnboundMethod`).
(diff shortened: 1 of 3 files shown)


core/method.rbs
```diff
@@ -161,7 +168,7 @@ class Method < Object
   #     m.call # => "bar"
   #     n = m.clone.call # => "bar"
   #
-  def clone: () -> Method
+  def clone: () -> instance
```
sampersand, line 171: `clone`, like `dup`, technically return `instance`s, but there's no actual way to test this (you can't create subclasses of `Method` to call `#clone` on)

lib/rbs/test/type_check.rb
```diff
@@ -80,7 +80,7 @@ def method_call(method_name, method_type, call, errors:)
             # Block is given, but not yielded
           end
         else
-          if call.block_given
+          if call.block_given && !method_type.type.is_a?(Types::UntypedFunction)
```
sampersand, line 83: This ensures that `UntypedFunction`s in tests can take blocks

test/stdlib/Method_test.rb
```diff
@@ -1,89 +1,181 @@
...
+    assert_send_type  '() -> Integer',
+                      METHOD, :hash
   end
 
-  def test_call
-    Foo.new.method(:foo).call
-    Foo.new.method(:foo_with_args).call(1)
-  end
+  def test_dup
+    omit_if RUBY_VERSION < '3.4'
```
sampersand, line 31: `Method#dup` was only added in ruby 3.4; should we do an `%a{rbs:since:3.4}` to the definition? probably not.

PR #1936, opened 2024-07-20
Update global_variables.rbs

sampersand:
This PR updates `global_variables.rbs`.

This PR requires an `alias` to be added for global variables (which'll be useful later on when we typecheck `English`)

Issue #1647, opened 2023-11-24
Add support for `alias $x $y`


sampersand:
(Originally a discussion, #1611, I think this is more apt as an issue).

One of the lesser-known facts of ruby is you can also use `alias` to alias global variables to one another, such as `alias $LOAD_PATH $:`. 

I propose we support this syntax in RBS (it'd be useful for `require 'English'`). It's very, _very_ niche, but should be supported eventually for completeness.

sampersand:
Shunt to the back of the backlog?

soutaro:
I understand we need it for completeness, but I don't think it's a priority because it's niche...

I'm not against implementing it if someone else can send a PR. The outline would be:

1. Add a syntax `AST::Declarations::GlobalAlias` or something?
2. Implement parser
3. Add a method to resolve type of global to `Environment#global_type: (Symbol) -> Types::t?`
4. Need a validation to reject circular definition

PR #1942, opened 2024-07-31
Updated all global_variables, and added tests

sampersand:
This PR updates `global_variables`, but more importantly adds tests.

One thing to note is that a lot of these global variables are actually aliases for one another. There may eventually be support for `alias $x $y` (cf https://github.com/ruby/rbs/issues/1647 and https://github.com/ruby/rbs/pull/1936), but this PR was made without them. However, all the tests are written as if they were aliased, so when we eventually convert over, we won't need to update the tests.

I've also added a `RUBY_EXECUTABLE` constant that's constantly visible. Since we don't know which platform we're running on, the only command we're guaranteed to work (and not possibly be something malicious) is the ruby executable. 

Most of the tests for global variables are pretty simple, but some require some more finesse. I've added comments for those.

### DIFF
--- a/core/global_variables.rbs
+++ b/core/global_variables.rbs
@@ -118,7 +118,7 @@ $;: Regexp | String | nil
 $<: RBS::Unnamed::ARGFClass
 
 # This variable is no longer effective. Deprecated.
-$=: bool
+$=: false
 
 # The default output stream for Kernel#print and Kernel#printf. $stdout by default.
 $>: IO
@@ -150,7 +150,7 @@ $LOADED_FEATURES: Array[String]
 $LOAD_PATH: Array[String] & _LoadPathAPI
 
 interface _LoadPathAPI
-  def resolve_feature_path: (String) -> [:rb | :so, String]?
+  def resolve_feature_path: (path feature_path) -> [:rb | :so, String]?
 end
 
 # Contains the name of the script being executed. May be assignable.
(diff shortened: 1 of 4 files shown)


core/global_variables.rbs
```diff
@@ -118,7 +118,7 @@ $;: Regexp | String | nil
 $<: RBS::Unnamed::ARGFClass
 
 # This variable is no longer effective. Deprecated.
-$=: bool
+$=: false
```
sampersand, line 121: Technically `$=` always is `false` now.

core/global_variables.rbs
```diff
@@ -150,7 +150,7 @@ $LOADED_FEATURES: Array[String]
 $LOAD_PATH: Array[String] & _LoadPathAPI
 
 interface _LoadPathAPI
-  def resolve_feature_path: (String) -> [:rb | :so, String]?
+  def resolve_feature_path: (path feature_path) -> [:rb | :so, String]?
```
sampersand, line 153: This accepts a `path`, not just a `String`

test/stdlib/IO_test.rb
```diff
@@ -49,12 +49,8 @@ def test_open
     end
   end
 
-  def ruby
-    ENV["RUBY"] || RbConfig.ruby
-  end
-
   def test_popen
-    with_string("#{ruby} -v") do |command|
+    with_string("#{RUBY_EXECUTABLE} -v") do |command|
```
sampersand, line 53: Now uses `RUBY_EXECUTABLE`

soutaro (approved):
🙏

PR #1943, opened 2024-08-01
Updated TracePoint

sampersand:
This PR updates `TracePoint`. There's a few idiosyncrasies (especially with some tests) that're discussed in the review comments

### DIFF
--- a/core/trace_point.rbs
+++ b/core/trace_point.rbs
@@ -63,7 +63,7 @@
 # `:script_compiled`
 # :   new Ruby code compiled (with `eval`, `load` or `require`)
 #
-class TracePoint < Object
+class TracePoint
   # <!--
   #   rdoc-file=trace_point.rb
   #   - TracePoint.new(*events) { |obj| block }         -> obj
@@ -110,7 +110,7 @@ class TracePoint < Object
   #
   # Access from other threads is also forbidden.
   #
-  def initialize: (*Symbol events) { (TracePoint tp) -> void } -> void
+  def self.new: (*_ToSym events) { (instance tp) -> void } -> instance
 
   # <!--
   #   rdoc-file=trace_point.rb
@@ -172,7 +172,7 @@ class TracePoint < Object
   # (note that we needed to filter out calls by itself from :line handler,
   # otherwise it will call itself infinitely).
   #
-  def self.allow_reentry: () { () -> void } -> void
+  def self.allow_reentry: [T] () { (nil) -> T } -> T
 
   # <!--
   #   rdoc-file=trace_point.rb
@@ -199,7 +199,7 @@ class TracePoint < Object
   #
   #     trace.enabled? #=> true
   #
-  def self.trace: (*Symbol events) { (TracePoint tp) -> void } -> TracePoint
+  def self.trace: (*_ToSym events) { (instance tp) -> void } -> instance
 
   # <!--
   #   rdoc-file=trace_point.rb
@@ -218,7 +218,7 @@ class TracePoint < Object
   # -->
   # Return the called name of the method being called
   #
-  def callee_id: () -> Symbol
+  def callee_id: () -> Symbol?
 
   # <!--
   #   rdoc-file=trace_point.rb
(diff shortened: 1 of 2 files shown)


core/trace_point.rbs
```diff
@@ -110,7 +110,7 @@ class TracePoint < Object
   #
   # Access from other threads is also forbidden.
   #
-  def initialize: (*Symbol events) { (TracePoint tp) -> void } -> void
+  def self.new: (*_ToSym events) { (instance tp) -> void } -> instance
```
sampersand, line 113: One of the few locations in the stdlib that has `_ToSym` and isn't an `intern`.

core/trace_point.rbs
```diff
@@ -172,7 +172,7 @@ class TracePoint < Object
   # (note that we needed to filter out calls by itself from :line handler,
   # otherwise it will call itself infinitely).
   #
-  def self.allow_reentry: () { () -> void } -> void
+  def self.allow_reentry: [T] () { (nil) -> T } -> T
```
sampersand, line 175: While the documentation says the block takes nothing, it actually is given a single `nil` as an argument

core/trace_point.rbs
```diff
@@ -257,7 +257,7 @@ class TracePoint < Object
   #       C.foo
   #     end
   #
-  def defined_class: () -> Module
+  def defined_class: () -> (Class | Module)?
```
sampersand, line 260: Even though `Class | Module` is technically `Module`, i think this is much clearer for the end-user

core/trace_point.rbs
```diff
@@ -352,8 +352,8 @@ class TracePoint < Object
   #     trace.enable { p tp.lineno }
   #     #=> RuntimeError: access from outside
   #
-  def enable:     (?target: (Method | UnboundMethod | Proc)?, ?target_line: Integer?, ?target_thread: Thread?) -> bool
-            | [R] (?target: (Method | UnboundMethod | Proc)?, ?target_line: Integer?, ?target_thread: Thread?) { () -> R } -> R
+  def enable: (?target: Method | RubyVM::InstructionSequence | Proc | nil, ?target_line: int?, ?target_thread: Thread | :default | nil) -> bool
+            | [T] (?target: Method | RubyVM::InstructionSequence | Proc | nil, ?target_line: int?, ?target_thread: Thread | :default | nil) { () -> T } -> T
```
sampersand, line 356: **IMPORTANT**: How should this signature work for non-MRI platforms?

test/stdlib/TracePoint_test.rb
```diff
@@ -1,160 +1,233 @@
...
+    tracepoint :c_call, do: method(:print) do |tp|
+      assert_send_type  '() -> nil',
+                        tp, :binding
+    end
   end
 
+  TracePoint.new(:b_return) { |tp|
+    const_set :CALLEE_ID_OUTSIDE_OF_A_METHOD, tp.callee_id
+    const_set :METHOD_ID_OUTSIDE_OF_A_METHOD, tp.method_id
+  }.enable { }
```
sampersand, line 67: There's no way to get `nil` for the  `callee_id` and `method_id` methods within a method—it needs to be done at top-level.

core/trace_point.rbs
soutaro: Ok. Give the type a try.

core/trace_point.rbs
```diff
@@ -421,7 +421,7 @@ class TracePoint < Object
   # Value from exception raised on the `:raise` event, or rescued on the `:rescue`
   # event.
   #
-  def raised_exception: () -> untyped
+  def raised_exception: () -> Exception
```
soutaro, line 424: How is the case of other events than `:raise` nor `:rescue`?

```suggestion
  def raised_exception: () -> Exception?
```

core/trace_point.rbs
soutaro: We can keep the insn type here, because we have its type definition.
I understand that it may make more sense to extract the implementation specific types for other gems.

core/trace_point.rbs
sampersand: If you call `raised_exception` and there's no exception, it raises an error:
```
$ ruby -e 'TracePoint.trace(:line) { p _1.raised_exception }; p 1'
<internal:trace_point>:402:in `raised_exception': not supported by this event (RuntimeError)
	from -e:1:in `block in <main>'
	from -e:1:in `<main>'
```

soutaro (approved):
💪

PR #1945, opened 2024-08-02
Kernel instance methods

sampersand:
This PR updates all the instance methods on `Kernel`, and also adds tests for them.

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -282,7 +282,7 @@ module Kernel : BasicObject
   #     1.class      #=> Integer
   #     self.class   #=> Object
   #
-  def class: () -> untyped
+  def class: () -> Class
 
   # <!--
   #   rdoc-file=vm_eval.c
@@ -2066,7 +2066,7 @@ module Kernel : BasicObject
   # Returns true if two objects do not match (using the *=~* method), otherwise
   # false.
   #
-  def !~: (untyped) -> bool
+  def !~: (untyped other) -> bool
 
   # <!--
   #   rdoc-file=object.c
@@ -2086,7 +2086,7 @@ module Kernel : BasicObject
   # When you define #<=>, you can include Comparable to gain the methods #<=, #<,
   # #==, #>=, #> and #between?.
   #
-  def <=>: (untyped) -> Integer?
+  def <=>: (untyped other) -> 0?
 
   # <!--
   #   rdoc-file=object.c
@@ -2096,7 +2096,7 @@ module Kernel : BasicObject
   # typically overridden by descendants to provide meaningful semantics in `case`
   # statements.
   #
-  def ===: (untyped) -> bool
+  alias === ==
 
   # <!--
   #   rdoc-file=kernel.rb
@@ -2152,8 +2152,8 @@ module Kernel : BasicObject
   #     chris.define_singleton_method(:greet) {|greeting| "#{greeting}, I'm Chris!" }
   #     chris.greet("Hi") #=> "Hi, I'm Chris!"
   #
-  def define_singleton_method: (interned, Method | UnboundMethod | Proc method) -> Symbol
-                             | (interned) { (?) -> untyped } -> Symbol
+  def define_singleton_method: (interned name, Method | UnboundMethod | Proc method) -> Symbol
+                             | (interned name) { (?) -> untyped } -> Symbol
 
   # <!--
   #   rdoc-file=io.c
@@ -2170,7 +2170,7 @@ module Kernel : BasicObject
   #
   #     1cat[4, 5, 6]
   #
-  def display: (?_Writer port) -> void
+  def display: (?_Writer port) -> nil
 
   # <!--
   #   rdoc-file=object.c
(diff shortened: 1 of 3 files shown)


core/kernel.rbs
```diff
@@ -2267,10 +2267,8 @@ module Kernel : BasicObject
   #     enum.first(4) # => [1, 1, 1, 2]
   #     enum.size # => 42
   #
-  def enum_for: (Symbol method, *untyped, **untyped) ?{ (?) -> Integer } -> Enumerator[untyped, untyped]
-              | () ?{ () -> Integer } -> Enumerator[untyped, self]
```
sampersand, line 2271: this `, self` is actually incorrect; if `method` isn't given it defaults to `each`, and the return value is whatever `each` returns (which is traditionally `self` but not always)

core/kernel.rbs
```diff
@@ -2647,7 +2644,7 @@ module Kernel : BasicObject
   #     Object.new.nil?   #=> false
   #     nil.nil?          #=> true
   #
-  def nil?: () -> bool
+  def nil?: () -> false
```
sampersand, line 2647: Should this be `false` or `bool`? `NilClass` overrides it to `true`

core/kernel.rbs
```diff
@@ -2782,8 +2779,7 @@ module Kernel : BasicObject
   #
   # See #respond_to?, and the example of BasicObject.
   #
-  %a{annotate:rdoc:copy:Object#respond_to_missing?}
-  private def respond_to_missing?: (Symbol, bool) -> bool
+  private def respond_to_missing?: (Symbol | String name, bool include_all) -> bool
```
sampersand, line 2782: The default implementation doesn't actually check the arguments and always returns false, but the documentation says it can also take a `String`, so I added that.

core/kernel.rbs
```diff
@@ -2919,17 +2915,19 @@ module Kernel : BasicObject
...
+  def initialize_copy: (instance object) -> self
 
-  private def initialize_dup: (self object) -> self
+  def initialize_clone: (instance object, ?freeze: bool?) -> self
+
+  def initialize_dup: (instance object) -> self
 end
 
+%a{steep:deprecated}
 Kernel::RUBYGEMS_ACTIVATION_MONITOR: untyped
```
sampersand, line 2933: Do we want to keep this around?

core/kernel.rbs
ParadoxV5: It should be `bool` to maintain polymorphism. See https://github.com/ruby/rbs/issues/1822#issuecomment-2120990437

core/kernel.rbs
ParadoxV5: This is meant to be an abstract method but with a default implementation given.

core/kernel.rbs
```diff
@@ -2447,7 +2445,7 @@ module Kernel : BasicObject
   #     b.instance_of? B   #=> true
   #     b.instance_of? C   #=> false
   #
-  def instance_of?: (Module) -> bool
+  def instance_of?: (Module | Class module_or_class) -> bool
```
soutaro, line 2448: Technically `Class < Module`, and the `| Class` is redundant.

core/kernel.rbs
soutaro: Haha, I'm good to delete it.

core/kernel.rbs
```diff
@@ -2919,17 +2915,52 @@ module Kernel : BasicObject
   #
   #     "my string".yield_self {|s| s.upcase }   #=> "MY STRING"
   #
-  def yield_self: [X] () { (self) -> X } -> X
-                | () -> Enumerator[self, untyped]
+  def yield_self: () -> Enumerator[self, untyped]
+                | [T] () { (self) -> T } -> T
 
-  %a{annotate:rdoc:skip}
+  # <!--
```
soutaro, line 2921: Wow, we have two RDoc entries for `yield_self` and `then`... 🤔

core/kernel.rbs
soutaro: See https://github.com/ruby/rbs/pull/1943/files#r1699454965

sampersand: While the `| Class` is technically redundant, i think it works well as a documentation—new ruby users might not know that all classes are modules, and might be confused as to why `is_a?` and friends don't say they accept `Class`es

ParadoxV5: Has there been discussion in Ruby Core on making `Class` not outright inherit from `Module`?

`Class` is not a strict subtype of `Module` because of the many places that can complain “wrong argument type Class (expected Module) (TypeError)”.
```ruby
include Class
```

core/kernel.rbs
soutaro: It's fine to have `false` here.
I don't think maintaining polymorphism is a good reason here, because Ruby allows (and many classes are already) breaking *Liskov substitution principle* by `undef` and incompatible override.

soutaro (approved):
👍

core/kernel.rbs
ParadoxV5: By polymorphism, my concern is that for `x.nil?` to typecheck correctly, it must be explicitly `Object?` to trigger typecheckers’ special handing. If we type `x` as `Object`, `x.nil?` would typecheck to only `false` even when `x` becomes `nil`.

soutaro: Oh... my bad. Will open another PR to fix the problem.

PR #1948, opened 2024-08-02
Allow alias to module-self-type from module

soutaro:
Defining an alias from a module to its module-self-type causes an error like

```
RBS::UnknownMethodAliasError: .../foo.rbs:2:2...2:18: Unknown method alias name: __id__ => foo (::Foo)
```

This PR fixes the problem.

### DIFF
--- a/lib/rbs/definition_builder.rb
+++ b/lib/rbs/definition_builder.rb
@@ -37,7 +37,7 @@ module RBS
       interface_methods = interface_methods(included_interfaces)
       methods = method_builder.build_interface(type_name)
 
-      import_methods(definition, type_name, methods, interface_methods, subst)
+      import_methods(definition, type_name, methods, interface_methods, subst, nil)
     end
 
     def build_interface(type_name)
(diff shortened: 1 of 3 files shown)


test/rbs/definition_builder_test.rb
```diff
@@ -2686,4 +2686,23 @@ module Bar : Foo
       end
     end
   end
+
+  def test_alias__to_module_self_indierect_method
+    SignatureManager.new(system_builtin: false) do |manager|
```
soutaro, line 2691: It's really complicated to reproduce the problem.

Simply `build_instance(TypeName("::Kernel"))` doesn't fail, because the `self_types` are given through `build_instance` method.

Building `::Foo` instance requires some of the methods of its module-self-type, `::Object`. `::Object` instance type requires the instance methods of `::Kernel`, but it's done other than `#build_instance` method. So, the `module-self-type` methods are not injected, and causes the problem.

lib/rbs/definition_builder.rb
```diff
@@ -86,6 +86,19 @@ def define_instance(definition, type_name, subst)
       one_ancestors = ancestor_builder.one_instance_ancestors(type_name)
       methods = method_builder.build_instance(type_name)
 
+      self_type_methods = one_ancestors.each_self_type.with_object({}) do |self_type, hash| #$ Hash[Symbol, Definition::Method]
```
soutaro, line 89: We cannot simply inject the `self_types` methods to the result `methods`/`definition`, because the `self_types` methods may include other inherited methods, like `#initialize` and having them in the module is something unexpected.

PR #1950, opened 2024-08-06
Dedup empty arrays in parsing

pocke:
This patch dedups empty arrays allocated during parsing.

Currently, RBS allocates many arrays during parsing. It consumes a lot of memory.

This patch reduces memory usage by deduping empty arrays.

BTW, I'm also working on deduping empty Hashes. It caused a performance problem. Therefore this PR focuses on Array and I'll make another PR for Hash.

## Profiling

I've checked the memory usage by `benchmark/memory_new_rails_env.rb` script.

before:

```
Total allocated: 158965078 bytes (1615620 objects)
Total retained:  56607852 bytes (566708 objects)

(snip)

allocated memory by class
-----------------------------------
  81452592  Hash
  20455584  Array
  20114192  RBS::Location
  13382510  String
```

after:

```
Total allocated: 153021766 bytes (1466984 objects)
Total retained:  52321620 bytes (459558 objects)

(snip)

allocated memory by class
-----------------------------------
  81456912  Hash
  20114856  RBS::Location
  14508496  Array
  13380142  String
```

It reduced ~8% retained memory.

I've confirmed `steep check` works well with this patch.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -71,7 +71,7 @@ GEM
     psych (4.0.6)
       stringio
     public_suffix (6.0.1)
-    raap (0.8.0)
+    raap (1.0.0)
       rbs (~> 3.0)
       timeout (~> 0.4)
     racc (1.8.1)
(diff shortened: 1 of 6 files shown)


soutaro (approved):
👍

PR #1958, opened 2024-08-10
Support Range#step behavior change in Ruby 3.4

zverok:
This is a PR to support this change in Ruby: https://github.com/ruby/ruby/pull/7444 (the design is approved by Matz, and the code itself is approved by the Ruby core team).

I am not sure how this actually should be addressed, so I will be thankful for any advice.

Basically, the essence of the change is that:
* Before Ruby 3.4, `Ruby#step`’s type signature was `step(Int step?)`
* After the linked PR, the signature would become `step(Addable step)`, with `Addable` being such that `Range#begin + step` will work and produce the next value.

So, my questions are:
1. How to describe that in RBS syntax?
2. How to adjust RBS `core/range.rbs` to specify that the signature was changed between 3.3 and 3.4?
3. How to proceed with those two PRs (into Ruby itself and into rbs)?

Please advise :pray:

### DIFF
--- a/core/range.rbs
+++ b/core/range.rbs
@@ -1017,8 +1017,10 @@ class Range[out Elem] < Object
   #     ('a'..'e').step { p _1 }
   #     # Default step 1; prints: a, b, c, d, e
   #
-  def step: (?Numeric | int n) -> Enumerator[Elem, self]
-          | (?Numeric | int n) { (Elem element) -> void } -> self
+  def step: (?Numeric | int) -> Enumerator[Elem, self]
+          | (?Numeric | int) { (Elem element) -> void } -> self
+          | (untyped) -> Enumerator[Elem, self]
+          | (untyped) { (Elem element) -> void } -> self
 
   # <!--
   #   rdoc-file=range.c
--- a/test/stdlib/Range_test.rb
+++ b/test/stdlib/Range_test.rb
@@ -110,9 +110,14 @@ class RangeTest < StdlibTest
   def test_step
     (1..10).step
     (1..10).step(2)
+
     if_ruby(..."3.4.0", skip: false) do
       ('A'...'Z').step { |s| s.downcase }
       ('A'...'Z').step(2) { |s| s.downcase }
+    end
+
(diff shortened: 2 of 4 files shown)


ParadoxV5:
> 1. How to describe that in RBS syntax?

It would be using an interface that’s along the lines of
```rbs
interface _Addable[T]
  def +: (T other) -> T
end
```
@sampersand has ongoing interest in interfaces like this; it’d be conveient to coördinate this addition with him.

> 2. How to adjust RBS `core/range.rbs` to specify that the signature was changed between 3.3 and 3.4?

RBS files doesn’t have Ruby versioning currently and the latest versions only track the latest Ruby release.
However, release notes include statements on updating Ruby version and links to PRs.

> 3. How to proceed with those two PRs (into Ruby itself and into rbs)?

Soutaro released [RBS 3.4](https://github.com/ruby/rbs/wiki/Release-Note-3.4) nearing Ruby 3.3’s release as the new bundled version.

I say, proceed with the Ruby side first.
RBSs in this repo (or at least those for the Ruby core) also include a copy of Ruby’s docs; so if the Ruby doc mentions using `#+` during Ruby 3.4’s preview, it gives Soutaro time to update the docs here for an accompanying RBS release.

zverok:
> I say, proceed with the Ruby side first.

I see. My concern here is that due to this “incompatibility”, the core Ruby CI is red (the `test-bundled-gems` task), therefore if I’ll merge it into the main branch, then it would be red for everybody till the rbs would be adjusted.

So I am looking for a way to mitigate it.

soutaro:
@zverok Thank you for letting us know the change.

How about using [rbs_skip_tests](https://github.com/ruby/ruby/blob/master/tool/rbs_skip_tests) in Ruby repo to put your change there? You can skip tests of RBS gem in Ruby repo, and then we can update the RBS tests.

zverok:
@soutaro Thank you for your help! :pray: 

I used rbs_skip_tests solution, and now the PR is merged :tada:

soutaro:
I'm taking over this PR.

PR #1963, opened 2024-08-13
Updated most of Kernel module methods

sampersand:


core/kernel.rbs
```diff
@@ -796,7 +796,7 @@ module Kernel : BasicObject
   # If *const* is defined as autoload, the file name to be loaded is replaced with
   # *filename*.  If *const* is defined but not as autoload, does nothing.
   #
-  def self?.autoload: (interned _module, String filename) -> NilClass
+  def self?.autoload: (interned const, path filename) -> nil
```
sampersand, line 799: if we _ever_ decide to use `const` as a keyword we might want to change this to `constant`?

core/kernel.rbs
```diff
@@ -970,7 +969,9 @@ module Kernel : BasicObject
   # For details on `format_string`, see [Format
   # Specifications](rdoc-ref:format_specifications.rdoc).
   #
-  def self?.format: (String format, *untyped args) -> String
+  def self?.sprintf: (string format, hash[Symbol, untyped] keywords) -> String
```
sampersand, line 972: Oddly enough, `sprintf` only allows positional or keyword arguments, but not both.

core/kernel.rbs
```diff
@@ -1087,8 +1089,8 @@ module Kernel : BasicObject
   #       puts enum.next
   #     } #=> :ok
   #
-  def self?.loop: () { () -> void } -> bot
-                | () -> ::Enumerator[nil, bot]
+  def self?.loop: () -> Enumerator[nil, bot]
+                | () { () -> void } -> untyped
```
sampersand, line 1093: `untyped` instead of `bot` because you can `break` out

core/kernel.rbs
```diff
@@ -1564,7 +1572,14 @@ module Kernel : BasicObject
...
+  def self?.select: ([]?, ?[]?, ?[]?, ?nil) -> bot
+                  | [R < _ToIO]                       (Array[R] read, ?[]?,            ?[]?,            ?nil                  ) -> [Array[R], [], []]
+                  | [R < _ToIO]                       (Array[R] read,  []?,             []?,            Time::_Timeout timeout) -> [Array[R], [], []]?
+                  | [R < _ToIO, W < _ToIO]            (Array[R] read, ?Array[W] write, ?[]?,            ?nil                  ) -> [Array[R], Array[W], []]
+                  | [R < _ToIO, W < _ToIO]            (Array[R] read,  Array[W] write,  []?,            Time::_Timeout timeout) -> [Array[R], Array[W], []]?
+                  | [R < _ToIO, W < _ToIO, E < _ToIO] (Array[R] read, ?Array[W] write, ?Array[E] error, ?nil                  ) -> [Array[R], Array[W], Array[E]]
+                  | [R < _ToIO, W < _ToIO, E < _ToIO] (Array[R] read,  Array[W] write,  Array[E] error, Time::_Timeout timeout) -> [Array[R], Array[W], Array[E]]?
+                  | (Array[_ToIO]? read, ?Array[_ToIO]? write, ?Array[_ToIO]? error, ?Time_Timeout? timeout) -> [Array[_ToIO], Array[_ToIO], Array[_ToIO]]?
```
sampersand, line 1582: The `?[]?` is because you can have empty arrays or `nil` as options for those arguments; the timeout can only be`nil`, or unspecified though.

core/kernel.rbs
```diff
@@ -1670,7 +1685,12 @@ module Kernel : BasicObject
   #     ">"  | boolean | True if the modification time of file1
   #          |         | is after that of file2
   #
-  def self?.test: (String | Integer cmd, String | IO file1, ?String | IO file2) -> (TrueClass | FalseClass | Time | nil | Integer)
+  def self?.test: ('b' | 'c' | 'd' | 'e' | 'f' | 'g' | 'G' | 'k' | 'o' | 'O' | 'p' | 'S' | 'u' | 'z' |  98 |  99 | 100 | 101 | 102 | 103 |  71 | 107 | 111 |  79 | 112 |  83 | 117 | 122, IO | path file) -> bool
+                | ('l' | 'r' | 'R' | 'w' | 'W' | 'x' | 'X' | 108 | 114 | 82 | 119 | 87 | 120 | 88, path filename) -> bool
+                | ('s' | 115, IO | path filename) -> Integer?
+                | ('M' | 'A' | 'C' | 77  | 65 | 67, IO | path file) -> Time
+                | ('-' | '=' | '<' | '>' | 45 | 60 | 61 | 62, IO | path file1, IO | path file2) -> bool
+                | (String | int cmd, IO | path file1, ?IO | path file2) -> (bool | Integer? | Time)
```
sampersand, line 1693: this is the "catch all" case, which covers all the other variants.

core/kernel.rbs
```diff
@@ -1728,7 +1748,7 @@ module Kernel : BasicObject
   # :experimental
   # :   Used for experimental features that may change in future releases.
   #
-  def self?.warn: (*_ToS msg, ?uplevel: int?, ?category: Warning::category?) -> nil
+  def self?.warn: (*_ToS msgs, ?uplevel: int?, ?category: Warning::category | _ToSym | nil) -> nil
```
sampersand, line 1751: One of the few places in the stdlib that uses `_ToSym` and not `interend`; the return value should be a `Warning::category`

lib/rbs/test/type_check.rb
```diff
@@ -331,7 +331,7 @@ def value(val, type)
         when Types::Variable
           true
         when Types::Literal
-          type.literal == val
+          defined?(val.==) and type.literal == val
```
sampersand, line 334: This is needed for the `Kernel#test`, as the `int` variants of the "catch all" variants don't define `==`.

lib/rbs/unit_test/type_assertions.rb
```diff
@@ -336,6 +336,27 @@ def break_from_block(value = nil)
       def pass(message = nil)
         assert true, message
       end
+
+      def assert_send_raises(sig, errclass, object, method, ...)
+        # TODO: ensure args match with signature
```
sampersand, line 341: TODO: The `sig` is actually entirely unused currently, and needs to be updated.

PR #1977, opened 2024-08-23
Backport GC fix PR

soutaro:
Backport #1971.


soutaro:
Confirmed manually all of the expected tests are passing, and the fails are known ones.

PR #1982, opened 2024-08-24
Isolate the case where the argument is Integer in the rand method

rhiroe:
Because it always returns Integer for non-zero integer values.

## Background

```rb
1 + rand(10)
```
Running steep check, this causes `UnresolvedOverloading`.
```
Integer + (Integer | Float)
```
Writing an expression like this seems to cause `UnresolvedOverloading`.

Although not a direct solution, this problem can be avoided if the possibility of a Float being returned when the argument is a non-zero integer can be eliminated.

Thank you.

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -1319,7 +1319,8 @@ module Kernel : BasicObject
   # See also Random.rand.
   #
   def self?.rand: (?0) -> Float
-                | (int arg0) -> (Integer | Float)
+                | (Integer arg0) -> Integer
+                | (_ToInt arg0) -> (Integer | Float)
                 | (::Range[Integer] arg0) -> Integer?
                 | (::Range[Float] arg0) -> Float?
 


ParadoxV5:
`0` is still an `Integer`. To support the concept of non-zero, we’d need #1875.

It’s strange, however, that this situation leads to `UnresolvedOverloading`. `Integer` should be able to `+` either `Integer` or `Float`.
```rbs
Integer + (Integer | Float) -> (Integer | Float)
```

rhiroe:
@ParadoxV5 

> `0` is still an `Integer`

You are right.
However, if the argument is `0`, `Float` is returned here.
https://github.com/ruby/rbs/blob/9244cd679c2abde430b465a7235f4daa35f94cbc/core/kernel.rbs#L1321
Thus, `Integer` here should come to a non-zero integer, right?
https://github.com/ruby/rbs/blob/9244cd679c2abde430b465a7235f4daa35f94cbc/core/kernel.rbs#L1322

ParadoxV5:
It *should*, but it *doesn’t*. That’s what #1875 was about.

rhiroe:
It made sense to me.

PR #1983, opened 2024-08-29
Handle ISEQs compiled by Prism

kddnewton:


soutaro:
Hi @kddnewton!

Does it mean getting an AST from method at runtime is not supported in prism? Or we can do equivalent using other API? Right now or in the future?

XrXr:
@soutaro I think error_highlight does something similar to what you need with Prism: https://github.com/ruby/ruby/blob/eef2121cf3e755cdf33538647d40993d296f9a04/lib/error_highlight/base.rb#L58-L65

Like RubyVM::AST.of, it's re-parsing then finding a node.

soutaro:
@XrXr Got it! Thank you very much.

kddnewton:
@soutaro basically exactly what Alan said. This method https://github.com/ruby/ruby/blob/eef2121cf3e755cdf33538647d40993d296f9a04/lib/error_highlight/base.rb#L86-L95 does effectively the same thing.

PR #1989, opened 2024-09-04
Fix #1988: Enumerable#chain should take enumerable objects

### LINKED ISSUE #1988: Enumerator::Chain#chain excepts Enumerator::Chain[T] as its arguments
The `Enumerator::Chain#chain` type expects an instance of `Enumerator::Chain[T]` as its argument.
As a result, Steep detects nested `#chain` calls as a type error.

Input:

```
array = [1, 2, 3]
array.chain([4, 5, 6]).chain([7, 8, 9])
```

Result:

```
$ bundle exec steep check
# Type checking files:

..........................................................................................................................................................................................F.......................................

app/example.rb:2:29: [error] Cannot pass a value of type `::Array[::Integer]` as an argument of type `::Enumerator::Chain[::Integer]`
│   ::Array[::Integer] <: ::Enumerator::Chain[::Integer]
│     ::Object <: ::Enumerator::Chain[::Integer]
│       ::BasicObject <: ::Enumerator::Chain[::Integer]
│
│ Diagnostic ID: Ruby::ArgumentTypeMismatch
│
└ array.chain([4, 5, 6]).chain([7, 8, 9])
                               ~~~~~~~~~

Detected 1 problem from 1 file
```

Environment:

* Steep 1.7.1
* rbs 3.5.3

tk0miya:
The `Enumerator::Chain#chain` type expects an instance of `Enumerator::Chain[Elem]` as its argument.  As a result, the nested `#chain` calls requires `Enumerator::Chain` object.

It should accept enumerable objects instead of the same kind of object (self).

refs: #1988

### DIFF
--- a/core/enumerable.rbs
+++ b/core/enumerable.rbs
@@ -1846,7 +1846,7 @@ module Enumerable[unchecked out Elem] : _Each[Elem]
   #     e = (1..3).chain([4, 5])
   #     e.to_a #=> [1, 2, 3, 4, 5]
   #
-  def chain: (*Enumerable[Elem] enumerables) -> ::Enumerator::Chain[Elem]
+  def chain: [Elem2] (*_Each[Elem2] enumerables) -> ::Enumerator::Chain[Elem | Elem2]
 
   # <!--
   #   rdoc-file=enum.c
--- a/core/enumerable.rbs
+++ b/core/enumerable.rbs
@@ -1846,7 +1846,7 @@ module Enumerable[unchecked out Elem] : _Each[Elem]
   #     e = (1..3).chain([4, 5])
   #     e.to_a #=> [1, 2, 3, 4, 5]
   #
-  def chain: (*self enumerables) -> ::Enumerator::Chain[Elem]
+  def chain: (*Enumerable[Elem] enumerables) -> ::Enumerator::Chain[Elem]
 
   # <!--
   #   rdoc-file=enum.c


tk0miya:
I did not update the test code because I'm not sure what it's doing.
https://github.com/ruby/rbs/blob/master/test/stdlib/Enumerable_test.rb#L57-L60

core/enumerable.rbs
```diff
@@ -1846,7 +1846,7 @@ module Enumerable[unchecked out Elem] : _Each[Elem]
   #     e = (1..3).chain([4, 5])
   #     e.to_a #=> [1, 2, 3, 4, 5]
   #
-  def chain: (*self enumerables) -> ::Enumerator::Chain[Elem]
+  def chain: (*Enumerable[Elem] enumerables) -> ::Enumerator::Chain[Elem]
```
ParadoxV5, line 1849: Looks like it takes anything that defines `#each`.
```suggestion
  def chain: (*_Each[Elem] enumerables) -> ::Enumerator::Chain[Elem]
```

maybe even?
```suggestion
  def chain: [Elem2] (*_Each[Elem2] enumerables) -> ::Enumerator::Chain[Elem | Elem2]
```

tk0miya: Good point! I adopted the 2nd suggestion :-)

soutaro (approved):
👍

PR #1994, opened 2024-09-05
Improve generics

soutaro:
This PR implements two major improvements on generics.

1. Generics upper bound now can be any type
2. Generic type variable on module/class/interface/type alias can have default type

```rbs
class Foo[T < Object? = String?]
end
```


ParadoxV5:
https://github.com/ruby/rbs/blob/master/docs/syntax.md#generics needs to be updated

PR #2000, opened 2024-09-11
Use IO::NULL instead of /dev/null

hsbt:
`/dev/null` is not provided at Windows platform. We should use `IO::NULL` for that.

### DIFF
--- a/test/rbs/collection/config_test.rb
+++ b/test/rbs/collection/config_test.rb
@@ -83,7 +83,7 @@ class RBS::Collection::ConfigTest < Test::Unit::TestCase
 
   def test_generate_lock_from_relative_git_repository
     mktmpdir do |git_tmpdir|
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍

test/rbs/collection/config_test.rb
```diff
@@ -83,7 +83,7 @@ def test_generate_lock_from_collection_repository
 
   def test_generate_lock_from_relative_git_repository
     mktmpdir do |git_tmpdir|
-      system('git', 'clone', 'https://github.com/ruby/gem_rbs_collection.git', git_tmpdir.to_s, exception: true, 2 => '/dev/null')
+      system('git', 'clone', 'https://github.com/ruby/gem_rbs_collection.git', git_tmpdir.to_s, exception: true, 2 => IO::NULL)
```
nobu, line 86: `err: IO::NULL` would be preferable.

PR #2001, opened 2024-09-11
Fix the portability of an external command

nobu:
`ls` is a very common command on UNIX-like platforms, but not on DOSish platforms.


soutaro (approved):
🙏

PR #2002, opened 2024-09-11
stdlib: Update signature of CSV.read

tk0miya:
* It takes an IO object as `path` argument
* It returns CSV::Table object when headers option given

refs:

* https://docs.ruby-lang.org/en/3.3/CSV.html#method-c-read
* https://docs.ruby-lang.org/en/3.3/CSV.html#class-CSV-label-Option+headers

### DIFF
--- a/stdlib/csv/0/csv.rbs
+++ b/stdlib/csv/0/csv.rbs
@@ -2031,7 +2031,8 @@ class CSV < Object
   #     File.write(path, string)
   #     CSV.read(path, headers: true) # => #<CSV::Table mode:col_or_row row_count:4>
   #
-  def self.read: (String path, ?::Hash[Symbol, untyped] options) -> ::Array[::Array[String?]]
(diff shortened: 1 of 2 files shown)


stdlib/csv/0/csv.rbs
```diff
@@ -2031,7 +2031,8 @@ class CSV < Object
   #     File.write(path, string)
   #     CSV.read(path, headers: true) # => #<CSV::Table mode:col_or_row row_count:4>
   #
-  def self.read: (String path, ?::Hash[Symbol, untyped] options) -> ::Array[::Array[String?]]
+  def self.read: (String | IO path, headers: true | :first_row | Array[untyped] | String, **untyped options) -> ::CSV::Table[CSV::Row]
```
tk0miya, line 2034: I'm not sure what is the best to type args for CSV::Table. Please let me know if there is a more appropriated one.

soutaro (approved):
Thanks!

tk0miya:
Thanks!

PR #2003, opened 2024-09-11
stdlib: Add types for CSV::Table#each

tk0miya:


stdlib/csv/0/csv.rbs
```diff
@@ -3002,7 +3002,7 @@ end
 #     table['Name'] # => ["Foo", "Bar", "Baz"]
 #
 class CSV::Table[out Elem] < Object
-  include Enumerable[untyped]
+  include Enumerable[Elem]
```
tk0miya, line 3005: Is this change correct? If my understanding is correct, `Elem` means a row of the table. So `#each` and Enumerable should use `Elem`, right?

soutaro (approved):
OK. Give it a try! 👍

PR #2008, opened 2024-09-15
Allow RBS/Style/InitializeReturnType

ksss:
Retry https://github.com/ruby/rbs/pull/1763 .

`#initialize` is often overridden, but specifying a return type can be inconvenient for type checking.

The `RBS/Style/InitializeReturnType` cop suggests specifying `void` as the return type when the return type of the `#initialize` method is not `untyped` or `void`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,6 +32,8 @@ RBS/Style:
   Exclude:
     - 'sig/**/*'
     - 'test/**/*'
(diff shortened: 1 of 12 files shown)


soutaro (approved):
OK. Let's make them `void`. 

(cc/ @sampersand)

sampersand:
Yeah I'm ok with this; makes sense when inheriting that overriding it would be annoying

PR #2011, opened 2024-09-16
writer: Preserve empty lines between decls for modules

tk0miya:
https://github.com/ruby/rbs/pull/146 adds preservation empty lines between declarations feature to the classes.

This also introduces the same feature to the modules.  Before this change, our writer inserts a line to between all declarations.  It's a bit difficult to read.  Therefore, this increases the readability of the generated RBS files.

### DIFF
--- a/lib/rbs/writer.rb
+++ b/lib/rbs/writer.rb
@@ -143,10 +143,10 @@ module RBS
         puts "module #{name_and_params(decl.name, decl.type_params)}#{self_type}"
 
         indent do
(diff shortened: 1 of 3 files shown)


soutaro:
I don't think deleting these line breaks makes sense. 🙏

tk0miya:
For example, the types of ActiveRecord models will become very huge. 
https://github.com/kaigionrails/conference-app/blob/main/sig/rbs_rails/app/models/talk.rbs

There are too many blank lines. I know this is not for human, but...

PR #2012, opened 2024-09-17
stdlib: Add types for CSV#headers

euglena1215:
ref. https://docs.ruby-lang.org/ja/latest/method/CSV/i/headers.html


soutaro (approved):
👍

PR #2014, opened 2024-09-17
stdlib: Add types for Open3

euglena1215:
ref. https://docs.ruby-lang.org/ja/latest/class/Open3.html#M_CAPTURE2E

### DIFF
--- a/test/stdlib/Open3_test.rb
+++ b/test/stdlib/Open3_test.rb
@@ -12,6 +12,6 @@ class Open3SingletonTest < Test::Unit::TestCase
     assert_send_type "(*::String, binmode: boolish) -> [ ::String, ::Process::Status ]",
                      Open3, :capture2e, 'echo "Foo"', binmode: true
     assert_send_type "(*::String, stdin_data: ::String) -> [ ::String, ::Process::Status ]",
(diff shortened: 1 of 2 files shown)


test/stdlib/Open3_test.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  library "open3"
+  testing "singleton(::Open3)"
+
+  def test_capture2e
+    assert_send_type "(*::String) -> [ ::String, ::Process::Status ]",
+                     Open3, :capture2e, 'echo "Foo"'
+    assert_send_type "(*::String, binmode: boolish) -> [ ::String, ::Process::Status ]",
+                     Open3, :capture2e, 'echo "Foo"', binmode: true
+    assert_send_type "(*::String, stdin_data: ::String) -> [ ::String, ::Process::Status ]",
```
soutaro, line 14: `cat` may be unavailable, on windows environments.

euglena1215: I have addressed the feedback and replaced the `cat` command with a `ruby` command in the test case to ensure compatibility across platforms, including Windows. This should make the example more portable. Please review the changes and let me know if further adjustments are needed.

PR #2027, opened 2024-09-24
Add attributes and methods to RubyVM::InstructionSequence

p-datadog:
At Datadog we use RubyVM::InstructionSequence objects (via line trace points) and we check types with steep. The type checking fails because none of the InstructionSequence methods are declared in the type file.

This PR adds the definitions with signatures taken from https://docs.ruby-lang.org/en/master/RubyVM/InstructionSequence.html.


soutaro (approved):
Thanks! 👍

PR #2029, opened 2024-09-27
Should set self_type to self

ksss:
Fixed self_type in block for `class_eval` and `module_eval`.
See also https://github.com/soutaro/steep/issues/1232


soutaro (approved):
🙏

PR #2031, opened 2024-09-29
stdlib: Add types for JSON.load_file and .load_file!

tk0miya:
refs:

* https://rubydoc.info/gems/json/JSON.load_file
* https://github.com/ruby/ruby/blob/master/ext/json/lib/json/common.rb#L242-L262
* https://github.com/ruby/ruby/pull/3581
* https://github.com/flori/json/blob/v2.7.2/lib/json/common.rb#L242-L262
* https://github.com/flori/json/pull/387


soutaro (approved):
👏

PR #2036, opened 2024-09-30
stdlib: Make the block arguments of Kernel#proc untyped

tk0miya:
This changes the block argument of `Kernel#proc` to UntypedFunction. Then the arguments of the block become untyped instead of nil.

It allows users to override the type via comment annotation.

For example:

```
proc do |a, b, c|
  # @type var a: String
  # @type var b: Integer
  # @type var c: Symbol
end
```

I found this problem on the Steep repo.
https://github.com/soutaro/steep/blob/v1.8.0/lib/steep.rb#L169-L176

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -1209,7 +1209,7 @@ module Kernel : BasicObject
   # -->
   # Equivalent to Proc.new.
   #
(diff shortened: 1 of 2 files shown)


soutaro (approved):
Thanks!

Issue #2024, opened 2024-09-19
Make C-extension Ractor-safe


mohamedhafez:
In planning ahead to the near future when TruffleRuby can run C-extensions marked with `rb_ext_ractor_safe(true)` in parallel, and for when Ractors are no longer just experimental, it would be great if the C-extension could be made Ractor-safe, or marked as such if it already is so! Would be happy to send in a PR in the later case.

eregon:
From a quick look at https://github.com/ruby/rbs/blob/c2554584cd70387ea54e8f330c63491c438d15f4/ext/rbs_extension/main.c#L4-L9 and the 3 callees it looks fine.

https://github.com/ruby/rbs/blob/master/ext/rbs_extension/parser.c#L2985-L2987 is fine because it's an immutable/deeply-frozen object.
https://github.com/ruby/rbs/blob/master/ext/rbs_extension/constants.c are just some written-once variables (so basically constants).
And https://github.com/ruby/rbs/blob/master/ext/rbs_extension/location.c seems to have no state as well.

@mohamedhafez Could you make a PR?

PR #2041, opened 2024-10-02
Mark as ractor-safe

mohamedhafez:
Closes https://github.com/ruby/rbs/issues/2024


soutaro (approved):
💯

soutaro:
Thank you @eregon for reviewing the source code to identify the extension is Ractor safe. (I think it's safe too.)

PR #2043, opened 2024-10-02
Randomize stdlib test order to uncover hidden assumptions

soutaro:
#2042 didn't reproduce in our CI, because `Net_HTTP_test.rb` sets `$stdout` between `Kernel_test.rb` and `Open3_test.rb`. CI in `ruby/ruby` skips the `Net::HTTP` tests to avoid unpredictable result caused by external network request, which uncovered the `$stdout` reset problem.

I hope randomizing the testing order will uncover this kind of problems.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -34,6 +34,8 @@ jobs:
             job: rubocop validate test_doc build test_generate_stdlib raap
           - ruby: "3.3"
             job: typecheck_test
+    env:
+      RANDOMIZE_STDLIB_TEST_ORDER: "true"
     steps:
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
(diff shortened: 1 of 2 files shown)


soutaro:
Is it reporting a true problem??? 👀 

```
Failure: test_accept_charset(CGISingletonTest):
  Call trace does not match with given method type: #<struct RBS::Test::CallTrace method_name=:accept_charset, method_call=<RBS::Test::ArgumentsReturn:@arguments: [], @exit_value: #<Encoding:UTF-8>, @exit_type: :return>, block_calls=[], block_given=false>.
  <["[Class#accept_charset] ReturnTypeError: expected `::String` but returns `#<Encoding:UTF-8>`"]> was expected to be empty.
/home/runner/work/rbs/rbs/lib/rbs/unit_test/type_assertions.rb:183:in 'block in RBS::UnitTest::TypeAssertions#assert_send_type'
/home/runner/work/rbs/rbs/lib/rbs/unit_test/type_assertions.rb:168:in 'RBS::UnitTest::TypeAssertions#send_setup'
/home/runner/work/rbs/rbs/lib/rbs/unit_test/type_assertions.rb:172:in 'RBS::UnitTest::TypeAssertions#assert_send_type'
test/stdlib/CGI_test.rb:25:in 'CGISingletonTest#test_accept_charset'
```

soutaro:
The problem will be fixed by #2045.

PR #2044, opened 2024-10-03
Omit IOConsoleTest with Windows platform

hsbt:
`IOConsoleTest` couldn't work with Windows platform. We should omit them.

```
<internal:C:/Users/hsbt/DevDrive/github.com/ruby/ruby/lib/rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require': cannot load such file -- pty (LoadError)
        from <internal:C:/Users/hsbt/DevDrive/github.com/ruby/ruby/lib/rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require'
        from test/stdlib/IO_console_test.rb:4:in '<top (required)>'
```

### DIFF
--- a/test/stdlib/IO_console_test.rb
+++ b/test/stdlib/IO_console_test.rb
@@ -1,7 +1,6 @@
 require_relative "test_helper"
 require "io/console"
 require "io/console/size"
-require 'pty'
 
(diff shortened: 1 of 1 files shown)


soutaro (approved):
👍

PR #2054, opened 2024-10-07
Fix UntypedFunction does not provide #map_type_name

tk0miya:
refs: https://github.com/soutaro/steep/issues/1233

### DIFF
--- a/lib/rbs/types.rb
+++ b/lib/rbs/types.rb
@@ -1221,6 +1221,12 @@ module RBS
         end
       end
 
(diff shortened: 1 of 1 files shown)


tk0miya:
I found the type definition of `#map_type_name`. But there is no implementation...
https://github.com/ruby/rbs/blob/v3.6.1/sig/types.rbs#L476

soutaro (approved):
🙇‍♂️

PR #2055, opened 2024-10-09
Skip PTY_test with Windows platform

hsbt:
This change is similar with https://github.com/ruby/rbs/pull/2044. Windows platform didn't have `pty` module. We should skip this test in Windows platform.


soutaro (approved):
🙏

PR #2057, opened 2024-10-10
Remove unused variables from Visitor test

Morriar:
Introduced in #1975 but we don't need these variables, only `declarations`.


soutaro (approved):
🙏

PR #2078, opened 2024-11-06
stdlib: Add types for base64digest of openssl

dak2:
refs:
 - https://github.com/ruby/openssl/blob/bd647c3b05513fe8782d8246b9b7a03381b67eaf/lib/openssl/hmac.rb#L13-L19
 - https://docs.ruby-lang.org/en/master/OpenSSL/HMAC.html#method-i-base64digest
 - https://github.com/ruby/openssl/blob/bd647c3b05513fe8782d8246b9b7a03381b67eaf/lib/openssl/hmac.rb#L60-L75
 - https://docs.ruby-lang.org/en/master/OpenSSL/HMAC.html#method-c-base64digest

### DIFF
--- a/stdlib/openssl/0/openssl.rbs
+++ b/stdlib/openssl/0/openssl.rbs
@@ -3665,6 +3665,23 @@ module OpenSSL
     #
     def self.hexdigest: (String | Digest algo, String key, String data) -> String
 
+    # <!--
+    #   rdoc-file=ext/openssl/lib/openssl/hmac.rb
+    #   - HMAC.base64digest(digest, key, data) -> aString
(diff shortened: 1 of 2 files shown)


tk0miya:
LGTM!
I'm not sure why CI has broken. It seems the error is not related to this PR.

It has been already broken.
https://github.com/ruby/rbs/actions/workflows/ruby.yml

dak2:
@tk0miya 

Thanks!

All CIs PASSED since https://github.com/ruby/rbs/pull/2084 resolved that issue.

soutaro (approved):
👏

Issue #2076, opened 2024-11-02
rbs prototype rb fails in cases like self::CONSTANTS.


fumihumi:
rbs prototype rb fails in cases like self::CONSTANTS.

hoge.rb like...

```ruby
class Hoge
  self::SAMPLE = "SAMPLE".freeze
end
```

$ ./exe/rbs prototype rb hoge.rb

```
/Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/rb.rb:447:in `const_to_name!': unhandled exception
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/rb.rb:438:in `const_to_name!'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/rb.rb:371:in `process'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/rb.rb:428:in `block in process_children'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/helpers.rb:91:in `block in each_node'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/helpers.rb:89:in `each'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/helpers.rb:89:in `each_node'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/helpers.rb:85:in `each_child'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/rb.rb:427:in `process_children'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/rb.rb:422:in `process'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/rb.rb:428:in `block in process_children'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/helpers.rb:91:in `block in each_node'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/helpers.rb:89:in `each'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/helpers.rb:89:in `each_node'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/helpers.rb:85:in `each_child'
	from /Users/me/Repositories/github.com/ruby/rbs/lib/rbs/prototype/rb.rb:427:in `process_children'
[...]

PR #2079, opened 2024-11-06
prototype rb: Fix crashed by self::CONST

tk0miya:
Closes: #2076


soutaro (approved):
Hmm. Let's give it a try!

PR #2083, opened 2024-11-14
Fix return type of `Process.detach` to return `Process::Waiter`

pocke:
`Process.detach` returns a `Process::Waiter`, a subclass of `Thread`. But the RBS defines that it returns a Thread.

This PR fixes the return type to `Process::Waiter`.

`Process.detach` returns `Process::Waiter` since Ruby 2.2.

```console
$ docker run -it --rm rubylang/all-ruby ./all-ruby -e 'p Process.detach(1).class'

(snip)

ruby-1.8.0            Thread
...
ruby-2.1.10           Thread
ruby-2.2.0-preview1   Process::Waiter
...
ruby-3.4.0-preview2   Process::Waiter
```


soutaro (approved):
👍

PR #2090, opened 2024-11-18
docs: Add singleton attribute members to syntax.md

tk0miya:
refs: https://github.com/soutaro/rbs-inline/pull/123#issuecomment-2482239594


soutaro (approved):
👍

PR #2091, opened 2024-11-19
fix for inherited not supporting subclassing

HoneyryderChuck:
the arg of inherited is a subclass of the current class, not (just) a ::Class.

There's also a similar limitation in `Kernel#class` definition, where `def class: () -> Class` should map out to something more useful, i.e. `def class: () -> instance_class`, but that would require support for a new keyword, and there may be other limitations?

### DIFF
--- a/core/class.rbs
+++ b/core/class.rbs
@@ -149,7 +149,7 @@ class Class < Module
   #     New subclass: Bar
   #     New subclass: Baz
   #
-  def inherited: (Class arg0) -> untyped
+  def inherited: (instance arg0) -> untyped
 
   # <!--
   #   rdoc-file=object.c


ParadoxV5:
> There's also a similar limitation in `Kernel#class` definition, where `def class: () -> Class` should map out to something more useful, i.e. `def class: () -> instance_class`, but that would require support for a new keyword, and there may be other limitations?

* Perhaps #1542?

soutaro:
Unfortunately, it doesn't work with the current `instance` type semantics.

This is an output to test `def Object.inherited(klass) = klass.fooo`.

```
test.rb:1:36: [error] Type `::Object` does not have method `fooo`
│
└ def Object.inherited(klass) = klass.fooo
                                      ~~~~
```

The `instance` type is resolved to `Object`, not `singleton(Object)`.

ParadoxV5:
> ```
> test.rb:1:36: [error] Type `::Object` does not have method `fooo`
> │ Diagnostic ID: Ruby::NoMethod
> │
> └ def Object.inherited(klass) = klass.fooo
>                                       ~~~~
> ```
> 
> The `instance` type is resolved to `Object`, not `singleton(Object)`.

This hints that `singleton(Object)` only reads from `Object` RBS and doesn’t consider itself an instance of `Class`; this is an oversight.

PR #2092, opened 2024-11-20
Make it usable even for stdlib if rubygems is available

ksss:
ref: https://github.com/ruby/mutex_m/pull/20

mutex_m v0.3.0 has been released. ([link](https://github.com/ruby/mutex_m/releases/tag/v0.3.0))
However, even if this gem is installed, there are limited situations where the signature can be used.

This fix allows gem signatures to be used in the above two situations as long as the gem is installed.

## Before

- Write `mutex_m` in rbs_collection.yaml
    - ❌ Always stdlib is selected.
- Find `mutex_m` in manifest.yaml
    - ❌ Always stdlib is selected.
- Find `mutex_m` in Gemfile.lock
    - ✅ Rubygems is selected.

## After

- Write `mutex_m` in rbs_collection.yaml
    - ✅ Rubygems is selected if gem was installed.
    - ✅ Stdlib is selected if gem was not installed.
- Find `mutex_m` in manifest.yaml
    - ✅ Rubygems is selected if gem was installed.
    - ✅ Stdlib is selected if gem was not installed.
- Find `mutex_m` in Gemfile.lock
    - ✅ Rubygems is selected.

⚠️ Assume that `source` is not specified in rbs_collection.yaml.

### DIFF
--- a/lib/rbs/collection/config/lockfile_generator.rb
+++ b/lib/rbs/collection/config/lockfile_generator.rb
@@ -7,6 +7,7 @@ module RBS
         DEPRECATED_STDLIBS = [
           ['mutex_m', '0.3.0']
         ]
+        ALMUNI_STDLIBS = []
 
         class GemfileLockMismatchError < StandardError
           def initialize(expected:, actual:)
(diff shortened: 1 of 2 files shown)


soutaro:
@ksss It looks like this change allows writing gem dependencies in `manifest.yaml`. (Is it?) I think we are not willing to allow it right now.

ksss:
Sorry, I did not intend to be able to specify third party gems.
Only stdlib or alumni gems can be specified.

ksss:
I found the following current behavior.
If a third party gem is listed in manifest.yaml of rubygems, it can be loaded.

https://github.com/ruby/rbs/blob/fff1f9a5de8100b7c569870ab2f66a03fa830a09/lib/rbs/environment_loader.rb#L73-L75

soutaro:
The feature is implemented in #2115.

PR #2094, opened 2024-11-21
Bundle update rubocop-on-rbs

ksss:
Top-level interfaces and type aliases should generally be avoided, but I think we can allow them as a special exception in this repository.


soutaro (approved):
👍

PR #2099, opened 2024-11-26
stdlib: csv: Add types for Array#to_csv and String#parse_csv

tk0miya:
refs:

* https://www.rubydoc.info/gems/csv/Array#to_csv-instance_method
* https://www.rubydoc.info/gems/csv/String#parse_csv-instance_method

### DIFF
--- a/stdlib/csv/0/csv.rbs
+++ b/stdlib/csv/0/csv.rbs
@@ -3756,3 +3756,21 @@ class CSV::Table[out Elem] < Object
   #
   def values_at: (*untyped indices_or_headers) -> untyped
 end
+
+%a{annotate:rdoc:skip}
+class Array[unchecked out Elem] < Object
+  # Equivalent to CSV::generate_line(self, options)
+  #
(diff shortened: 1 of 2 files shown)


stdlib/csv/0/csv.rbs
```diff
@@ -3814,3 +3814,21 @@ class CSV::Table[out Elem] < Object
   #
   def values_at: (*untyped indices_or_headers) -> untyped
 end
+
+%a{annotate:rdoc:skip}
+class Array[unchecked out Elem] < Object
+  # Equivalent to CSV::generate_line(self, options)
```
soutaro, line 3820: The source code has docs, but opted out. Opened a PR to include them to the doc. https://github.com/ruby/csv/pull/322

soutaro:
Confirmed the `rbs annotate` replaces the docs if it runs with the lates release of csv where the `:nodoc:` is deleted.
So, merging this. The release of ruby with updated csv-gem will fix the documentation issue finally.

PR #2098, opened 2024-11-25
Add templating mechanism to automatically generate `.c`/`.h` files

amomchilov:
This PR automates the creation of these pre-existing files:

1. `constants.h`
1. `constants.c`
1. `ruby_objs.h`
1. `ruby_objs.c`

In this initial PR, we aim to generate these files to match as closely to their original hand-authored versions. To facilitate this, we added some trickery to the ERB files, to get it as close as practical. These will be removed in a follow-up PR.

PR #2102, opened 2024-11-27
Clean up temporary workarounds from templated code

amomchilov:
This PR is a follow-up built on top of #2098. It doesn't change any of its behaviour, but just removes the temporary workarounds that we used to minimize the diff of #2098.

Only [the last 4 commits](https://github.com/ruby/rbs/pull/2102/files/081154f17a2ca069b3b0def06a769ac948361750..d56b699cd93726b13043b3cc609c2e9c6ccd46ae) are new to this PR, the rest just come from #2098.

This PR is best reviewed commit-by-commit.


templates/src/ruby_objs.c.erb
```diff
@@ -10,107 +10,17 @@
...
-VALUE <%= node.c_function_name %>(<%= node.fields.map { |field| "#{field.c_type} #{field.c_name}" }.join(", ") %>) {
-  <%- hash_var = uses_args_hash.include?(node.c_function_name) ? "args" : "kwargs" -%>
-  <%- hash_var = "kw_args" if node.c_function_name == "rbs_ast_decl_module_self" -%>
-  VALUE <%= hash_var %> = rb_hash_new();
+<%- nodes.each do |node| -%>
+VALUE <%= node.c_function_name %>(<%= node.fields.map { |field| "#{field.c_type} #{field.name}" }.join(", ") %>) {
+  VALUE _init_kwargs = rb_hash_new();
   <%- node.fields.each do |field| -%>
-  rb_hash_aset(<%= hash_var %>, ID2SYM(rb_intern("<%= field.name %>")), <%= field.c_name %>);
+  rb_hash_aset(_init_kwargs, ID2SYM(rb_intern("<%= field.name %>")), <%= field.name %>);
```
amomchilov, line 17: We picked `_init_kwargs` instead of a name like `kwargs`, `args`, etc. to make it less likely that it collides with the name of a parameter to this function.

soutaro (approved):
Thanks! Looks great. 👍

PR #2103, opened 2024-11-28
Mark all internal `parse_` methods as `static`

Morriar:
From all the methods only called from inside `parse.c`, some methods where marked as `static` some other not.

Let's standardize all of them as `static` unless exposed in the `parse.h` file.


soutaro (approved):
🙏

ParadoxV5:
IIRC, marking internals as `static` removes them from the shared library’s function table. This is beneficial in binary size reductions if the header implies that they’re not meant to be exposed.

PR #2105, opened 2024-11-29
Compile without activesupport

soutaro:
CI in `ruby/ruby` compiles rbs-gem because it's a bundled gem. But adding `activesupport` to compile environment is not a good idea.

This PR contains a few changes:

1. Stop hardcoding `bundle exec ruby` commandline (this is in fact unrelated to activesupport)
2. Stop triggering `templates` task always
3. Add generated files

With this PR, we can compile the extension without installing activesupport. 🎉


soutaro:
@Morriar @amomchilov  The problem is having activesupport in ruby/ruby CI. Another solution would be stop using activesupport. (Doesn't look too much difficult, but avoiding `templates` task looks easier to me.)

Morriar:
iirc correctly we only use `activesupport` for `underscore` and `demodulize`, we could copy over these methods and remove the dependency if needed.

Rakefile
```diff
@@ -44,14 +56,18 @@ task :confirm_annotation do
...
+  sh "#{ruby} templates/template.rb src/ruby_objs.c"
 end
 
 task :compile => "ext/rbs_extension/lexer.c"
-Rake::Task[:compile].prereqs.prepend :templates
+task :compile => "include/rbs/constants.h"
+task :compile => "include/rbs/ruby_objs.h"
+task :compile => "src/constants.c"
+task :compile => "src/ruby_objs.c"
+# Rake::Task[:compile].prereqs.prepend :templates
```
Morriar, line 70: I guess we don't need this anymore?

PR #2106, opened 2024-11-29
Remove the dependency to activesupport

mame:
... by replacing `demodulize` and `underscore` with hand-written regexps.

### DIFF
--- a/templates/template.rb
+++ b/templates/template.rb
@@ -4,9 +4,6 @@ require "erb"
 require "fileutils"
 require "yaml"
 
-require "active_support"
-require "active_support/core_ext/string/inflections"
-
 module RBS
   class Template
     class Field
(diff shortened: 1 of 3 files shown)


templates/template.rb
```diff
@@ -48,12 +45,12 @@ class Type
 
       def initialize(yaml)
         @ruby_full_name = yaml["name"]
-        @ruby_class_name = @ruby_full_name.demodulize
+        @ruby_class_name = @ruby_full_name[/[^:]+\z/] # demodulize-like
+        name = @ruby_full_name.gsub("::", "_")
+        name = name.gsub(/(^)?(_)?([A-Z](?:[A-Z]*(?=[A-Z_])|[a-z0-9]*))/) { ($1 || $2 || "_") + $3.downcase } # underscore-like
         @c_function_name = if @ruby_full_name =~ /^RBS::Types/
```
ParadoxV5, line 51:
```suggestion
        @c_function_name = if @ruby_full_name.start_with? 'RBS::Types'
```

ParadoxV5 (commented):
Fixes https://github.com/ruby/rbs/pull/2105#issuecomment-2507896809

`underscore`-like is so complex 😮.

amomchilov:
`activesupport` was already an indirect dependency of RBS, and it still is after this PR, because `steep` depends on it. Is there a plan to remove `activesupport` from `steep`?

HoneyryderChuck:
An alternative implementation is to provide the same method via a refinement.

soutaro:
@amomchilov Yeah, it's subtle thing... The problem is for Ruby core developers who don't install full-development setup of RBS. They often develop without bundler or maybe rubygems, so having less dependency to minimum setup for RBS makes more sense for the case.

It was my fault that I didn't notice the problem when I was reviewing your initial PR. Looks like this change works good.

PR #2108, opened 2024-12-02
add correct base class to OpenSSL::Digest

HoneyryderChuck:
some method definitions, like OpenSSL::Digest#file, are missing otherwise


soutaro (approved):
👍

Issue #2109, opened 2024-12-02
`Attribute#ivar_name` API is unclear


amomchilov:
The `#ivar_name` attribute of each `RBS::Members::Attribute*` class can return one of 3 things:

```ruby
# ivar_name will return `nil`, which indicates the default (@a) is inferred
attr_reader a: Integer

# ivar_name will return `:something_else`
attr_reader b(@something_else): String

# ivar_name will return `false`
attr_reader c(): bool
```

This has a few downsides:

1. The meaning of `nil` and `false` is non-obvious
1. Most callers probably care about the inferred name (e.g. `@a`), and will have to compute it themselves when they get `nil`.
1. The RBS type of this field is complex:
    https://github.com/ruby/rbs/blob/3fc6b05588e9378a387fca1caad58a75076dd90c/sig/members.rbs#L177
    
    * This will also makes the C API less clear, since the type has to be some general node type, rather than specifically a symbol type (or `NULL`).
    
I propose we deprecate this field, and come up with an easier-to-consume alternative.

PR #2110, opened 2024-12-02
RFC: new `Attribute#ivar` API

amomchilov:
My proposal for fixing #2109. Feedback welcome!


lib/rbs/ast/members.rb
```diff
@@ -304,6 +304,21 @@ def update(name: self.name, type: self.type, ivar_name: self.ivar_name, kind: se
...
+            ivar_name = case self.ivar_name
+            when false
+              return nil # Skip the instance variable declaration entirely
+            when nil
+              :"@#{name}" # Infer the instance variable name from the attribute name
+            else
+              self.ivar_name # Use the custom instance variable name given by the user
+            end
+
+            InstanceVariable.new(name: ivar_name, type: type, location: location, comment: comment)
```
amomchilov, line 319: Open question: what should the `location` and `comment` be? The same as the `self` attribute, or `nil`?

sig/members.rbs
```diff
@@ -185,6 +185,8 @@ module RBS
         include _HashEqual
 
         def update: (?name: Symbol, ?type: Types::t, ?ivar_name: Symbol | false | nil, ?kind: kind, ?annotations: Array[Annotation], ?location: loc?, ?comment: Comment?, ?visibility: visibility?) -> instance
+
+        def ivar: () -> InstanceVariable?
```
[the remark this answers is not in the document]
Morriar: Does it still make sense for `ivar_name` to accept false in the constructor and the attr_reader?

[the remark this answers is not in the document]
amomchilov: Good question, since there's backwards compatibility concerns there.

We can add a new keyword arg (perhaps `explicit_ivar_name:`?) and use that, while still supporting the old.

lib/rbs/ast/members.rb
soutaro: I don't think we should have the `comment` attribute, because it's given to the attribute syntax, not to the instance variable name.

Not very sure about `location`, but I think going without location should work for now.

lib/rbs/ast/members.rb
```diff
@@ -304,6 +304,21 @@ def update(name: self.name, type: self.type, ivar_name: self.ivar_name, kind: se
             visibility: visibility
           )
         end
+
+        def ivar
```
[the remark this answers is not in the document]
soutaro: No difference between the two looks fine, if we consider this is a kind of *utility* class.
If we really need the detail of the syntax, we can use the old API.

soutaro:
I'm generally good for this. My suggestion is dropping `location` and `comment`.

lib/rbs/ast/members.rb
amomchilov: We still have to fill in the parameters, are you suggesting I pass `location: nil, comment: nil`?

soutaro: Oops, I'm sorry. I was confused.

I thought the `InstanceVariable` is a new class for this API, but actually it is an existing class which models instance variable declaration.

soutaro:
I'm sorry for confusion. 🙇‍♂️ 

I think reusing the AST structure is good for this, while introducing a new class looks a bit too much.

Simply adding a new methods seem the best way:

* `instance_variable_name: Symbol | nil` -- the new method!
* `raw_instance_variable_name: Symbol | false | nil` -- to avoid information loss
* `ivar_name: Symbol | false | nil` -- keep it for compatibility until rbs-4.0, which will be deprecated soon.

Issue #1549, opened 2023-09-27
stringio should be a stdlib, but it's in `core/`


pocke:
In Ruby, `stringio` is a stdlib.
https://github.com/ruby/ruby/tree/50520cc1930331bccdb94730e17ddc01798f2be0/ext/stringio

But currently, RBS has `string_io` as a core library unexpectedly.
https://github.com/ruby/rbs/blob/abcc97d3c8b024856ba9603a7ef22d422652ff42/core/string_io.rbs

So we should do the two things:

* Move string_io.rbs from `core/` to `stdlib/`.
* Rename it to `stringio.rbs`
  * because the library name is `stringio`.

BTW, this will introduce a breaking change. We should do it carefully. For example, we can provide a migration process.

ksss:
I have a proposal.

## Our goal

I believe stringio is already gemmed and the goal is to move to gem.
https://github.com/ruby/stringio

## Problem

One thing to be aware of regarding migration is older versions of rbs. The following are possible problems with older rbs.

- Moving to gem conflicts with core in old rbs
- Adding `stringio` to manifest.yaml of gems in gem_rbs_collection crashes with error when loaded in old rbs.

## Proposal

1. Move to stdlib from core.
    - Read RBS from gem if exists in rubygems
    - Otherwise, **always** load RBS from stdlib.
1. Add stringio to manifest.yaml for cgi, csv, and open-uri in stdlib
1. If `StringIO` is found in RBS without `stringio` in both manifest.yaml and gem dependencies, issue a warning.
    - Optional because I don't know if I can do it.
1. Release.
1. Wait for penetration... and then add stringio to the manifest.yaml of gem_rbs_collection.
    - If executed too soon, errors will occur with old rbs.
1. Move to gem from stdlib.

PR #2114, opened 2024-12-03
Milestone to graduate stringio from core.

ksss:
Ref: https://github.com/ruby/rbs/issues/1549

I have implemented 1 through ~~3~~ 2 as proposed [here](https://github.com/ruby/rbs/issues/1549#issuecomment-2513755741).


ksss:
I have found that we should not issue warnings immediately because there is no method of recovery.
I will stop issuing warnings and scope up to stage 2.

soutaro (approved):
👍

PR #2093, opened 2024-11-21
Update minitest

ksss:
- Drop manifest.yaml
    - mutex_m has been removed from the minitest dependency.
- Update some signature semi-automatically.
    - Using by https://github.com/ksss/orthoses/blob/96a71c1116ebddd792fee3dc5a7abda402e18925/examples/minitest/Rakefile

PR #2115, opened 2024-12-04
mutex_m support

soutaro:
Design doc: https://hackmd.io/@soutaro/HkADVu3QJx

Related to: https://github.com/ruby/rbs/pull/2092

### DIFF
--- a/lib/rbs/collection/config/lockfile_generator.rb
+++ b/lib/rbs/collection/config/lockfile_generator.rb
@@ -4,6 +4,8 @@ module RBS
   module Collection
     class Config
       class LockfileGenerator
+        ALUMNI_STDLIBS = { "mutex_m" => ">= 0.3.0" }
+
         class GemfileLockMismatchError < StandardError
           def initialize(expected:, actual:)
             @expected = expected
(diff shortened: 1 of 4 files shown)


lib/rbs/collection/config/lockfile_generator.rb
```diff
@@ -149,16 +162,37 @@ def generate
...
+            version = ALUMNI_STDLIBS.fetch(name)
+            if from_gem
+              # From `dependencies:` of a `manifest.yaml` of a gem
+              source = find_source(name: name) or raise
+              if source.is_a?(Sources::Stdlib)
+                RBS.logger.warn {
+                  "`#{name}` is included in the RBS dependencies of `#{from_gem}`, but the type definition as a stdlib in rbs-gem is deprecated. Add `#{name}` (#{version}) to the dependency of your Ruby program to use the gem-bundled type definition."
+                }
+              else
+                RBS.logger.info {
```
soutaro, line 184: This is `INFO` message because the end-user is not responsible for the `dependencies:` of the type definition.

soutaro:
`stdlib/minitest` has `mutex_m` as RBS dependency. But we cannot delete it because the gem doesn't declare dependency to `mutex_m` gem.

ksss:
RBS dependency of stdlib/minitest has been removed by https://github.com/ruby/rbs/pull/2093 .

PR #2123, opened 2024-12-05
Deprecate `Kernel#Namespace`

ksss:
The **Namespace** functionality will be added in Ruby4. https://bugs.ruby-lang.org/issues/19744
If added, `Namespace()` is likely to create confusion.

I propose that we deprecate it now.


soutaro (approved):
Thanks! 👍

PR #2128, opened 2024-12-05
Un-hardcode repo name in test suite

amomchilov:
These tests inadvertently require that the devs working on RBS have it checked out in a folder called `rbs`.

This becomes an issue if a dev wants to check out the RBS repo under a different name, such as to check out multiple worktrees at a time.

```sh
git worktree add ../rbs1
git worktree add ../rbs1
```


soutaro (approved):
💯

PR #2131, opened 2024-12-06
Designated initializers

amomchilov:
This PR changes several repeated assignments with [C99 designated initializers](https://en.cppreference.com/w/c/language/struct_initialization).

We're already requiring C99 (well, the GNU dialect of it), so there shouldn't be any compiler versioning concerns:

https://github.com/ruby/rbs/blob/4077aae170a816e4adc6df0549d84d13863ccd71/ext/rbs_extension/extconf.rb#L13

These have 3 benefits:

1. In some places they can replace a `calloc`, since they automatically zero-out any fields that that weren't explicitly set.
2. They have better IDE auto-completions to help remember to set all fields.
3. As a combination of 1 and 2, using these now will help us land our large C API changes on top.

### DIFF
--- a/ext/rbs_extension/location.c
+++ b/ext/rbs_extension/location.c
@@ -36,9 +36,12 @@ void rbs_loc_alloc_children(rbs_loc *loc, unsigned short cap) {
   size_t s = RBS_LOC_CHILDREN_SIZE(cap);
   loc->children = malloc(s);
 
-  loc->children->len = 0;
-  loc->children->required_p = 0;
-  loc->children->cap = cap;
+  *loc->children = (rbs_loc_children) {
+    .len = 0,
+    .required_p = 0,
+    .cap = cap,
+    .entries = { 0 },
+  };
 }
 
 static void check_children_cap(rbs_loc *loc) {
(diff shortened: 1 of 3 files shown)


ext/rbs_extension/location.c
```diff
@@ -54,27 +57,28 @@ static void check_children_cap(rbs_loc *loc) {
 }
 
 void rbs_loc_add_required_child(rbs_loc *loc, ID name, range r) {
-  check_children_cap(loc);
-
-  unsigned short i = loc->children->len++;
-  loc->children->entries[i].name = name;
-  loc->children->entries[i].rg = rbs_new_loc_range(r);
```
amomchilov, line 61: This code is the same as `rbs_loc_add_optional_child`, so we can reuse that, and just set `required_p` after.

ext/rbs_extension/location.c
```diff
@@ -133,8 +139,12 @@ static VALUE location_initialize_copy(VALUE self, VALUE other) {
   rbs_loc *self_loc = rbs_check_location(self);
   rbs_loc *other_loc = rbs_check_location(other);
 
-  self_loc->buffer = other_loc->buffer;
-  self_loc->rg = other_loc->rg;
+  *self_loc = (rbs_loc) {
+    .buffer = other_loc->buffer,
+    .rg = other_loc->rg,
+    .children = NULL,
```
amomchilov, line 145: Not technically necessary, but IMO explicit+clear > implicit+terse

PR #2139, opened 2024-12-11
Fix typo in `instance` usage example

Morriar:
I believe what we want an `Array[instance]`?

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -291,7 +291,7 @@ class Foo
 
   def foo: () -> instance
 
-  @@foos: Array[instances]
+  @@foos: Array[instance]
 
   include Enumerable[class]
 end


docs/syntax.md
```diff
@@ -291,7 +291,7 @@ class Foo
 
   def foo: () -> instance
```
Morriar, line 292: @soutaro I'm also surprised by how it's used in this case? What's the meaning of it? Is it the same as `self`?

ParadoxV5: Dummy names like ‘foo’ don’t always make comprehensible general examples. 👍

soutaro: @Morriar This is in fact same with `self`. They are different on `def self?.foo: `, but yeah, it's exactly same in this case.

soutaro (approved):
🙏

PR #2151, opened 2024-12-15
`IO#each_line` and `Kernel#readlines` accept `chomp: true`

ParadoxV5:
* Help wanted: testing
  * Broken Windows Theory: We don’t need to test those because `IO` and `Kernel` don’t have complete test coverage.

### DIFF
--- a/core/io.rbs
+++ b/core/io.rbs
@@ -3104,8 +3104,8 @@ class IO < Object
   #
   # Returns an Enumerator if no block is given.
   #
-  def each_line: (?String sep, ?Integer limit) { (String line) -> void } -> self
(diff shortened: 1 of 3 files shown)


soutaro:
`exit` won't have any output... https://github.com/ruby/rbs/commit/14ea43f92832469d7f2e8f4241b14121b8cba54b

soutaro:
Inserting something to `ARGV` helps returning the `readlines` calls. https://rubyapi.org/3.3/o/argf

soutaro:
I tried to implement a test for `Kernel#readlines` through `ARGV`, but it blocks on the second call during the test. 😢

PR #2158, opened 2024-12-16
Dir.mktmpdir and File.join accept Pathname

znz:
`File.join` uses `FilePathStringValue`, so it accepts `Pathname`.

https://github.com/ruby/ruby/blob/7c2660b34774c017accff8a7d78371c2f368b9dc/file.c#L5160

`Dir.mktmpdir` uses `File.join` internally, so it also accepts `Pathname` as parent.
And it use `String.try_convert` for `prefix_suffix`, so they accept `string` instead of `String`.

https://github.com/ruby/ruby/blob/923f831804b8bc26bf0412b932791d9090ff8ac4/lib/tmpdir.rb#L159-L165

### DIFF
--- a/core/file.rbs
+++ b/core/file.rbs
@@ -1314,7 +1314,7 @@ class File < IO
   #
   #     File.join("usr", "mail", "gumby")   #=> "usr/mail/gumby"
   #
-  def self.join: (*string) -> String
+  def self.join: (*path) -> String
 
   # <!--
   #   rdoc-file=file.c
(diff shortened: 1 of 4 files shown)


soutaro:
@znz Thanks! Can you add some tests to `File_test.rb`?

We don't have good place for `Dir.mktmpdir` tests. It will be really great if you can rewrite the test cases in `Dir_tmpdir_test.rb`. (It has only two test methods, not too difficult.)

PR #2174, opened 2024-12-17
Document a few details about `class`

Morriar:
Following the discussion in https://github.com/ruby/rbs/pull/2139#discussion_r1880317126.

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -289,7 +289,9 @@ The following `class`/`instance` types are allowed.
 class Foo
   attr_reader parent: class
 
-  def foo: () -> instance
+  def foo: () -> instance # behaves like `self` in this context
+
+  def self?.bar: () -> instance # behaves like `class` for `def self.bar()` and `self` for `def bar()`
 
   @@foos: Array[instance]
 


ParadoxV5:
`class` refers to the class itself. From my understanding,
```rbs
class A
  X: class # A
end
class B < A
  # B::X is of type B
end
```

---

I’ve always considered `instance` and `self` to be different:
```ruby
SUBJECT = Object.new
SUBJECT.extend Comparable # i.e., SUBJECT.singleton_class.include Comparable
```
```rbs
class Object
  def dup: () -> instance # Object
  def itself: () -> self # Object
end
SUBJECT: Object & Comparable
# SUBJECT.dup    -> Object
# SUBJECT.itself -> Object & Comparable
```

soutaro:
Thanks. Looks good.

I found the biggest difference is when the class is generic. When the type is `C[Integer]`, `self` is `C[Integer]`, `instance` is `C[untyped]`.

soutaro (approved):
👍

Morriar:
Rebased on `master` 👍

PR #2175, opened 2024-12-18
Extract out usages of Ruby IDs

amomchilov:
This PR extracts out an `rbs_constant_pool` interface, which mimics the [`pm_constant_pool`](https://github.com/ruby/prism/blob/88aae15b94b514e50d608e3ce26c0f08e0baceac/include/prism/util/pm_constant_pool.h) from Prism.

I use it to wrap all usages of `ID` as location child keys, instead of directly depending Ruby APIs like `rb_intern()`, `ID2SYM`, `SYM2ID`, etc. This includes the keys used to identify the children in a `rb_loc`, and the keys used in the parser's typevar tables.

For now, this is just a dummy implementation which delegates to those same Ruby APIs, but will eventually be replaced with Prism's real logic, which doesn't depend on `ruby.h` at all.

For easier reviewing, have a look at each commit separately.

### DIFF
--- a/ext/rbs_extension/main.c
+++ b/ext/rbs_extension/main.c
@@ -1,12 +1,23 @@
 #include "rbs_extension.h"
+#include "rbs/util/rbs_constant_pool.h"
+
+#include "ruby/vm.h"
+
+static
+void Deinit_rbs_extension(ruby_vm_t *_) {
+  rbs_constant_pool_free(RBS_GLOBAL_CONSTANT_POOL);
+}
 
 void
 Init_rbs_extension(void)
 {
 #ifdef HAVE_RB_EXT_RACTOR_SAFE
   rb_ext_ractor_safe(true);
-#endif  
+#endif
   rbs__init_constants();
   rbs__init_location();
   rbs__init_parser();
+
+  rbs_constant_pool_init(RBS_GLOBAL_CONSTANT_POOL, 0);
+  ruby_vm_at_exit(Deinit_rbs_extension);
 }
--- a/ext/rbs_extension/parser.c
+++ b/ext/rbs_extension/parser.c
@@ -1031,8 +1031,13 @@ static VALUE parse_simple(parserstate *state) {
     return parse_symbol(state);
   }
   case tUIDENT: {
-    ID name = INTERN_TOKEN(state, state->current_token);
+    const char *name_str = peek_token(state->lexstate, state->current_token);
+    size_t name_len = token_bytes(state->current_token);
+
+    rbs_constant_id_t name = rbs_constant_pool_find(&state->constant_pool, (const uint8_t *) name_str, name_len);
+
     if (parser_typevar_member(state, name)) {
+      ID name = rb_intern3(name_str, name_len, rb_enc_get(state->lexstate->string));
       return rbs_variable(ID2SYM(name), rbs_location_current_token(state));
     }
     // fallthrough for type name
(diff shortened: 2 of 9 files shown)


ext/rbs_extension/parser.c
```diff
@@ -1023,8 +1031,13 @@ static VALUE parse_simple(parserstate *state) {
     return parse_symbol(state);
   }
   case tUIDENT: {
-    ID name = INTERN_TOKEN(state, state->current_token);
```
amomchilov, line 1026: This `INTERN_TOKEN` macro was calling `rb_intern3()`, which was slowly leaking memory by creating static ("immortal") symbols.

The replacement `rbs_constant_pool_find()` function uses `rb_check_id_cstr()` instead, which checks for a match without intern a new symbol if there wasn't one.

include/rbs/util/rbs_constant_pool.h
```diff
@@ -0,0 +1,94 @@
...
+ * A constant id is a unique identifier for a constant in the constant pool.
+ */
+typedef uintptr_t rbs_constant_id_t;
+
+/** A constant in the pool which effectively stores a string. */
+typedef uintptr_t rbs_constant_t;
+
+/** The overall constant pool, which stores constants found while parsing. */
+typedef struct {
+    void *dummy; // Workaround for structs not being allowed to be empty.
```
amomchilov, line 36: Eventually, this will be replaced with the real storage that will store the constant pool, instead of the one built into the Ruby VM.

src/util/rbs_constant_pool.c
```diff
@@ -0,0 +1,107 @@
...
+#else
+    _Static_assert(
+        __builtin_types_compatible_p(ID, rbs_constant_id_t),
+        "rbs_constant_id_t must be the same as a Ruby ID for now."
+    );
+
+    _Static_assert(
+        __builtin_types_compatible_p(VALUE, rbs_constant_t),
+        "rbs_constant_t must be the same as a Ruby Symbol for now."
+    );
```
amomchilov, line 22: These assertions confirm that our temporary definitions of `rbs_constant_id_t` and `rbs_constant_t` are memory-compatible with `ID` and `VALUE` (Symbol), so that we can safely cast back/forth between them. They'll go away once I add in the pure C implementation of the constant pool.

ext/rbs_extension/parser.c
soutaro: Wondering if you plan to introduce another helper function instead of writing the token manipulation code everywhere in the codebase? (maybe later.)

soutaro:
@amomchilov Thanks!

Let me confirm that: all other usage of Ruby symbols will finally be replaced to `rbs_constant_pool` to make the implementation independent to MRI C API, right?

(I'm a bit confused that some code continue using `INTERN_TOKEN` and other Ruby APIs instead of the constant pool.)

ext/rbs_extension/main.c
```diff
@@ -1,12 +1,23 @@
...
 #ifdef HAVE_RB_EXT_RACTOR_SAFE
   rb_ext_ractor_safe(true);
-#endif  
+#endif
   rbs__init_constants();
   rbs__init_location();
   rbs__init_parser();
+
+  rbs_constant_pool_init(RBS_GLOBAL_CONSTANT_POOL, 0);
+  ruby_vm_at_exit(Deinit_rbs_extension);
```
soutaro, line 22: I'm not very sure if we need the deinit function. How can a Ruby vm exit without process termination? (Maybe there’s a reason I’m not aware of…)

ParadoxV5: ~~Embedding CRuby?~~

amomchilov:
> Let me confirm that: all other usage of Ruby symbols will finally be replaced to `rbs_constant_pool` to make the implementation independent to MRI C API, right?

Eventually... yes! But that requires we introduce the C data structures (to replace the Ruby objects) and translation layer (to translate to Ruby objects, for the Ruby API), which aren't ready to ship yet. 

This PR chips away by migrating all usages related to location child names and typevars, and leaves the rest as-is for now.

ext/rbs_extension/main.c
amomchilov: I just added this to minimize the amount of leaked memory at the time the process terminates, so I could more easily find other leaks with tooling like valgrind.

We can remove it and just rely on process termination to clean up everything, but that has the downside of adding noise to any memory-related tooling. Perhaps we could add a `DEBUG` flag, and only do the cleanup when it's true?

> How can a Ruby vm exit without process termination?

Not that I know of, but it seemed like the exact API for this (ther's also `atexit()`)

soutaro: > I could more easily find other leaks with tooling like valgrind.

It makes sense! Thanks for the explanation.

soutaro (approved):
Thanks! 🎉

PR #2100, opened 2024-11-26
stdlib: Update docstrings

tk0miya:
Run `rake annotate` in the HEAD.

PR #2183, opened 2024-12-18
Generate docs based on ruby 3.4.0-rc1

soutaro:


soutaro:
Closes https://github.com/ruby/rbs/pull/2100

PR #2192, opened 2024-12-20
Should validate self-type on Proc

ksss:
Proc's self_type is not a validate target.

```rbs
class Foo
  # validate pass
  def foo: () -> ^() [self: nothing] -> void
end
```

I fixed by including `self_type` in the checks for `Types::Proc#{each_type,with_nonreturn_void?}`.


ParadoxV5:
I’ve wondered how much code does RBS reuse on methods, blocks and procs.

soutaro (approved):
🙏

PR #2197, opened 2024-12-23
Adjust capacity of location children

ksss:
I believe this is an oversight in https://github.com/ruby/rbs/pull/1994 modifications.
There is almost no change in performance.


soutaro (approved):
The prev code is not optimal but works, as `rbs_loc_add_required_child` and `rbs_loc_add_optional_child` automatically resize the capacity if needed.

PR #2211, opened 2024-12-28
docs: Add `method-interface-member` definition to syntax.md

tk0miya:
Interface definition only accepts instance methods without visibility for its members.  This separates `method-member` definition into `method-class-member` and `method-interface-member`.


soutaro (approved):
Thanks!

PR #2212, opened 2024-12-28
`#gets`s, `#readline`s and `#readlines`s *also* accept `chomp: true`

ParadoxV5:
* Sibing of #2151 😄

---

Test for `ARGF` and `StringIO` are included.
Testing `Kernel` currently unavailable [due to Broken Windows Theory](https://github.com/ruby/rbs/pull/2151#issuecomment-2558879163).
Testing `OpenSSL::Buffering` TBA… how do we use this thing?

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -112,7 +112,7 @@ end
 
 task :stdlib_test => :compile do
   test_files = FileList["test/stdlib/**/*_test.rb"].reject do |path|
-    path =~ %r{Ractor} || path =~ %r{Encoding}
+    path =~ %r{Ractor} || path =~ %r{Encoding} || path =~ %r{CGI_test}
   end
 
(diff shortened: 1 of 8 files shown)


ParadoxV5:
Here we go @soutaro. Force-pushed to fix my copy-paste and rebased to main while there

ParadoxV5:
Oh BTW,

> I tried to implement a test for `Kernel#readlines` through `ARGV`, but it blocks on the second call during the test. 😢

Have we tried subbing out `$stdin`?

---

Backstory:
I used (toplevel) `gets(chomp: true)` frequently for [Advent of Code](https://adventofcode.com) 2024.
Then just recently I realized I could use `readline` instead to skip the `nil` warning.

soutaro:
> Have we tried subbing out $stdin?

It works perfectly! Thanks. 👏 
I added a test case. (But yeah, I should have opened another PR for the test fix...)

PR #2216, opened 2024-12-30
Fix signature of IO.read and so on

KitaitiMakoto:
Hi,

I fixed the signatures of `IO.binread`, `IO.copy_stream` and `IO.read` so that they can accept `nil` as a `length` argument. It is useful when we skip known length header and read data until EOF.

Thank you.


soutaro (approved):
💯

PR #2221, opened 2025-01-06
Reimplement the RBS Constant Pool in pure C

amomchilov:
This PR replaces the Ruby-based implementation of the constant pool added in #2175, and uses a pure C implementation from Prism.

### DIFF
--- a/ext/rbs_extension/location.c
+++ b/ext/rbs_extension/location.c
@@ -168,10 +168,13 @@ static VALUE location_end_pos(VALUE self) {
   return INT2FIX(loc->rg.end);
 }
 
-static rbs_constant_id_t rbs_find_constant_id_from_ruby_symbol(VALUE symbol) {
+static rbs_constant_id_t rbs_constant_pool_insert_ruby_symbol(VALUE symbol) {
   VALUE name = rb_sym2str(symbol);
 
-  return rbs_constant_pool_find(RBS_GLOBAL_CONSTANT_POOL, (const uint8_t *) RSTRING_PTR(name), RSTRING_LEN(name));
+  // Constants inserted here will never be freed, but that's acceptable because:
+  // 1. Most symbols passed into here will be the ones already inserted into the constant pool by `parser.c`.
+  // 2. Methods like `add_required_child` and `add_optional_child` will usually only get called with a few different symbols.
+  return rbs_constant_pool_insert_constant(RBS_GLOBAL_CONSTANT_POOL, (const uint8_t *) RSTRING_PTR(name), RSTRING_LEN(name));
 }
 
 static VALUE location_add_required_child(VALUE self, VALUE name, VALUE start, VALUE end) {
@@ -181,7 +184,7 @@ static VALUE location_add_required_child(VALUE self, VALUE name, VALUE start, VA
   rg.start = rbs_loc_position(FIX2INT(start));
   rg.end = rbs_loc_position(FIX2INT(end));
 
-  rbs_loc_add_required_child(loc, rbs_find_constant_id_from_ruby_symbol(name), rg);
+  rbs_loc_add_required_child(loc, rbs_constant_pool_insert_ruby_symbol(name), rg);
 
   return Qnil;
 }
@@ -193,7 +196,7 @@ static VALUE location_add_optional_child(VALUE self, VALUE name, VALUE start, VA
   rg.start = rbs_loc_position(FIX2INT(start));
   rg.end = rbs_loc_position(FIX2INT(end));
 
-  rbs_loc_add_optional_child(loc, rbs_find_constant_id_from_ruby_symbol(name), rg);
+  rbs_loc_add_optional_child(loc, rbs_constant_pool_insert_ruby_symbol(name), rg);
 
   return Qnil;
 }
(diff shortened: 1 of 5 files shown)


ext/rbs_extension/location.c
```diff
@@ -181,7 +184,7 @@ static VALUE location_add_required_child(VALUE self, VALUE name, VALUE start, VA
   rg.start = rbs_loc_position(FIX2INT(start));
   rg.end = rbs_loc_position(FIX2INT(end));
 
-  rbs_loc_add_required_child(loc, rbs_find_constant_id_from_ruby_symbol(name), rg);
```
amomchilov, line 184: This was a mistake in the "dummy" Ruby-based implementation. The Ruby symbols were already interned by the fact they were used in Ruby code, so the lookup here always succeeded.

We need an "insert" operation here, not a "find", to ensure that if a child is added with a new name, that its name is present in the constant pool.

ext/rbs_extension/location.c
```diff
@@ -224,10 +227,16 @@ static VALUE rbs_new_location_from_loc_range(VALUE buffer, rbs_loc_range rg) {
...
+  VALUE name = rb_sym2str(symbol);
+
+  return rbs_constant_pool_find(RBS_GLOBAL_CONSTANT_POOL, (const uint8_t *) RSTRING_PTR(name), RSTRING_LEN(name));
+}
+
 static VALUE location_aref(VALUE self, VALUE name) {
   rbs_loc *loc = rbs_check_location(self);
 
-  rbs_constant_id_t id = rbs_find_constant_id_from_ruby_symbol(name);
+  rbs_constant_id_t id = rbs_constant_pool_find_ruby_symbol(name);
```
amomchilov, line 239: This is just a rename, but keeps the `find` semantics. So doing `loc[name]` with a bunch of names will _not_ permenantly leak a bunch of constants into the global pool.

ext/rbs_extension/main.c
```diff
@@ -18,6 +18,16 @@ Init_rbs_extension(void)
...
-  rbs_constant_pool_init(RBS_GLOBAL_CONSTANT_POOL, 0);
+  // Calculated based on the number of unique strings used with the `INTERN` macro in `parser.c`.
+  //
+  // ```bash
+  // grep -o 'INTERN("\([^"]*\)")' ext/rbs_extension/parser.c \
+  //     | sed 's/INTERN("\(.*\)")/\1/' \
+  //     | sort -u \
+  //     | wc -l
+  // ```
+  const size_t num_uniquely_interned_strings = 26;
```
amomchilov, line 29: In practice, the `rbs_constant_pool_init` implementation will round this up to the next power of 2 (32), so there's some wiggle room to insert a few extra constants without triggering any reallocations.

ext/rbs_extension/location.c
st0012: Since it's technically a fix, can this and other similar changes be upstreamed separately?

ext/rbs_extension/location.c
```diff
@@ -168,10 +168,13 @@ static VALUE location_end_pos(VALUE self) {
   return INT2FIX(loc->rg.end);
 }
 
-static rbs_constant_id_t rbs_find_constant_id_from_ruby_symbol(VALUE symbol) {
+static rbs_constant_id_t rbs_constant_pool_insert_ruby_symbol(VALUE symbol) {
   VALUE name = rb_sym2str(symbol);
 
-  return rbs_constant_pool_find(RBS_GLOBAL_CONSTANT_POOL, (const uint8_t *) RSTRING_PTR(name), RSTRING_LEN(name));
+  // Constants inserted here will never be freed, but that's acceptable because:
```
st0012, line 174: Should this be documented above the function itself too?

ext/rbs_extension/location.c
```diff
@@ -224,10 +227,16 @@ static VALUE rbs_new_location_from_loc_range(VALUE buffer, rbs_loc_range rg) {
   return obj;
 }
 
+static rbs_constant_id_t rbs_constant_pool_find_ruby_symbol(VALUE symbol) {
```
st0012, line 230: So it's just `rbs_find_constant_id_from_ruby_symbol` renamed and moved to a different place?

If that's the case, can we not move the location of so it's easier to review?

amomchilov: The diff was pretty messed up anyway, because every single line of it changed. So I just moved it closer to its point of use (there's only 1 caller)

ext/rbs_extension/location.c
amomchilov: It wasn't an issue in the previous implementation, and it only matters starting in this PR. It's small enough that I don't think it's worth shipping separately.

amomchilov:
Hi @soutaro, this PR is ready to review, but I think there's some issue with CI.

Two of the test runners report that the `constants.c` and `ruby_objs.c` are stale. They're not (this PR doesn't even touch those templates). Perhaps there's a stale copy of the repo on the runners, for some reason?

soutaro:
The check can be safely disabled during CI because we have `confirm_lexer` and `confirm_templates` tasks to make sure they are updated.

amomchilov:
Great @soutaro, rebased this. It's green now.

PR #2225, opened 2025-01-08
stdlib: Accept keyword arguments in `Net::HTTP.start`

euglena1215:
In the original definition, keyword arguments could not be specified without specifying all positional arguments.

Specifically, the following call results in an error with `Ruby::ArgumentTypeMismatch`.

```rb
Net::HTTP.start('www.ruby-lang.org', 443, use_ssl: true)
```

This is because opt is treated as a hash positional argument. 
This problem is solved by treating opt as a variable-length keyword argument.

ref. https://docs.ruby-lang.org/ja/latest/method/Net=3a=3aHTTP/s/start.html


soutaro:
My commit adds `Hash` positional parameter to keep the updated signature compatible with existing ones, while I don't think we frequently use the hash version...

PR #2231, opened 2025-01-13
Union#to_s should follow RBS/Lint/AmbiguousOperatorPrecedence

tk0miya:
At present, the result of `Union#to_s` does not respect `RBS/Lint/AmbiguousOperatorPrecedence` cop from rubocop-on-rbs.

It would be better to wrap intersections in unions by paranthesis.

* Before: `Integer | String & bool`
* After:  `Integer | (String & bool)`

ref: https://github.com/soutaro/rbs-inline/issues/174

### DIFF
--- a/lib/rbs/types.rb
+++ b/lib/rbs/types.rb
@@ -751,10 +751,19 @@ module RBS
       end
 
       def to_s(level = 0)
+        strs = types.map do |ty|
+          case ty
+          when Intersection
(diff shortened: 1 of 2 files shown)


tk0miya:
I posted https://github.com/ruby/rbs/pull/2232 to resolve the CI problems.

soutaro (approved):
👍

PR #2232, opened 2025-01-13
deps: Add pstore, ostruct to Gemfile

tk0miya:
pstore became a bundled gem since Ruby-3.5.  Therefore it should be added to the dependency list to pass CI.

ref: https://bugs.ruby-lang.org/issues/20309

### DIFF
--- a/test/rbs/cli_test.rb
+++ b/test/rbs/cli_test.rb
@@ -1174,7 +1174,7 @@ Processing `lib`...
           path: #{dir.join('gem_rbs_collection')}
         YAML
 
-        bundle_install('ast', 'abbrev', 'bigdecimal')
+        bundle_install('ast', 'abbrev', 'bigdecimal', 'logger')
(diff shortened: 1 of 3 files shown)


test/rbs/cli_test.rb
```diff
@@ -1174,7 +1174,7 @@ def test_collection_install
           path: #{dir.join('gem_rbs_collection')}
         YAML
 
-        bundle_install('ast', 'abbrev', 'bigdecimal')
+        bundle_install('ast', 'abbrev', 'bigdecimal', 'logger')
```
tk0miya, line 1177: Some test cases failed due to a lack of dependencies. The `logger` package has also been moved into the bundled gem since ruby-3.5.
https://github.com/ruby/rbs/actions/runs/12742248343/job/35510085716?pr=2232

soutaro (approved):
Thanks! 🙏

PR #2233, opened 2025-01-13
Fix typos

tk0miya:


lib/rbs/test/hook.rb
```diff
@@ -27,8 +27,8 @@ module Hook
         :! => "not",
         :<< => "lshift",
         :>> => "rshift",
-        :~ => "tilda",
-        :=~ => "eqtilda",
+        :~ => "tilde",
+        :=~ => "eqtilde",
```
tk0miya, line 31: Is it okay to rename them? I could not find "tilda" in other places. So I hope nothing will be broken.

soutaro (approved):
👍

PR #2234, opened 2025-01-15
Add `resolve-type-names: false` magic comment

soutaro:
The `resolved: true` magic comment allows skipping type name resolution on a file-by-file basis.

* The magic comment can be placed on the first line of RBS files.
* Writer doesn't work if the magic comment is given, and the documentation is also given from the first line.

```rbs
# resolved: true
# The class is something to explain the writer bug.
class WriterHasABug
end
```

The output will have duplicated `resolved: true` lines.

### DIFF
--- a/ext/rbs_extension/parser.c
+++ b/ext/rbs_extension/parser.c
@@ -2935,12 +2935,12 @@ parse_signature_try(VALUE a) {
 }
 
 static VALUE
-rbsparser_parse_signature(VALUE self, VALUE buffer, VALUE end_pos)
+rbsparser_parse_signature(VALUE self, VALUE buffer, VALUE start_pos, VALUE end_pos)
 {
   VALUE string = rb_funcall(buffer, rb_intern("content"), 0);
   StringValue(string);
-  lexstate *lexer = alloc_lexer(string, 0, FIX2INT(end_pos));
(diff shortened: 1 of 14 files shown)


ParadoxV5:
What’s `resolved`? How about `rbs:ignore`?

soutaro:
Loading RBS files requires resolving relative type names to absolute ones. This takes ~10 secs for bigger libraries, like aws-sdk. The `resolved: true` magic comment solves this problem, because the type definitions included in aws-sdk are generated from other type declarations. The type name resolution can be done during generation.

ParadoxV5:
`resolve-relative-types-during-generation`? `lazy-relative-resolve`?? 🤔

soutaro:
Hmm. More descriptive syntax would make sense, while I prefer the compact syntax.

```rbs
# resolved: true
# resolved-type-names: true
# no-relative-type-names: true
```

It's not *lazy* resolution nor only for *generation*.

ParadoxV5:
`resolve-type-names: false` (note lack of -`d`) sounds good. or `resolve-types`, `resolve-relative-types`…

soutaro:
Going with `resolve-type-names: false` syntax. `true` is also allowed, but it doesn't make much sense.

PR #2237, opened 2025-01-15
Validate variable types

ksss:
The `rbs validate` command now checks the type of instance/class-instance/class variables.


soutaro (approved):
👍

PR #2241, opened 2025-01-16
Check variable duplication

ksss:
* Instance variable: Only same type name
* Class instance variable: Only same type name
* Class variable: Including inheritance

### DIFF
--- a/lib/rbs/errors.rb
+++ b/lib/rbs/errors.rb
@@ -322,6 +322,51 @@ module RBS
     end
   end
 
+  class VariableDuplicationError < DefinitionError
+    include DetailedMessageable
+
+    attr_reader :member
+
+    def initialize(member:)
+      @member = member
+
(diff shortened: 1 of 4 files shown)


lib/rbs/errors.rb
```diff
@@ -322,6 +322,51 @@ def type_name
...
+    def self.check!(variables:, member:, type_name:)
+      if old = variables[member.name]
+        if old.declared_in == type_name
+          raise new(member: member)
+        end
+      end
+    end
+  end
+
+  class ClassVariableDuplicationError < VariableDuplicationError
```
ksss, line 362: Respect https://github.com/soutaro/steep/blob/5f40fc1c599476cf31259c6402746e9b31d14727/lib/steep/diagnostic/signature.rb#L276

soutaro (approved):
:+1:

PR #2249, opened 2025-01-21
Remove call to `Any#todo!` from C parser

Morriar:
As we're trying to detangle the C parser implementation from Ruby, the call to `Any#todo!` is problematic.

Instead we can store a flag in the C structure and move this call to the Ruby implementation during the object initialization.

### DIFF
--- a/lib/rbs/types.rb
+++ b/lib/rbs/types.rb
@@ -107,9 +107,16 @@ module RBS
       class Bool < Base; end
       class Void < Base; end
       class Any < Base
+        def initialize(location:, todo: false)
+          super(location: location)
+          todo! if todo
+        end
+
         def to_s(level=0)
           @string || "untyped"
         end
+
+        private
 
         def todo!
           @string = '__todo__'
--- a/sig/types.rbs
+++ b/sig/types.rbs
@@ -113,6 +113,10 @@ module RBS
       class Any < Base
         @string: String?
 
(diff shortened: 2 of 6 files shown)


lib/rbs/types.rb
```diff
@@ -107,10 +107,17 @@ def with_nonreturn_void?
...
+        def initialize(location:, todo: false)
+          super(location: location)
+          todo! if todo
+        end
+
         def to_s(level=0)
           @string || "untyped"
         end
 
+        private
```
Morriar, line 119: The `private` here is not required for the change to work but `todo!` shouldn't be called from clients.

This will impact the public API, maybe we should make a deprecation first?

sig/types.rbs
```diff
@@ -113,6 +113,10 @@ module RBS
       class Any < Base
         @string: String?
 
+        def initialize: (location: Location[bot, bot]?, ?todo: bool) -> void
```
soutaro, line 116: 👍

lib/rbs/types.rb
soutaro: Yeah, can you keep the method and print a deprecation message?

soutaro (changes requested):
Looks good. The requested change is to keep the `#todo!` method public with deprecation.

Morriar:
I removed the `private`, added a deprecation comment in the Ruby file and a deprecation annotation in the RBS file 👍

soutaro (approved):
👍

PR #2250, opened 2025-01-22
Rename to reduce top-level interface and type alias

ksss:
Type aliases and interfaces cannot overlap.
Therefore, if they are defined at the top level, they affect all RBS files being loaded, so caution is required when using them.

I moved the four interfaces and four type aliases defined in `json` under the `JSON` module.

This change is incompatible, and any type definitions already using these signatures will need to be updated.
However, in the scope of my investigation, I did not find any usage of these.

As for `_ToJson`, it is also used in the RBS library, so I believe it is fine to leave it as it is.

### DIFF
--- a/stdlib/json/0/json.rbs
+++ b/stdlib/json/0/json.rbs
@@ -94,7 +94,6 @@ end
 class JSON::Pure::Parser
 end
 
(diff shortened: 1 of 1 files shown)


soutaro (approved):
Okay, give it a try! 👍

PR #2256, opened 2025-01-24
Remove call to `TypeParam#unchecked!` from C parser

Morriar:
Similar to #2249, as we're trying to detangle the C parser implementation from Ruby, the call to `TypeParam#unchecked!` is problematic.

Instead we can store a flag in the C structure and move this call to the Ruby implementation during the object initialization.


soutaro (approved):
👍

PR #2257, opened 2025-01-26
[rbs/unit_test] Treat nil as a return value

ksss:
On the Ruby side, when `nil` is returned, `rbs/unit_test` interprets it as a break, which revealed that the return value was not being checked.
I have fixed this issue and also fixed the tests that depended on this issue.


soutaro (approved):
Thanks! 🙏

PR #2264, opened 2025-01-31
Fix UNIXSocket#send_io and recv_io types

pocke:
This PR fixes the method type definitions for `UNIXSocket#{send_io,recv_io}`.

## send_io

https://docs.ruby-lang.org/en/3.3/UNIXSocket.html#method-i-send_io

`send_io` actually receives an IO, but the type definition only accepts `BasicSocket`, which is a subclass of `IO`. So I fixed it to allow `IO`.

## recv_io

https://docs.ruby-lang.org/en/3.3/UNIXSocket.html#method-i-recv_io

It's a bit complex. It has three patterns.

* Without argument, it returns an `IO`.
* With the `Integer` class or nil, it returns an Integer.
* With any object having `for_fd` method, it calls this method and returns the returned value of `for_fd`.
  * The first argument of `for_fd` is the file descripter. The second argument is `mode`, and the second argument of `recv_io`.


core/io.rbs
```diff
@@ -3251,6 +3251,10 @@ class IO < Object
   #     f.close
   #
   alias to_i fileno
+
+  interface _ForFd[RET]
+    def for_fd: (?) -> RET
```
pocke, line 3256: The argument is rough.

Actually, the definition should be the following:


```rbs
interface _ForFd[MODE, RET]
  def for_fd: (Integer fd, ?MODE) -> RET
end
```

But the definition is wrong because the following class does not match the interface.

```rbs
class ForFd
  def self.for_fd: (Integer fd) -> instance
end
```

soutaro (approved):
👍

PR #2267, opened 2025-02-05
Fix type of `OpenSSL::BN` methods

ksss:
Several fixes to `OpenSSL::BN` method argument types.
All type errors were found by raap.
Therefore, the test should be run by raap.
For cases that are difficult to express with raap (e.g., numeric or string constraints), we will continue testing them with rbs/unit_test.

Additionally, I separated the executable files to allow running raap with different options.


soutaro (approved):
👍

PR #2273, opened 2025-02-11
core: Array#include? and Enumerable#include? should take non-Elem types

tk0miya:
For example, `Array[Integer]` can take other number types validly:

```
> [1, 2, 3].include? 2.0
=> true
> [1, 2, 3].include? Complex(2)
=> true
> [1, 2, 3].include? Rational(2)
=> true
```

refs: https://github.com/soutaro/steep/issues/1482

### DIFF
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -2182,7 +2182,7 @@ class Array[unchecked out Elem] < Object
   #
   # Related: see [Methods for Querying](rdoc-ref:Array@Methods+for+Querying).
   #
-  def include?: (Elem object) -> bool
+  def include?: (untyped object) -> bool
 
   # <!-- rdoc-file=array.c -->
   # Returns the zero-based integer index of a specified element, or `nil`.
--- a/core/enumerable.rbs
+++ b/core/enumerable.rbs
@@ -718,7 +718,7 @@ module Enumerable[unchecked out Elem] : _Each[Elem]
   #     {foo: 0, bar: 1, baz: 2}.include?('foo') # => false
   #     {foo: 0, bar: 1, baz: 2}.include?(0)     # => false
   #
-  def include?: (Elem arg0) -> bool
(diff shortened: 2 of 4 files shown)


ksss:
In reality, it seems that the `==` method is being used as an interface.

```rb
irb(main):005> TracePoint.new(:a_call) { |tp| p tp }.enable { [1, 2, 3].include? 2.0 }
#<TracePoint:b_call (irb):5>
#<TracePoint:c_call 'include?' (irb):5>
#<TracePoint:c_call '==' (irb):5>
#<TracePoint:c_call '==' (irb):5>
=> true
```

However, since `==` is a method of `BasicObject`, it applies to almost all objects.
Therefore, using `untyped` is fine.

core/array.rbs
```diff
@@ -2182,7 +2182,7 @@ class Array[unchecked out Elem] < Object
   #
   # Related: see [Methods for Querying](rdoc-ref:Array@Methods+for+Querying).
   #
-  def include?: (Elem object) -> bool
+  def include?: (untyped object) -> bool
```
tk0miya, line 2185: Initially, I thought to allow taking nil value. Therefore I changed this type as follows once:

```suggestion
  def include?: (Elem? object) -> bool
              | (untyped object) -> false
```

But I noticed it does not help the numbers' case. Therefore, I used `untyped` here finally.

soutaro:
I prefer keeping the *strict* method type, because finding a code passing an unexpected object to `include?` actually helps.

Can you share the code where you want this? @tk0miya

soutaro:
Ah, https://github.com/soutaro/steep/issues/1482... Hmm...

soutaro:
No conclusion yet, though I'm not very positive about this change.

soutaro:
Merged in https://github.com/ruby/rbs/pull/2878

PR #2274, opened 2025-02-12
Reduce object creation during substitution

ksss:
The `#sub` method generates many objects, but in most classes, substitution does not occur at all. By modifying it so that no objects are created when there are no variables to substitute, we will optimize performance.

A side effect is that `#sub` no longer guarantees that a new object will be created.

## Benchmark

```rb
# validate.rb

require 'rbs'
require 'rbs/cli'
# require 'memory_profiler'

# MemoryProfiler.report do
  loader = RBS::EnvironmentLoader.new
  Dir["stdlib/*"].each do |path|
    lib = File.basename(path).to_s
    loader.add(library: lib, version: nil)
  end
  options = Data.define(:loader).new(loader:)
  RBS::CLI::Validate.new(args: [], options:).run
# end.pretty_print
```

```
$ time bundle exec ruby validate.rb
```

### Before

```
bundle exec ruby validate.rb  1.22s user 0.13s system 86% cpu 1.555 total
bundle exec ruby validate.rb  1.23s user 0.13s system 86% cpu 1.574 total
bundle exec ruby validate.rb  1.23s user 0.13s system 86% cpu 1.576 total
```

### After

```
bundle exec ruby validate.rb  1.00s user 0.13s system 82% cpu 1.374 total
bundle exec ruby validate.rb  1.00s user 0.13s system 84% cpu 1.334 total
bundle exec ruby validate.rb  1.00s user 0.13s system 84% cpu 1.335 total
```

### Before

```
Total allocated: 309209155 bytes (3277248 objects)
Total retained:  191993 bytes (3441 objects)
```

### After

```
Total allocated: 190730219 bytes (1904943 objects)
Total retained:  191993 bytes (3441 objects)
```

### DIFF
--- a/lib/rbs/ast/members.rb
+++ b/lib/rbs/ast/members.rb
@@ -30,6 +30,8 @@ module RBS
           end
 
           def sub(subst)
+            return self if subst.empty?
+
             update(method_type: self.method_type.sub(subst))
           end
 
--- a/lib/rbs/definition.rb
+++ b/lib/rbs/definition.rb
@@ -14,6 +14,8 @@ module RBS
       end
 
       def sub(s)
+        return self if s.empty?
+
         self.class.new(
           parent_variable: parent_variable,
           type: type.sub(s),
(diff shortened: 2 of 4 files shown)


soutaro (approved):
👍

Issue #2258, opened 2025-01-27
The documentation for the bundled gem has not been updated


ksss:
The `rake annotation` task generates documentation for RBS from the ruby/ruby source code. However, bundled gems are not included in the ruby/ruby source code. As a result, the `rake annotate` task that updates documentation does not update the documentation for bundled gems.

I think we need to decide on how to approach the documentation for bundled gems.
 
* minitest
* net-smtp
* base64
* bigdecimal
* abbrev
* nkf
* csv

ParadoxV5:
* Ruby core: We supply their RBS in RBS-gem.
* Non-bundled gems: Their RBS are either self-managed or, less ideally, community-managed at [ruby/gem_rbs_collection](https://github.com/ruby/gem_rbs_collection).

Default and bundled gems needs to pick a side.

I personally prefer both to self-manage RBS in their own packages.
* Separate repositories for default gems was a relatively new convention that [doesn’t quite apply de facto](https://bugs.ruby-lang.org/issues/18169).
* Bundled gems are not tied to ruby/ruby though, so maybe it makes sense for the RBS to rather be in gem_rbs_collection at least for the time beïng, considering the uninstallation these (helpful!) bundleware isn’t restricted?

ksss:
Discussed with the committers (@soutaro @pocke @mame) and decided on the following policy.

* Move RBS of Bundled gem to gem_rbs_collection.
* Move to gem if management by gem is approved.
* Deprecation warning if it is written in manifest.yaml.

ksss:
### TODO

- [] Add bundled gems to gem_rbs_collection
- [] Remove bundled gems from rbs/stdlib and add gems to `RBS::Collection::Config::LockfileGenerator::ALUMNI_STDLIBS`

PR #2288, opened 2025-02-21
Add bundled gems to alumnus 

ksss:
Prepare to graduate bundled gems from the rbs repo.

The newly added gem has not been included in the gem repo itself, so its version cannot be specified.
Therefore, I set it to `nil`. Once it is added to the gem, I plan to specify the version accordingly.  

Since `bigdecimal-math` does not exist as a gem and is unlikely to exist in the future, adding it to `ALUMNI_STDLIBS` or not will lead to the same issue: when `bigdecimal-math` is removed from the RBS repository, it will fail midway because it cannot find a gem with the same name.  

To handle this, `bigdecimal-math` will receive special treatment—removing the specification will be encouraged, while still using it if it remains in the standard library.

## refs

* https://github.com/ruby/rbs/issues/2258
* https://github.com/ruby/gem_rbs_collection/pull/782

### DIFF
--- a/lib/rbs/collection/config/lockfile_generator.rb
+++ b/lib/rbs/collection/config/lockfile_generator.rb
@@ -4,7 +4,17 @@ module RBS
   module Collection
     class Config
       class LockfileGenerator
-        ALUMNI_STDLIBS = { "mutex_m" => ">= 0.3.0" }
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍

PR #2289, opened 2025-02-21
Validate superclass and module-self-type

ksss:
I propose to validate for type application on the superclass and modules-self-type.
This change will allow users to quickly recognize this type specification error.

## Note

This change will cause Unexpected error in steep and needs to be corrected.
This is because `RBS::InvalidTypeApplicationError` is not expected to occur in certain cases.
https://github.com/soutaro/steep/blob/5705673a6ef631771cda36a77b4dc7c0294a3941/lib/steep/services/signature_service.rb#L297-L298


soutaro (approved):
👍

PR #2292, opened 2025-02-24
Apply rubocop style to RBS

ksss:
* [RBS/Style/InstanceWithInstance](https://github.com/ksss/rubocop-on-rbs/blob/main/docs/modules/ROOT/pages/cops_rbs_style.adoc#rbsstyleinstancewithinstance)
    * https://github.com/ruby/rbs/pull/2139
    * https://github.com/ruby/rbs/pull/2174
    * Use `self` instead of `instance` in instance context without generics class or module
* [RBS/Style/ClassWithSingleton](https://github.com/ksss/rubocop-on-rbs/blob/main/docs/modules/ROOT/pages/cops_rbs_style.adoc#rbsstyleclasswithsingleton)
    * Use `self` instead of `class` in singleton context.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -46,6 +46,8 @@ RBS/Style:
     - 'test/**/*'
 RBS/Style/BlockReturnBoolish:
   Enabled: true
+RBS/Style/ClassWithSingleton:
+  Enabled: true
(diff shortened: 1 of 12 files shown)


soutaro (approved):
Give the cops a try! 👍

PR #2294, opened 2025-02-25
Exclude `attr_*` methods from duplicate checks.

ksss:
I will fix this since unintended duplicate checks were being performed.

## Before

```rb
# Error (not expected)
class Foo
  attr_accessor instance: Integer
  @instance: Integer
end

# No Error (expected)
class Foo
  @instance: Integer
  attr_accessor instance: Integer
end

# Error (expected)
class Foo
  @instance: Integer
  @instance: Integer
end
```

## After

```rb
# No Error (expected)
class Foo
  attr_accessor instance: Integer
  @instance: Integer
end

# No Error (expected)
class Foo
  @instance: Integer
  attr_accessor instance: Integer
end

# Error (expected)
class Foo
  @instance: Integer
  @instance: Integer
end
```

### DIFF
--- a/lib/rbs/definition.rb
+++ b/lib/rbs/definition.rb
@@ -6,11 +6,13 @@ module RBS
       attr_reader :parent_variable
       attr_reader :type
       attr_reader :declared_in
+      attr_reader :source
 
-      def initialize(parent_variable:, type:, declared_in:)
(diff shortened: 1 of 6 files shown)


soutaro (approved):
Thanks! 👍

PR #2300, opened 2025-03-04
Fix class variable

soutaro:
This PR fixes class-variable accessibility, that is defined in modules and mixed-in to a class.

Class variables in modules are accessible to the class when the module is mixed-in by `include` or `prepend`, but not accessible when the module is `extend`-ed.

### DIFF
--- a/lib/rbs/definition_builder.rb
+++ b/lib/rbs/definition_builder.rb
@@ -82,7 +82,7 @@ module RBS
       AST::TypeParam.application(params, args) || Substitution.new()
     end
 
-    def define_instance(definition, type_name, subst)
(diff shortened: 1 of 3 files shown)


lib/rbs/definition_builder.rb
```diff
@@ -260,7 +261,7 @@ def build_singleton0(type_name)
             end
 
             subst = tapp_subst(mod.name, mod.args)
-            define_instance(definition, mod.name, subst)
+            define_instance(definition, mod.name, subst, define_class_vars: false)
```
soutaro, line 264: Importing modules which Is `extend`-ed should not define class variables.

PR #2304, opened 2025-03-05
Reduce Array object allocation during parsing

ksss:
The most basic function for parsing a type, `parse_type`, always generates two Array objects each time it is executed.
However, for simple types, this Array object is not always needed, and the object allocation becomes unnecessary.

In this PR, I will make modifications to minimize unnecessary object allocations in `parse_type` as much as possible.

## Profile script

```rb
require 'rbs'
require 'rbs/cli'
require 'memory_profiler'

loader = RBS::EnvironmentLoader.new
Dir["stdlib/*"].each do |path|
  lib = File.basename(path).to_s
  loader.add(library: lib, version: nil)
end

MemoryProfiler.report do
  RBS::Environment.from_loader(loader)
end.pretty_print
```

Before

```
Total allocated: 49914684 bytes (453299 objects)
allocated memory by class: 3219256  Array
allocated objects by class: 70648  Array
```

After

```
Total allocated: 48873724 bytes (427275 objects)
allocated memory by class: 2178296  Array
allocated objects by class: 44624  Array
```

### DIFF
--- a/ext/rbs_extension/parser.c
+++ b/ext/rbs_extension/parser.c
@@ -1092,10 +1092,13 @@ static VALUE parse_intersection(parserstate *state) {
   rg.start = state->next_token.range.start;
 
   VALUE type = parse_optional(state);
-  VALUE intersection_types = rb_ary_new();
+  VALUE intersection_types = EMPTY_ARRAY;
 
-  rb_ary_push(intersection_types, type);
   while (state->next_token.type == pAMP) {
+    if (intersection_types == EMPTY_ARRAY) {
+      melt_array(&intersection_types);
+      rb_ary_push(intersection_types, type);
+    }
     parser_advance(state);
     rb_ary_push(intersection_types, parse_optional(state));
   }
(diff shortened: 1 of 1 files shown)


ext/rbs_extension/parser.c
```diff
@@ -1092,10 +1092,13 @@ static VALUE parse_intersection(parserstate *state) {
   rg.start = state->next_token.range.start;
 
   VALUE type = parse_optional(state);
-  VALUE intersection_types = rb_ary_new();
+  VALUE intersection_types = EMPTY_ARRAY;
 
-  rb_ary_push(intersection_types, type);
   while (state->next_token.type == pAMP) {
+    if (intersection_types == EMPTY_ARRAY) {
```
soutaro, line 1098: We can early return before allocating `intersection_types`, and I think it's better code organization.

```c
...

VALUE type = parse_optional(state);

if (state->next_token.type != pAMP) {
  return type;
}

VALUE intersection_types = rb_ary_new();
rb_ary_push(intersection_types, type);

...
```

soutaro (approved):
👍

PR #2309, opened 2025-03-07
Run Valgrind on CI and fix memory leaks

st0012:
This PR:

- Sets up [Valgrind](https://valgrind.org/) through [ruby_memcheck](https://github.com/Shopify/ruby_memcheck), which is also used by `prism` and `json` and [many other projects](https://github.com/Shopify/ruby_memcheck?tab=readme-ov-file#success-stories)
    - We need to skip 2 test cases when running with Valgrind due to https://bugs.ruby-lang.org/issues/21173
- Fixes the 2 leaks Valgrind detects

(Big thanks to @peterzhu2118 for helping on both the setup and fixing the issues 🙏 )

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,10 @@ require "rake/testtask"
 require "rbconfig"
 require 'rake/extensiontask'
 
+on_windows = /mswin|mingw/ =~ RUBY_PLATFORM
+
+require "ruby_memcheck" if !on_windows
+
 $LOAD_PATH << File.join(__dir__, "test")
 
 ruby = ENV["RUBY"] || RbConfig.ruby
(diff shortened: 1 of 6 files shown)


Rakefile
```diff
@@ -11,14 +15,22 @@ bin = File.join(__dir__, "bin")
...
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList["test/**/*_test.rb"].reject do |path|
     path =~ %r{test/stdlib/}
   end
 end
 
+Rake::TestTask.new(test: :compile, &test_config)
+
+if !on_windows
```
amomchilov, line 28:
```suggestion
unless on_windows
```

amomchilov (commented):
Nice!

Rakefile
```diff
@@ -3,6 +3,10 @@ require "rake/testtask"
 require "rbconfig"
 require 'rake/extensiontask'
 
+on_windows = /mswin|mingw/ =~ RUBY_PLATFORM
```
Morriar, line 6: It seems there are more patterns, could we use `Gem.win_platform?` instead? https://github.com/rubygems/rubygems/blob/master/lib/rubygems.rb#L1028-L1035

Rakefile
```diff
@@ -11,14 +15,22 @@ bin = File.join(__dir__, "bin")
...
-Rake::TestTask.new(:test => :compile) do |t|
+test_config = lambda do |t|
   t.libs << "test"
   t.libs << "lib"
   t.test_files = FileList["test/**/*_test.rb"].reject do |path|
     path =~ %r{test/stdlib/}
   end
 end
 
+Rake::TestTask.new(test: :compile, &test_config)
```
Morriar, line 26: What do you think about moving the check and the require here so everything is close together?

st0012: Updated 👍

soutaro (approved):
Thanks! 🎉

PR #2310, opened 2025-03-08
Add information for VariableDuplicationError

ksss:
To provide a more detailed message, I added information about which class/module the variable is defined in.
I would also like to offer a similar message at Steep.

```
$ bundle exec rbs -I sample.rbs validate
E, [2025-03-08T16:53:11.061295 #80748] ERROR -- rbs: sample.rbs:3:2...3:15: Duplicated instance variable name `@foo` in `::Foo` (RBS::InstanceVariableDuplicationError)

    @foo: Integer
    ^^^^^^^^^^^^^
```

```
$ bundle exec rbs -I sample.rbs validate
E, [2025-03-08T16:53:28.271229 #80802] ERROR -- rbs: sample.rbs:3:2...3:20: Duplicated class instance variable name `@foo` in `::Foo` (RBS::ClassInstanceVariableDuplicationError)

    self.@foo: Integer
    ^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/lib/rbs/definition_builder.rb
+++ b/lib/rbs/definition_builder.rb
@@ -569,11 +569,11 @@ module RBS
       case l.source
       when AST::Members::InstanceVariable
         if r.source.instance_of?(AST::Members::InstanceVariable) && l.declared_in == r.declared_in
-          raise InstanceVariableDuplicationError.new(member: l.source)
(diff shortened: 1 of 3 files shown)


soutaro (approved):
👍

PR #2315, opened 2025-03-11
Add return types for callbacks in Ripper::SexpBuilder

apiology:


apiology:
I wasn't quite sure how to test callbacks like these with your tooling - open to ideas!

stdlib/ripper/0/ripper.rbs
```diff
@@ -458,382 +458,382 @@ class Ripper
     attr_reader error: untyped
 
     alias compile_error on_error
-    def on_BEGIN: (*untyped args) -> untyped
+    def on_BEGIN: (*untyped args) -> Array
 
-    def on_CHAR: (untyped tok) -> untyped
+    def on_CHAR: (untyped tok) -> [Symbol, String, [int, int]]
```
apiology, line 463: These two patterns of callbacks are defined here:

https://github.com/ruby/ruby/blob/master/ext/ripper/lib/ripper/sexp.rb#L116-L131

stdlib/ripper/0/ripper.rbs
```diff
@@ -458,382 +458,382 @@ class Ripper
     attr_reader error: untyped
 
     alias compile_error on_error
-    def on_BEGIN: (*untyped args) -> untyped
+    def on_BEGIN: (*untyped args) -> Array[untyped]
 
-    def on_CHAR: (untyped tok) -> untyped
+    def on_CHAR: (String tok) -> [Symbol, String, [Integer?, Integer?]]
```
apiology, line 463: These two patterns of callbacks are defined here:

https://github.com/ruby/ruby/blob/master/ext/ripper/lib/ripper/sexp.rb#L116-L131

apiology:
This change is limited as I'm not terribly familiar with Ripper.  These changes will help https://github.com/castwide/solargraph pass its own typechecking; here's the use it makes of Ripper: https://github.com/castwide/solargraph/blob/41b892397be99fd6b445d1e308a3f512c5ec32e9/lib/solargraph/parser/comment_ripper.rb#L4

cc: @aamine in case you have thoughts

apiology:
Should I push up additional fixes to RBS types?

apiology:
I'm sorry - I don't know how to test callbacks like these with your tooling.  Do you have any ideas?  A small example and I can fill in the rest.

ksss (commented):
Since the return values of all callback methods are never used, `untyped` or `void` seems appropriate.  
Callback methods are meant to be overridden in subclasses, so perhaps they shouldn't be defined at all.  

I wrote the original signatures myself, but I honestly have no memory of it...

apiology:
> Since the return values of all callback methods are never used, untyped or void seems appropriate.
> Callback methods are meant to be overridden in subclasses, so perhaps they shouldn't be defined at all.

The return values actually are used - they determine what is returned by the overall parse() method - see https://github.com/ruby/ruby/blob/a4a60195502add094fb52a587411bbd0c19facce/ext/ripper/lib/ripper/filter.rb#L59

As you note, Ripper::SexpBuilder is used by subclassing.  The return types indicate to the developer what they must return from the methods they write as well as what information they can gain from the superclass method's return value.

Here is the code I would like to be well-typed: https://github.com/castwide/solargraph/blob/0e82bbdcc9da7343dd028373797f2fe0373dbf6a/lib/solargraph/parser/comment_ripper.rb - and here is how the resulting code is used: https://github.com/castwide/solargraph/blob/767a5fceb74935d03f6757711bdf610df3e6fd7a/lib/solargraph/parser/parser_gem/class_methods.rb#L17

> I wrote the original signatures myself, but I honestly have no memory of it...

I am very familiar with that feeling 😁

ksss:
## `Ripper::Filter`

After looking into it, it seems that for `Ripper::Filter`, it would be appropriate to restrict the `on_*` methods to return the same type as the second `data` argument.

```rbs
# ripper.rbs
class Ripper
  class Filter[T = untyped]
    def on_kw: (untyped tok, T data) -> T
  end
end

class Foo < Ripper::Filter[Array[String]]
end
```

```rb
# foo.rb
class Foo < Ripper::Filter
  def on_kw(tok, data)
    # `data` is `T(Array[String])`
    data << "aaa"
    nil # typecheck should fail because `on_kw` return type is `T(Array[String])`
  end
end
```

## `Ripper::SexpBuilderPP`

As for `Ripper::SexpBuilderPP`, even though its return values aren’t actually used, adding type restrictions to the overridden methods would just make things less practical.
In the first place, `Ripper::SexpBuilderPP` is an undocumented internal class, so I think most of its methods shouldn’t be explicitly defined at all.

Issue #558, opened 2020-12-31
Using `singleton` on variables


grncdr:
It would be useful to be able to apply `singleton` to type variables as well as concrete types. As a motivating example consider this simple event bus:

```rb
interface _Subscriber[E]
  def call: (E event) -> void
end

class EventBus
  def subscribe: [E] (_Subscriber[E] subscriber, to: singleton(E)) -> void
  def dispatch: (untyped event) -> void
end
```

Here I want to specify that `EventBus#subscribe` expects the `subscriber` to handle events of the type given as the `to:` argument.

mikhliuk-k:
I faced such problem too

For example, in my case I would like to do smth like:
```ruby
class Builder
  def self.build: [C] (singleton(C) class) -> Array[C]
end
```

PR #2324, opened 2025-03-14
Fix block self types of `define_singleton_method`

soutaro:


core/module.rbs
```diff
@@ -731,8 +731,8 @@ class Module < Object
   #     I'm Dino!
   #     #<B:0x401b39e8>
   #
-  def define_method: (interned symbol, ^() [self: self] -> untyped | Method | UnboundMethod method) -> Symbol
-                   | (interned symbol) { () [self: self] -> untyped } -> Symbol
+  def define_method: (interned symbol, ^() [self: top] -> untyped | Method | UnboundMethod method) -> Symbol
```
soutaro, line 734: We cannot have meaningful self type here because the type of `self` inside the block is the instance of modules that is an instance of `Module`. `instance` is `Module`, `self` is also `Module`.

Writing `top` here enforces writing type annotations for `self` -- `# @type self: T` in Steep.

ParadoxV5: * It’ll have to be `singleton(self)` similar to #558.
* Or #1542.

PR #2336, opened 2025-03-18
Implement context validation on parse

ksss:
Contextual limitations are now performed at parse time.

This change was originally planned for version 3.4 but remained unaddressed.

> We plan to change the parser to reject those types if it breaks the contextual limitations in next release -- `3.4`.

I propose introducing it with the release of version 4.0.

## Changes

- The check will only be performed in `Parser.parse_signature`.
- `RBS::UnitTest`, they have not been removed in this PR.
- Deprecate `--[no-]exit-error-on-syntax-error` option.
- Raise `RBS::ParsingError` instead of `RBS::WillSyntaxError`.
- Remove `RBS::WillSyntaxError`.
- Contextual limitations checks in the `rbs validate` command have been stopped.
- Gave up analyzing `CONST: self` in the `rbs prototype rb`.

## Not changes

- I have not changed that `RBS::ParsingError` is raised for compatibility with Steep. It may be possible to use a different error name.
- `Parser.parse_type` and `Parser.parse_method_type` do not currently perform checks. It might be worth adding them.
- Since `has_self_type?`, `has_classish_type?`, and `with_nonreturn_void?` are used in `RBS::UnitTest`, they have not been removed in this PR.

### DIFF
--- a/include/rbs/parser.h
+++ b/include/rbs/parser.h
@@ -127,7 +127,7 @@ rbs_ast_comment_t *rbs_parser_get_comment(rbs_parser_t *parser, int subject_line
 void rbs_parser_set_error(rbs_parser_t *parser, rbs_token_t tok, bool syntax_error, const char *fmt, ...) RBS_ATTRIBUTE_FORMAT(4, 5);
 
 /**
- * rbs_type_validation_t represents the validation rules for type parsing.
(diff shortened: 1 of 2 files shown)


ksss:
Rebased on master.

ksss:
Thank you for reviewing.
I'll fix soon.

PR #2355, opened 2025-03-24
Update rdoc to 6.13

ksss:
There are several breaking changes in RDoc.

* Interface of `RDoc::Store.new` has been changed.
    * https://github.com/ruby/rdoc/pull/1309
* Return value of `RDoc::Attr#comment` has been changed from `RDoc::Markup::Document` to `RDoc::Comment`.
    * https://github.com/ruby/rdoc/pull/1210
* Interface of `RDoc::Alias.new` has been changed
    * https://github.com/ruby/rdoc/pull/1312

This PR limits rdoc to **6.13** or higher to support these breaking changes.

### DIFF
--- a/lib/rbs/annotate/formatter.rb
+++ b/lib/rbs/annotate/formatter.rb
@@ -81,10 +81,17 @@ module RBS
         end
       end
 
-      def self.translate(doc)
-        if doc.file
+      def self.translate(doc_or_comment)
+        if doc_or_comment.file
           formatter = RDoc::Markup::ToMarkdown.new
-          doc.accept(formatter).strip.lines.map(&:rstrip).join("\n")
+          case doc_or_comment
+          when RDoc::Markup::Document
+            doc_or_comment
+          when RDoc::Comment
+            doc_or_comment.parse
+          else
+            raise "Unexpected comment class: #{doc_or_comment.class}"
+          end.accept(formatter).strip.lines.map(&:rstrip).join("\n")
         end
       end
     end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,6 +3,7 @@ PATH
   specs:
     rbs (4.0.0.dev)
       logger
+      rdoc (>= 6.13)
 
 PATH
   remote: test/assets/test-gem
@@ -93,7 +94,7 @@ GEM
     rb-fsevent (0.11.2)
     rb-inotify (0.11.1)
       ffi (~> 1.0)
-    rdoc (6.11.0)
+    rdoc (6.13.0)
       psych (>= 4.0.0)
     regexp_parser (2.10.0)
     rspec (3.13.0)
--- a/lib/rbs/annotate/rdoc_source.rb
+++ b/lib/rbs/annotate/rdoc_source.rb
@@ -23,7 +23,7 @@ module RBS
         @stores.clear()
 
         RDoc::RI::Paths.each(with_system_dir, with_site_dir, with_home_dir, with_gems_dir ? :latest : false, *extra_dirs.map(&:to_s)) do |path, type|
-          store = RDoc::Store.new(path, type)
+          store = RDoc::Store.new(RDoc::Options.new, path:, type:)
           store.load_all
 
           @stores << store
--- a/lib/rdoc_plugin/parser.rb
+++ b/lib/rdoc_plugin/parser.rb
@@ -86,7 +86,7 @@ module RBS
       end
 
       def parse_method_alias_decl(decl:, context:, outer_name: nil)
-        alias_def = RDoc::Alias.new(nil, decl.old_name.to_s, decl.new_name.to_s, nil, decl.kind == :singleton)
+        alias_def = RDoc::Alias.new(nil, decl.old_name.to_s, decl.new_name.to_s, nil, singleton: decl.kind == :singleton)
         alias_def.comment = construct_comment(context: context, comment: comment_string(decl)) if decl.comment
         context.add_alias(alias_def)
       end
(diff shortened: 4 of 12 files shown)


lib/rbs/annotate/formatter.rb
```diff
@@ -81,10 +81,17 @@ def self.each_part(doc, &block)
...
-        if doc.file
+      def self.translate(doc_or_comment)
+        if doc_or_comment.file
           formatter = RDoc::Markup::ToMarkdown.new
-          doc.accept(formatter).strip.lines.map(&:rstrip).join("\n")
+          case doc_or_comment
+          when RDoc::Markup::Document
+            doc_or_comment
+          when RDoc::Comment
+            doc_or_comment.parse
```
tompng, line 91: I think the test code `RBS::Annotate::Formatter.translate(something.comment)` should be fixed instead of accepting comment here.

`def self.each_part` (in line 60) can be also simplified.
```diff
       def self.each_part(doc, &block)
         if block
-          document =
-            case doc
-            when String
-              raise
-            when RDoc::Comment
-              document = doc.parse
-            when RDoc::Markup::Document
-              document = doc
-            end
+          document = doc


-  Formatter.each_part(subject.comment) do |doc|
+  raise if subject.comment.is_a?(String)
+  Formatter.each_part(subject.comment.parse) do |doc|
```
With some type modification
```diff
   interface _WithRDocComment
-    def comment: () -> (RDoc::Markup::Document | RDoc::Comment | String)
+    def comment: () -> (RDoc::Comment | String)
   end

   class CodeObject
-    attr_reader comment: Markup::Document | Comment | String
+    attr_reader comment: Comment | String
 
-    def comment=: (Markup::Document | Comment | String | nil) -> (Markup::Document | Comment | String)
+    def comment=: (Comment | String | nil) -> (Comment | String | nil)
    end
```

st0012:
It looks like rbs uses rdoc (and ri) for 2 things:
- Locate generated ri document
- Read the comments from RDoc classes/modules/methods...etc. that's constructed by the `Store` instance

If that's correct, I'd recommend using Prism to write a simple visitor to index the comments. Here's a good [example](https://github.com/Shopify/ruby-lsp/blob/main/lib/ruby_indexer/lib/ruby_indexer/declaration_listener.rb) from Ruby LSP. Since all you need are comments, the implementation could be much simpler than it as you won't need to care about stuff like visibility...etc.

It will take some effort but will have a few benefits:
- It'll be much less likely to break as prism is built for providing these information reliably and always be up-to-date with the latest Ruby features.
    - rdoc, on the other hand, is not built to be a document source provider library as it currently is used. It's a understandable choice before prism, but now rdoc needs to focus on providing missing features. And as a result it won't be able to maintain the desired stability across most of its internal components.
- It will be easier to work with as rdoc doesn't have the best design in terms of data representation (handling both `RDoc::Comment` and `String` in multiple places is just unnecessary, which we want to fix inside RDoc too).
- It will likely be A LOT faster as you can avoid a lot of unnecessary abstractions rdoc has

lib/rbs/annotate/formatter.rb
ksss: Thanks!

ksss:
@st0012 Thank you for reviewing.
I think using Prism is a great idea.  
To reduce the dependency on RDoc, I believe we need to solve the following two issues:

- Reading comments from C source files 
- Conversion to Markdown for LSP

st0012:
Does rbs use rdoc to parse C source code? Or just uses it to retrieve the generated ri data?
AFAICT, Ruby's C source usually isn't stored locally after installation (related irb issue: https://github.com/ruby/irb/issues/664).

> Conversion to Markdown for LSP

I see that makes sense.

Interestingly, since ruby-lsp uses rbs to provide Ruby core's documentation, the dependency chain seems to be: ruby-lsp -> rbs -> rdoc.

ksss:
I didn’t fully understand before.  
After looking into it, it seems that RBS uses RDoc mainly in two ways, with the following process:

# Writing RDoc documentation into RBS files

1. Generate ri using the RDoc CLI
    - https://github.com/ruby/rbs/blob/468b036deaa8f87c5137d0e80d22a3e69463ac4c/bin/generate_docs.sh#L30
1. Read the ri and convert comments into Markdown
    - https://github.com/ruby/rbs/blob/468b036deaa8f87c5137d0e80d22a3e69463ac4c/lib/rbs/annotate/formatter.rb#L86
1. Write the Markdown comments into RBS files
    - https://github.com/ruby/rbs/blob/468b036deaa8f87c5137d0e80d22a3e69463ac4c/lib/rbs/annotate/rdoc_annotator.rb#L24

# Reading RBS file comments as RDoc documentation

1. Use an RDoc plugin
    - https://github.com/ruby/rbs/blob/468b036deaa8f87c5137d0e80d22a3e69463ac4c/lib/rdoc/discover.rb#L11
1. Parse the RBS file
    - https://github.com/ruby/rbs/blob/468b036deaa8f87c5137d0e80d22a3e69463ac4c/lib/rdoc_plugin/parser.rb#L17
1. Read comments in the RBS file as RDoc documentation
    - https://github.com/ruby/rbs/blob/master/lib/rdoc_plugin/parser.rb

We can try it with:  
`$ bundle exec rdoc core/array.rbs`

st0012:
@ksss Thank you for the summary, this is very helpful 🙏 

I wonder if we should move the plugin into `rdoc` and become its own parser. It'll make `rdoc` and `rbs` less coupled as `rdoc` can simply disable that parser when `rbs` gem is not installed. This should reduce the chance of `rdoc` breaking `rbs`. WDYT?

ksss:
@st0012 I think it is a very excellent idea!

@soutaro How about it?

soutaro:
@st0012 Hi Stan!

Moving the plugin to `rdoc` gem would make sense. (But it may result breaking `rdoc`, when RBS is updated...)

We have `rbs annotate` command that also uses RDoc, and I don't think this feature cannot be easily moved to `rdoc`. It means `rbs` gem will continue depending on `rdoc` internal (and break again easily.)

So, we should focus if it would benefit the end users, and because the rdoc plugin has a little (maybe zero?) users, the conclusion doesn't looks obvious to me.

soutaro (approved):
Merge this for now.
We can consider if we can move some parts to RDoc later.

st0012:
Yeah let's get it merged first and perhaps talk a bit more about future integration during RubyKaigi 😄

PR #2357, opened 2025-03-25
Add signature for top-level using

KitaitiMakoto:
Hi,

I added signature for top-level `using`.
But I couldn't see how to test it. Could you advise me?

Thank you.


ksss:
- The `Kernel` does not have a `using` method, which should not be defined in the `Kernel`.
- RBS does not seem to have a way to represent the `main` object
- It is currently difficult to represent Refinement in RBS.

ksss:
@KitaitiMakoto 
I realized that we should support the main object. Thank you for the request.
https://github.com/ruby/rbs/issues/2361

PR #2359, opened 2025-03-25
Fix IO#reopen, add open3 missing sigs

HoneyryderChuck:
`IO#reopen` should be able to receive the same kwargs that `IO#open` receives; also, the mode can be an integer too.

Have a separate commit adding missing sigs for `open3`. @soutaro this is currently a WIP, as the sigs can't be expressed fully in RBS:

* most functions accept an optional options hash at the end, despite commands (string) already supporting optional (string) args as well. it's like a second set of optional args, and this can't be expressed in RBS.
* some functions (i.e. `pipeline_w` return an array where the first element is stdin/stdout (IO) and the rest is one or more wait threads. "variable size" tuples can't be expressed in RBS yet as well.

### DIFF
--- a/stdlib/open3/0/open3.rbs
+++ b/stdlib/open3/0/open3.rbs
@@ -50,6 +50,103 @@
 module Open3
   # <!--
   #   rdoc-file=lib/open3.rb
+  #   - Open3.capture2([env, ] command_line, options = {}) -> [stdout_s, status]
+  #   - Open3.capture2([env, ] exe_path, *args, options = {}) -> [stdout_s, status]
+  # -->
+  # Basically a wrapper for Open3.popen3 that:
+  #
+  # *   Creates a child process, by calling Open3.popen3 with the given arguments
+  #     (except for certain entries in hash `options`; see below).
+  # *   Returns as string `stdout_s` the standard output of the child process.
+  # *   Returns as `status` a `Process::Status` object that represents the exit
+  #     status of the child process.
+  #
+  # Returns the array `[stdout_s, status]`:
+  #
+  #     stdout_s, status = Open3.capture2('echo "Foo"')
+  #     # => ["Foo\n", #<Process::Status: pid 2326047 exit 0>]
+  #
+  # Like Process.spawn, this method has potential security vulnerabilities if
+  # called with untrusted input; see [Command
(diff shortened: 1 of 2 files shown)


stdlib/open3/0/open3.rbs
```diff
@@ -48,6 +48,103 @@
...
+  #
+  # If one or more `args` is given, each is an argument or option to be passed to
+  # the executable:
+  #
+  #     Open3.capture2('echo', 'C #')
+  #     # => ["C #\n", #<Process::Status: pid 2326267 exit 0>]
+  #     Open3.capture2('echo', 'hello', 'world')
+  #     # => ["hello world\n", #<Process::Status: pid 2326299 exit 0>]
+  #
+  def self.capture2: (?Hash[String, String] env, *String cmds, ?Hash[untyped, untyped] options) -> [IO, Process::Status]
```
ksss, line 146: This is a syntax error.
It seems to me that it would be better expressed in overload.

ksss (commented):
I think it would be better with test code.

stdlib/open3/0/open3.rbs
HoneyryderChuck: yes, hence why I opened this PR in a broken state. I believe that this is a valid and common signature across the stdlib (i.e. options hash optionally at the end) that RBS should be able to acknowledge. I don't think that overload would express it better.

ParadoxV5: +1 for [`IO#popen`](https://rubyapi.org/3.4/o/io#method-c-popen)

stdlib/open3/0/open3.rbs
```diff
@@ -143,5 +240,827 @@ module Open3
   #     Open3.capture2e('echo', 'hello', 'world')
   #     # => ["hello world\n", #<Process::Status: pid 2371894 exit 0>]
   #
-  def self.capture2e: (*String, ?stdin_data: String, ?binmode: boolish) -> [String, Process::Status]
+  def self.capture2e: (?Hash[String, String] env, *String cmds, ?Hash[untyped, untyped] options) -> [IO, Process::Status]
```
apiology, line 243: The capture* methods return strings, not streams

PR #2378, opened 2025-04-01
Make IO.binread, IO.binwrite, IO.read and IO.write accept _ToPath

KitaitiMakoto:
Hi,

I found `IO.binread`, `IO.binwrite`, `IO.read` and `IO.write`'s signatures don't allow `Pathname` for the first argument and fixed it.

Thanks.

### DIFF
--- a/core/io.rbs
+++ b/core/io.rbs
@@ -2278,7 +2278,7 @@ class IO < Object
   # potential security vulnerabilities if called with untrusted input; see
   # [Command Injection](rdoc-ref:command_injection.rdoc).
   #
-  def self.binread: (_ToPath | string name, ?Integer? length, ?Integer offset) -> String
+  def self.binread: (path name, ?Integer? length, ?Integer offset) -> String
 
   # <!--
   #   rdoc-file=io.c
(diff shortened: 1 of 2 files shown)


core/io.rbs
```diff
@@ -2278,7 +2278,7 @@ class IO < Object
   # potential security vulnerabilities if called with untrusted input; see
   # [Command Injection](rdoc-ref:command_injection.rdoc).
   #
-  def self.binread: (String name, ?Integer? length, ?Integer offset) -> String
+  def self.binread: (_ToPath | string name, ?Integer? length, ?Integer offset) -> String
```
ksss, line 2281: We can use  builtin type alias `path` instead of `_ToPath | string`.
https://github.com/ruby/rbs/blob/b51d7ded7d901051cbb6f86fa78974ddd15c00d5/core/builtin.rbs#L255

KitaitiMakoto: Oh, nice. Thanks.

KitaitiMakoto:
@ksss 
Thank you for the advice. I fixed.

PR #2394, opened 2025-04-08
Validate that all core RBS files parse without error

apiology:
I'm seeing a parse failure in module.rbs while using the code on the master branch - this regression test reproduces it


apiology:
I may have been confused - will reopen if needed.  Sorry about that!

PR #2419, opened 2025-04-18
Remove unused code path from trailing_block methods

st0012:
The `trailing_block!` method is only called in one place:

```
end_loc = node.rparen_loc || node.parameters&.location || node.name_loc
```

And all three possible values for `end_loc` are of type `Prism::Location`, so the `trailing_block!` method's implementation and signature can be simplified to only handle `Prism::Location`.


st0012:
Turns out @soutaro plans to use it for nodes later! So closing now 😄

PR #2398, opened 2025-04-10
Introduce standalone C parser for RBS with arena allocation

st0012:
# RBS C Parser Library Refactoring

This PR refactors the RBS parser and related components into a standalone C library that no longer depends on the Ruby runtime. This architectural change enables direct integration with static analysis tools like Sorbet while potentially improving performance.

[Sorbet's RBS support](https://sorbet.org/docs/rbs-support) already runs on this new architecture and we haven't discovered any major issues around it.

This work was a collaborative effort by

- @amomchilov
- @Morriar
- @st0012 

## Key Improvements

- **Ruby-Independent Implementation**: Extracted parser from `ext` folder into a standalone C library with a clean API, which can now be embedded in non-Ruby tools without Ruby runtime dependency (e.g. Sorbet, JRuby)
- **Enhanced Memory Management**: Implemented arena allocator to efficiently manage parser object lifecycles
- **Improved Architecture**: Clear separation between public API (headers) and implementation
- **Performance**: Potential performance gains from custom memory management and reduced overhead

### Enhanced Memory Management

Arena allocator handles all memory for parser objects, including parser itself, lexer, constant pool, strings...etc. When the parser is freed by calling `rbs_parser_free`, the allocator will free all the objects it allocated. This eliminates the need to manually free individual objects and reduces the risk of memory leaks.

## Component Architecture

```mermaid
graph TD
    RubyClient[Ruby Client] --> RubyAPI[Ruby API]
    CClient[C Client] --> CAPI[C API]

    RubyAPI --> CExtension[C Extension]
    CExtension --> CLibrary
    CAPI --> CLibrary

    subgraph CLibrary[C Library]
        subgraph Parser1[Parser Instance 1]
            direction TB
            ConstantPool1[Constant Pool]
            Lexer1[Lexer]
            ArenaAllocator1[Arena Allocator]
        end

        subgraph Parser2[Parser Instance 2]
            direction TB
            ConstantPool2[Constant Pool]
            Lexer2[Lexer]
            ArenaAllocator2[Arena Allocator]
        end
    end

    subgraph "Public API"
        RubyAPI
        CAPI
    end

    %% Parser1 --> ConstantPool1
    %% Parser1 --> Lexer1
    %% Parser1 --> ArenaAllocator1
    %% Parser2 --> ConstantPool2
    %% Parser2 --> Lexer2
    %% Parser2 --> ArenaAllocator2
```

PR #2437, opened 2025-04-30
Setup clang format for C code

st0012:
While working on #2398, we found it necessary to have some sort of auto-formatting for C code as well to ensure consistent coding style. But instead of adding it as part of #2398, which already introduces many changes on the implementation itself, I think it'd be better to have the auto-formatting setup in a separate PR.

This PR:
- Sets up `clang-format` for the project, including
    - `.clang-format` defines the rules of formatting (like `rubocop.yml`)
    - A new `format:c` rake task will perform the formatting to all C code under `ext/`, `src/`, and `include/`
    - A new `format:c_check` rake task will check if formatting is required without actually making the changes
    - A new CI workflow to perform `format:c_check`
    - Introductions about C formatting in the readme
- Sets up auto-formatting in VS Code, including:
    - Extension recommendation for the formatting
    - Workspace settings to enable it
- Formats the code with the given rules. The most noticeable changes are indentation related. Currently we have both 2-space and 4-space based indentation. Since the majority of the code uses 2-space indentation, which is what #2398 mostly follows, I set the rule to 2 and update the code that current indents with 4 spaces


ext/rbs_extension/parser.c
```diff
@@ -89,34 +87,32 @@ static bool rbs_is_untyped_params(method_params *params) {
...
 /**
  * Raises RuntimeError with "Unexpected error " message.
  * */
 static NORETURN(void) rbs_abort(void) {
   rb_raise(
-    rb_eRuntimeError,
-    "Unexpected error"
-  );
+      rb_eRuntimeError,
+      "Unexpected error");
```
amomchilov, line 111: Could we set `BracketAlignmentStyle: BAS_BlockIndent`

```suggestion
      rb_eRuntimeError,
      "Unexpected error"
  );
```

src/util/rbs_constant_pool.c
amomchilov, line 1: Is it possible to make a single-file exception for this file to be indented with 4 spacs?

st0012: I don't think it's possible to do dynamic value assignment in `.clang-format`. We may be able to hack it in the rake task but I don't think it's worth it.
Since we're gong to edit the file synced from Prism anyway, why don't we just format it at the same time?

ext/rbs_extension/parser.c
st0012: I think it's `AlignAfterOpenBracket: BlockIndent` instead. Updated

.clang-format
```diff
@@ -0,0 +1,75 @@
+---
+Language: Cpp
+
+# Indentation
+UseTab: Never
+IndentWidth: 2
```
Morriar, line 6: Shouldn't we stick to 4? I believe this is what's being used in cruby and generally the expected value for C code?

st0012: Most of the current code uses 2. But if @soutaro is happy to update them all at once I'll be happy to update it too.

src/util/rbs_constant_pool.c
st0012: Actually, hacking the rake task isn't a great idea either as there would then be inconsistency between VS Code's auto-formatting and the task.

.clang-format
soutaro: Let's move to 4 to align the codebase with cruby. 👍

soutaro:
I'm good for clang-format. Can you add a commit for 4-space indentation?

st0012:
Updated the indent width to 4 👍

soutaro (approved):
Thanks!

soutaro:
Updated the branch protection setting to require `format-check` job.

.github/workflows/c-check.yml
```diff
@@ -0,0 +1,53 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.4"
+          bundler-cache: none
+      - name: Set working directory as safe
+        run: git config --global --add safe.directory $(pwd)
+      - name: Set up permission
```
soutaro, line 19:
```suggestion
```

soutaro, line 19: This step causes an error...
(Other jobs are fixed in https://github.com/ruby/rbs/pull/2485.)

soutaro:
Merged this PR to continue working on other PRs.

PR #2439, opened 2025-05-01
Use `erb` instead of `set` for load path testing

st0012:
The recent CI failure seems to be related to https://github.com/ruby/ruby/pull/13074. So changing the testing target to `erb` should resolve it.


soutaro (approved):
Thanks! 💯

PR #2445, opened 2025-05-02
Fix error at Ruby CI

soutaro:
This PR is to fix build errors which is reported in https://github.com/ruby/ruby/pull/13237.

* This PR adds CI job to compile without C23 extensions.
* You can compile in C99 mode in your local environment with `compile_c99` Rake task

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -106,3 +106,31 @@ jobs:
       - run: bundle exec rake test:valgrind
         env:
           RUBY_FREE_AT_EXIT: 1
+  C99_compile:
+    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: ['3.4', head]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          brew install ruby-build
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler: none
+      - name: Set working directory as safe
+        run: git config --global --add safe.directory $(pwd)
+      - name: Update rubygems & bundler
+        run: |
+          ruby -v
+          gem update --system
+      - name: clang version
+        run: clang --version
+      - name: bin/setup
+        run: |
+          bin/setup
+      - run: bundle exec rake clean compile_c99
+
--- a/ext/rbs_extension/compat.h
+++ b/ext/rbs_extension/compat.h
@@ -1,0 +1,10 @@
+#ifdef __clang__
+#define SUPPRESS_RUBY_HEADER_DIAGNOSTICS_BEGIN \
+  _Pragma("clang diagnostic push")        \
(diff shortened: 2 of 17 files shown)


src/lexstate.c
```diff
@@ -92,7 +92,7 @@ static const char *RBS_TOKENTYPE_NAMES[] = {
   "tANNOTATION",      /* Annotation */
 };
 
-rbs_token_t NullToken = { .type = NullType, .range = {} };
+rbs_token_t NullToken = { .type = NullType, .range = {0} };
```
amomchilov, line 95: You'll need to make the same change here:

https://github.com/ruby/rbs/blob/6c08f20a3dfcba8efebd54da1010738c3c0a3ed1/src/parser.c#L3418

```diff
 .constant_pool = {0}, 
```

soutaro:
I've been trying a CI task to compile the library without C23 extensions, but not finished yet...

.github/workflows/ruby.yml
```diff
@@ -106,3 +106,31 @@ jobs:
       - run: bundle exec rake test:valgrind
         env:
           RUBY_FREE_AT_EXIT: 1
+  C99_compile:
```
soutaro, line 109: This job compiles the extension without C23 extensions to ensure it compiles with C99 compilers.

To compile it with `clang`, not `gcc`, the macos platform is used.

ext/rbs_extension/compat.h
```diff
@@ -0,0 +1,10 @@
+#ifdef __clang__
```
soutaro, line 1: Including `ruby.h` caused compilation errors without C23 extensions. This macro is to suppress diagnostics in the header files.

templates/ext/rbs_extension/ast_translation.c.erb
```diff
@@ -183,4 +183,6 @@ VALUE rbs_struct_to_ruby_value(rbs_translation_context_t ctx, rbs_node_t *instan
             return ID2SYM(rb_intern3((const char *) constant->start, constant->length, ctx.encoding));
         }
     }
+
+    rb_raise(rb_eRuntimeError, "Unknown node type: %d", instance->type);
```
soutaro, line 187: GCC reports `warning: control reaches end of non-void function [-Wreturn-type]` diagnostic here.

ext/rbs_extension/extconf.rb
```diff
@@ -13,5 +13,9 @@
 
 append_cflags ['-std=gnu99', '-Wimplicit-fallthrough', '-Wunused-result']
 append_cflags ['-O0', '-g'] if ENV['DEBUG']
+if ENV["TEST_NO_C23"]
```
soutaro, line 16: Using `append_cflags` doesn't work because `ruby.h` issues compilation error without C23 extensions here.

src/lexstate.c
soutaro: Reusing `NULL_RANGE` didn't work with clang-15, on GitHub Actions... 😢

PR #2453, opened 2025-05-05
Replace reference with official documentation

p8:
ruby-doc.org isn't official documentation.
This also adds the `Thread` namespace and links to the latest Ruby version.


soutaro:
Thanks. But I found the method is removed in Ruby 3.0.

p8:
Thanks!

PR #2456, opened 2025-05-06
Expose and fix `Block#location`

Morriar:
This pull-request expose the `location` field for the `Block` class.

It's useful when we want to point an error to the block inside the signature.

The location was already computed and associated with the node but it wasn't exposed to Ruby, so hard to test.

I also fixed the location that was pointing at the wrong range.


soutaro (approved):
Thanks! 🎉

PR #2457, opened 2025-05-06
Expose a method to parse type parameters

Morriar:
With Sorbet we use RBS signatures directly as comments:

```rb
#: (Integer, Integer) -> Integer
def add(a, b)
  a + b
end
```

We parse these comments using `rbs_parse_method_type`.

When it comes to generic class we want to offer the same syntax so the user can simply specify the type parameters:

```rb
#: [in A, B < Object]
class Foo
end
```

We need to expose a parsing method for those.

Note that the Ruby method is not required per se, it just makes it easier to test.


soutaro (approved):
👍

PR #2463, opened 2025-05-08
Enable `-Wc++-compat`

amomchilov:
Adds a flag which makes the compiler warn if we use anything that's invalid C++. This is handy because we compile RBS as part of Sorbet. There's no warnings today, but it'll catch them going forward.

Docs: [Clang](https://clang.llvm.org/docs/DiagnosticsReference.html#wc-compat), [GCC](https://gcc.gnu.org/onlinedocs/gcc-4.4.0/gcc/Warning-Options.html#:~:text=%2DWc%2B%2B%2Dcompat%20(C%20and%20Objective%2DC%20only))


raiderss:
Amazing work! As someone from Eyes Store developing FiveM resources, I find this really inspiring.

soutaro (approved):
👍

PR #2464, opened 2025-05-08
fixing sig for addrinfo

HoneyryderChuck:
can receive family param as an integer


raiderss:
Excellent contribution to open source! At Eyes Store, we believe in sharing knowledge for the FiveM community too.

PR #2482, opened 2025-05-13
Skip tests that are obsolete after CGI removal

st0012:
This should fix the build.

Related to https://github.com/ruby/ruby/pull/13275


soutaro:
Merging this PR manually, without the queue.
The `format-check` is introduced at #2437, but it is blocked by the `CGI` removal.

PR #2498, opened 2025-05-19
Only load extconf_compile_commands_json when compiling through Rake 

st0012:
Otherwise, it'd cause LoadError in official release because `extconf_compile_commands_json` is not a runtime dependency.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -11,6 +11,14 @@ bin = File.join(__dir__, "bin")
 
 Rake::ExtensionTask.new("rbs_extension")
 
+compile_task = Rake::Task[:compile]
+
+task :setup_extconf_compile_commands_json do
+  ENV["COMPILE_COMMANDS_JSON"] = "1"
+end
+
+compile_task.prerequisites.unshift(:setup_extconf_compile_commands_json)
+
 test_config = lambda do |t|
   t.libs << "test"
   t.libs << "lib"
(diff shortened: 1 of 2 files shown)


soutaro:
@st0012 Does it fix the error? It looks like `rake compile` always requires `extconf_compile_commands_json`. How about simply rescuing `LoadError` in `extconf.rb`?

st0012:
rubygems actually doesn't use `rake` to build gems with extension. It has a specific [builder](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/ext/ext_conf_builder.rb#L9) to handle `extconf.rb`, which just calls `ruby extconf.rb` with the right flags, which can be verified with:

```
From: /opt/rubies/3.4.3/lib/ruby/3.4.0/rubygems/ext/ext_conf_builder.rb @ line 30 :

    25:     begin
    26:       cmd = ruby << File.basename(extension)
    27:       cmd << "--target-rbconfig=#{target_rbconfig.path}" if target_rbconfig.path
    28:       cmd.push(*args)
    29: 
 => 30:       binding.irb
    31:       run(cmd, results, class_name, extension_dir) do |s, r|
    32:         mkmf_log = File.join(extension_dir, "mkmf.log")
    33:         if File.exist? mkmf_log
    34:           unless s.success?
    35:             r << "To see why this extension failed to compile, please check" \

irb(Gem::Ext::ExtConfBuilder):001> cmd
=> ["/opt/rubies/3.4.3/bin/ruby", "extconf.rb"]
```

So during normal gem installation, no rake tasks are executed, and the env var won't be set.

Issue #1521, opened 2023-09-16
Cannot specify type variables for `instance` 


ParadoxV5:
```rbs
class MyWrapper[E]
  def self.of: (E) -> instance # RBS::UnknownTypeName (it’s not inferred from `MyWrapper[E]`)
  def self.of: (E) -> instance[E] # RBS::SyntaxError
  def self.of: [E] (E) -> instance # ok, but `[E]` is not used in practice (it’s not the same `[E]` as the one in `MyWrapper[E]`)
  def self.of: (untyped) -> instance # ok, but `untyped` is bad
end
```

Even Steep is aware of this limitation:
https://github.com/soutaro/steep/blob/v1.5.3/lib/steep/subtyping/check.rb#L307

Proposal: `instance[E]`?

The problem might be significant for `class` as well:
https://github.com/soutaro/steep/blob/v1.5.3/lib/steep/subtyping/check.rb#L322

soutaro:
I know there are some cases that requires that kind of types, but we don't have a plan to support it at least for now.

A class inheriting from a generic class may be non-generic, and what can we do for `instance[T]` (and maybe `self[T]`) types?

```rbs
class Generic[T]
  def foo: () -> instance[Array[T]]
end

class NonGeneric < Generic[String]
end

# Even worse if it inherited by another generic class.
class AnotherGeneric[T] < NonGeneric
end
```

ParadoxV5:
> `def self.of: (E) -> instance # RBS::UnknownTypeName (it’s not inferred from `MyWrapper[E]`)`

How about letting class methods pick up generics?

ParadoxV5:
> > `` def self.of: (E) -> instance # RBS::UnknownTypeName (it’s not inferred from `MyWrapper[E]`) ``
> 
> How about letting class methods pick up generics?

Would this rather be a Steep job? Unless the RBS gem itself also involved…

soutaro:
The type you want would be this?

```
def self.of: [E, T < MyWrapper[E]] (E) -> T
```

Steep allows explicit type application, and it will help:

```
MyWrapper.of(123) #$ Integer, AnotherWrapper         # E=Integer, T=AnotherWrapper
```

ParadoxV5:
> The type you want would be this?
> 
> ```
> def self.of: [E, T < MyWrapper[E]] (E) -> T
> ```

[...]

PR #2502, opened 2025-05-20
Add type arguments support to singleton types

allcre:


ParadoxV5:
@soutaro Did `singleton(Array)[String]` (or `singleton(Array[String])`, whatever) make sense in the first place?

Morriar:
@soutaro we use it in Sorbet to represent the type of the attached class to a singleton.

It's useful around factories, here's a simple example:

```rb
class Box
  extend T::Sig
  extend T::Generic
  
  E = type_member

  sig { params(e: E).void }
  def initialize(e)
    @e = e
  end

  sig { returns(E) }
  def e
    @e
  end
end

extend T::Sig

sig { returns(T.class_of(Box)[Box[Integer]]) }
def example
  Box
end

x = example
T.reveal_type(x) # => T.class_of(Box)[Box[Integer]]
x.new("str") # error: Expected `Integer` but found `String` for argument `e`
```

More involved examples can be found in the documentation: https://sorbet.org/docs/class-of#tclass_of-applying-type-arguments-to-a-singleton-class-type.

ParadoxV5:
RBS expects us to genericize class methods because it only recognizes type parameters at the instance level.
```rbs
class Set[E]
  def self.[]: [E] (*E elements) -> Set[E] # `E` is totally not duplicated
end
```

Honestly, that might be a flawed design:
```rbs
class Set[E]
  def self.[]: (*untyped elements) -> instance # `instance` is `Set[untyped]`!
end
```

soutaro:
Thanks @Morriar,

I got the use case and agree that it cannot be written in RBS now. Using an interface would be a workaround, but not sure if it can cover the existing use cases.

Let me confirm the semantics: the type `singleton(T)[S, ...]` means that *a class object of T but the instance created through the value of the type is T[S, ...]*. Looks like it makes some sense, except if it needs a big overhaul in RBS (and Steep).

What should we do for the other singleton methods?

```rbs
class Box[T]
  def initialize: (T) -> void

  def self.new_array: [T] (T) -> Box[Array[T]]
end

b = Box #: singleton(Box)[String]
b.new("")           # => Box[String]
b.new(1)            # => type error
b.new_array(1)      #=> ???
```

Looks like it only works for `.new` method in Sorbet?

ParadoxV5:
> What should we do for the other singleton methods?

How about changing the semantics so that type parameters apply on the class level as well?
It would resolve my #1521.
```rbs
class Box[T]
  def self.new_array: (T) -> Box[Array[T]] # No need to genericize the method separately
  def self.[]: (T) -> instance # `instance` was `Box[untyped]`, but now `Box[T]`
end
```

soutaro:
@ParadoxV5 Yeah, it would make sense. But, how can we associate the type parameter `T` in `self.[]` and the type parameter given to the singleton class. `singleton(Box)[T]`? Can we do that by adding another type `instance[T]`?

ParadoxV5:
`singleton(Box)[T]` should probably be `singleton(Box[T])` rather, so that `singleton(Box[Integer])#[]` → `Box[Integer].[]` → `T` in `self.[]` is `Integer`.

Allowing `instance[T]` is an option to, as I requested in #1521 long ago.
Its flexibility would enable `def self.new_array: (T) -> instance[Array[T]]`.

Morriar:
> Let me confirm the semantics: the type singleton(T)[S, ...] means that a class object of T but the instance created through the value of the type is T[S, ...].

Yes, when limited to the type of the attached class (`instance` in RBS?). Here's [an example on sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Wrapper%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20Elem%20%3D%20type_member%0A%0A%20%20class%20%3C%3C%20self%0A%20%20%20%20extend%20T%3A%3ASig%0A%0A%20%20%20%20sig%20do%0A%20%20%20%20%20%20type_parameters%28%3AE%29%0A%20%20%20%20%20%20%20%20.params%28e%3A%20T%3A%3AClass%5BT.type_parameter%28%3AE%29%5D%29%0A%20%20%20%20%20%20%20%20.returns%28T.all%28T.self_type%2C%20T.class_of%28Wrapper%29%5BWrapper%5BT.type_parameter%28%3AE%29%5D%5D%29%29%0A%20%20%20%20end%0A%20%20%20%20def%20of%28e%29%0A%20%20%20%20%20%20self%0A%20%20%20%20end%0A%20%20end%0A%0A%20%20sig%20%7B%20returns%28Elem%29%20%7D%0A%20%20def%20elem%20%3D%20T.unsafe%28nil%29%0Aend%0A%0Aclass%20MyWrapper%20%3C%20Wrapper%0A%20%20Elem%20%3D%20type_member%0A%0A%20%20def%20self.bar%3B%20end%0Aend%0A%0Ares%20%3D%20Wrapper.of%28Integer%29%0AT.reveal_type%28res%29%20%23%20T.class_of%28Wrapper%29%5BWrapper%5BInteger%5D%5D%0Ares.new.elem%20%23%20ok%0A%0Ares%20%3D%20MyWrapper.of%28Integer%29%0AT.reveal_type%28res%29%20%23%20T.class_of%28MyWrapper%29%5BMyWrapper%5BInteger%5D%5D%0Ares.new.elem%20%23%20ok%0Ares.bar%20%23%20ok).

But it can also be used to specify the type of the generic parameters to the singleton class:

```rb
class BoxFactory
  class << self
    extend T::Sig
    extend T::Generic

    Kind = type_member

    sig { returns(Box[Kind]) }
    def create
      Box[Kind].new
    end
  end
end

class Box
  extend T::Generic

  Kind = type_member
end

extend T::Sig

sig do
  type_parameters(:T)
    .params(factory: T.class_of(BoxFactory)[BoxFactory, T.type_parameter(:T)])
    .returns(Box[T.type_parameter(:T)])
end
def create_box(factory)
  factory.create
end
```

Here's the equivalent RBS inline syntax we want with Sorbet:

```rb
class BoxFactory
  #: [E]
  class << self
    #: -> Box[E]
    def create
      Box.new #: Box[E]
    end
  end
end

#: [E]
class Box
end

#: [T] (singleton(BoxFactory)[BoxFactory, T]) -> Box[T]
def create_box(factory)
  factory.create
end
```

soutaro:
Finally adding a support for RBS version of `T.class_of(T)(T[X])`. This type is essential for RBS support in Sorbet.

`singleton(T)[T[S]]` is a type something like `singleton(T) & T::_Factory[T[S]]` where `T::_Factory[X]` is:

```rbs
class T
  interface _Factory[Return]
    def new: (arguments_of_T_new) -> Return
  end
end
```

Note that we cannot have a generic `_Factory` interface which can be used for any classes/modules because RBS doesn't have a mechanics to copy the type of arguments. (We may want to use `(?) -> Return` type, but it's more permissive than what we expect for the types.)

The type itself might not be immediately very useful for RBS and Steep.

PR #2542, opened 2025-06-06
[prototype runtime] Find redefined methods

ksss:
Since methods redefined in subclasses may have different types, I believe we should display the prototypes of such redefined methods when using `--todo`.

# Repro

```rb
# sample.rb
class A
  def foo; 1; end
end
class B < A
  def foo; 'a'; end
end
```

```rbs
# sample.rbs
class A
  def foo: () -> Integer
end
class B < A
end
```

```
$ bundle exec rbs -I sample.rbs prototype runtime --require-relative sample.rb --todo B
```

### Actual

```rbs
class B < ::A
end
```

### Expect

```rbs
class B < ::A
  def foo: () -> untyped
end
```

### DIFF
--- a/lib/rbs/prototype/runtime.rb
+++ b/lib/rbs/prototype/runtime.rb
@@ -27,6 +27,10 @@ module RBS
           method_definition = @builder.build_singleton(module_name.absolute!).methods[method.name]
           return false unless method_definition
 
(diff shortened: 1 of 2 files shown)


soutaro (approved):
👍

PR #2543, opened 2025-06-08
Remove undocumented `#nonzero?`

ksss:
`{Integer,Float,Complex,Rational}#nonzero?` methods are not defined as a method in ruby and is not useful at all because it also inhibits the display of documents in steep.
It should be removed.


soutaro (approved):
👍

PR #2562, opened 2025-06-22
Add documentation for `Range#{minmax,count,to_a,entries}`

ksss:
I added several methods that were documented in RDoc but did not exist in the RBS.
With this change, it becomes easier to reference method documentation in tools like LSP.
Additionally, `Range#count` takes into account the case where it returns infinity.


soutaro (approved):
👍

PR #2601, opened 2025-07-15
Add explicit dependency on the tsort gem

ima1zumi:
This is being bundled, starting in the next Ruby it will warn: https://github.com/ruby/ruby/commit/65a0f46880ecb13994d3011b7a95ecbc5c61c5a0


soutaro (approved):
🙏

PR #2602, opened 2025-07-17
Add bundle update schedule

soutaro:


.github/workflows/bundle-update.yml
```diff
@@ -0,0 +1,59 @@
+name: Scheduled Bundle Update PR
+
+on:
+  schedule:
+    - cron: '0 0 * * 2'
```
ParadoxV5, line 5: [Dependabot supports weekly scheduling and even arbitrary `cron` too](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#schedule-), if you’re looking to update the Lockfile less frequently.

.github/workflows/bundle-update.yml
```diff
@@ -0,0 +1,59 @@
...
+        if: env.has_update == 'true'
+        run: git push origin HEAD
+
+      - name: Create Pull Request
+        if: env.has_update == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh pr create \
+            --title "bundle update `/Gemfile`" \
```
ParadoxV5, line 56:
```suggestion
            --title "bundle update `/Gemfile.lock`" \
```

.github/workflows/bundle-update.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.4'
+
+      - name: Set up git
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
```
ParadoxV5, line 24: I’d suggest setting up Git after “Run bundle update”, but it would mean more `if: env.has_update`s.

.github/workflows/bundle-update.yml
```diff
@@ -0,0 +1,59 @@
...
+          if git diff --cached --quiet; then
+            echo "No changes to commit, exiting."
+            exit 0
+          fi
+
+          git commit -m "bundle update"
+          echo "has_update=true" >> $GITHUB_ENV
+
+      - name: Push branch
+        if: env.has_update == 'true'
```
ParadoxV5, line 47: Speaking of which… why `== 'true'`?

.github/workflows/bundle-update.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Create Pull Request
+        if: env.has_update == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh pr create \
+            --title "bundle update `/Gemfile`" \
+            --body "Automated bundle update" \
+            --head "$(git rev-parse --abbrev-ref HEAD)" \
```
ParadoxV5, line 58: Using the branch name is more convenient.

.github/workflows/bundle-update.yml
```diff
@@ -0,0 +1,59 @@
...
+          fi
+
+          git commit -m "bundle update"
+          echo "has_update=true" >> $GITHUB_ENV
+
+      - name: Push branch
+        if: env.has_update == 'true'
+        run: git push origin HEAD
+
+      - name: Create Pull Request
```
ParadoxV5, line 50: Nit: Capitalization differs between the step titles

ParadoxV5 (commented):
I don’t trust AI as long as they parse in tokens rather than letters.

PR #2612, opened 2025-07-17
Allows the use of a path list as RBS_SKIP_TESTS

nobu:
In ruby/ruby test-bundled-gems, rbs has many failures on Windows. Currently we ignore all failures in this test totally, but that is the test-run makes no sense but just wastes times.
So I want to also specify another file for the tests to be skipped only on Windows, in addition to the platform independent skipping list file, that we are using already.


soutaro:
The `master` branch is unstable, so I don’t think it makes sense to send it to the Ruby repo.

I think opening a PR for `aaa-3.9.x` branch and testing it on Ruby CI with the commit would be fine.

PR #2626, opened 2025-07-26
Module-self allow void once

ksss:
This is a regression case for `module Foo : Array[void] end`.
I added the test cases I had on hand.

This is a follow-up to https://github.com/ruby/rbs/pull/2590


soutaro (approved):
💯

PR #2635, opened 2025-08-07
Inline constant declaration

soutaro:


lib/rbs/buffer.rb
```diff
@@ -89,9 +89,9 @@ def inspect
 
     def rbs_location(location, loc2=nil)
       if loc2
-        Location.new(self, location.start_character_offset, loc2.end_character_offset)
+        Location.new(self.top_buffer, location.start_character_offset, loc2.end_character_offset)
```
soutaro, line 92: Let the location associated to the `#top_buffer` to make it compatible with Prism locations.

PR #2644, opened 2025-08-25
Introduce `type_fingerprint`

soutaro:
This PR introduces `#type_fingerprint` method that compute a *fingerprint* of inline RBS type declarations.

* The *fingerprint* is changed if the inline Ruby code changes it's RBS type declarations -- adding classes, renaming methods, changing method types
* The *fingerprint* stays same if the inline Ruby code is updated only for it's implementations -- changing the implementation of `def` node or adding new lines/white spaces to the source code

Using the method, type checkers can detect if the codebase needs re-type-checking after inline Ruby code edits.

### DIFF
--- a/docs/type_fingerprint.md
+++ b/docs/type_fingerprint.md
@@ -1,0 +1,21 @@
+# Type Fingerprint of RBS Inline AST
+
+Type fingerprint of RBS Inline AST is an object that can be used to detect if the RBS Inline AST is updated and the type checker should type check the whole codebase again.
+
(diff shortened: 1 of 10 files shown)


ParadoxV5:
🧠.

Using full nesting arrays over hashes/checksums… intersting.

soutaro:
The string digest can be computed with the result of `to_json`. I want a strict change detection, but the digest on the `to_json` result would work well usually (with lower memory footprint).

Issue #2563, opened 2025-06-23
Degradation of parsing performance with v4.0.0.dev.4


ksss:
# Prepare

The target file is https://github.com/ruby/gem_rbs_collection/blob/main/gems/activerecord/8.0/activerecord-generated.rbs

```rb
$ cat parse.rb
require 'rbs'
require 'benchmark/ips'

sig = Pathname('../gem_rbs_collection/gems/activerecord/8.0/activerecord-generated.rbs').read
Benchmark.ips do |x|
  x.report("rbs v#{RBS::VERSION} parse activerecord-generated.rbs") do
    RBS::Parser.parse_signature(sig)
  end
end
```

# Runner

```
$ bundle exec rake templates lexer compile
$ bundle exec ruby parse.rb
```

# Comparison

## v3.9.4

```
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [arm64-darwin23]
Warming up --------------------------------------
rbs v3.9.4 parse activerecord-generated.rbs
                         3.000 i/100ms
Calculating -------------------------------------
rbs v3.9.4 parse activerecord-generated.rbs
                         34.219 (± 0.0%) i/s   (29.22 ms/i) -    174.000 in   5.085551s
```

## v4.0.0.dev4

```
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [arm64-darwin23]
Warming up --------------------------------------
rbs v4.0.0.dev.4 parse activerecord-generated.rbs
                         2.000 i/100ms
Calculating -------------------------------------
rbs v4.0.0.dev.4 parse activerecord-generated.rbs
                         30.880 (± 3.2%) i/s   (32.38 ms/i) -    156.000 in   5.053877s
```

# Rate

```rb
34.219 / 30.880
=> 1.108128238341969
```

I could observe a degradation in parsing performance of about 10%.

amomchilov:
@soutaro did some performance analysis in #2652 and merged some fixes:

1. https://github.com/ruby/rbs/pull/2665
2. https://github.com/ruby/rbs/pull/2666

Is this enough to close this issue, Soutaro?

ksss:
I confirmed that there is no difference compared to the current master, and it is even 10% faster. I will close this issue.

v3.9.5

```
ruby 4.0.0dev (2025-12-04T02:27:56Z master 0af85a1fe2) +PRISM [arm64-darwin24]
Warming up --------------------------------------
[...]

PR #2652, opened 2025-09-02
Parsing performance

soutaro:
This is related to [the RBS file parsing performance degradation](https://github.com/ruby/rbs/issues/2563).

The RBS file parsing in rbs-4.0 (dev) is ~5x slower than rbs-3.9, based on the `core/**/*.rbs` files, especially for smaller files.

https://docs.google.com/spreadsheets/d/1dQBGIC1_zWco6c5OHH5VunP8ZuZLzwaMDB-U31OgUAc/edit?gid=0#gid=0

I implemented minor changes to improve the parsing performance, reusing empty array and hash objects. The changes improved the parsing performance slightly, but it still clearly slower than rbs-3.9.

### DIFF
--- a/benchmarks3.rb
+++ b/benchmarks3.rb
@@ -1,0 +1,20 @@
+require "rbs"
+require "benchmark/ips"
+require "csv"
+require "pathname"
+
+files = {}
+ARGV.each do |file|
+  content = File.read(file)
+  files[file] = RBS::Buffer.new(content: content, name: Pathname(file))
+end
+
+puts "Benchmarking parsing #{files.size} files..."
+
+Benchmark.ips do |x|
+  x.report("batch parsing") do
+    RBS::Parser._parse_signatures(files)
+  end
+
+  x.compare!
+end
--- a/ext/rbs_extension/main.c
+++ b/ext/rbs_extension/main.c
@@ -452,6 +452,53 @@ static VALUE rbsparser_lex(VALUE self, VALUE buffer, VALUE end_pos) {
     return results;
 }
 
+VALUE rbsparser_parse_signatures(VALUE self, VALUE files) {
+    Check_Type(files, T_HASH);
+
+    VALUE result_hash = rb_hash_new();
+    VALUE keys = rb_funcall(files, rb_intern("keys"), 0);
+
+    for (long i = 0; i < RARRAY_LEN(keys); i++) {
+        VALUE file_path = rb_ary_entry(keys, i);
+        VALUE buffer = rb_hash_aref(files, file_path);
+
+        VALUE string = rb_funcall(buffer, rb_intern("content"), 0);
+        StringValue(string);
+        rb_encoding *encoding = rb_enc_get(string);
+
+        // Prepare parser
+        rbs_parser_t *parser = alloc_parser_from_buffer(buffer, 0, RSTRING_LEN(string));
+        struct parse_signature_arg arg = {
+            .buffer = buffer,
+            .encoding = encoding,
+            .parser = parser,
+            .require_eof = false
+        };
+
+        VALUE entry = rb_hash_new();
+
+        int state = 0;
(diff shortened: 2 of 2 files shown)


soutaro:
* Try another allocator API, not memmap with linked list of heap (try reverting https://github.com/Shopify/rbs/pull/38)
* Try introduce batch loading API -- to reuse the allocators, reduce C-Ruby swtiching
* Reuse hash object for keyword arguments

amomchilov:
Another thought:

How much time do we spend computing assertions? Should we disable them in "release" mode?

Currently, `rbs_assert()` is just a function that's never disabled:

https://github.com/ruby/rbs/blob/4baf465ba2617a3b612ca3354e5438f4cb769e2e/src/util/rbs_assert.c#L8-L19

Perhaps we should rename the function to `rbs_assert_impl()`, and create an `#define rbs_assert()` macro that calls it only if we're in debug mode.

For comparison, C's standard library's [`assert.h`](https://en.cppreference.com/w/cpp/error/assert.html), makes `assert()` a no-op if you `#define NDEBUG`

amomchilov:
Another thought: our `extconf.rb` doesn't specify an optimization level. We disable optimizations (`-O0`) if `DEBUG` is defined:

https://github.com/ruby/rbs/blob/4baf465ba2617a3b612ca3354e5438f4cb769e2e/ext/rbs_extension/extconf.rb#L21

But we never specify `-O3` if we're in a "release" mode

soutaro:
It looks like `Makefile` has `-O3`.

```
...
cflags   = $(hardenflags) -fdeclspec  $(optflags) $(debugflags) $(warnflags)
optflags = -O3 -fno-fast-math
CFLAGS   = $(CCDLFLAGS) $(cflags) -fno-common -pipe -std=gnu99 -Wimplicit-fallthrough -Wunused-result -Wc++-compat $(ARCH_FLAG)
...
```

soutaro:
It looks like ~3% are spent for `rbs_assert` in debug build. Removing the calls in release build slightly improved.

soutaro:
Finally the parser is about twice faster than 3.9 parser!

The benchmark parses all of the files given as `ARGV`, and runs the benchmark-ips: Higher number means better.

```
> BUNDLE_GEMFILE=Gemfile-39 bundle exec ruby benchmarks2.rb core/**/*.rbs ../../ruby/gem_rbs_collection/gems/activerecord/8.0/*.rbs sig/**/*.rbs
Benchmarking parsing 177 files...
ruby 3.4.4 (2025-05-14 revision a38531fd3f) +PRISM [arm64-darwin23]
             parsing     1.000 i/100ms
             parsing      9.749 (± 0.0%) i/s  (102.58 ms/i) -     49.000 in   5.030113s
> bundle exec ruby benchmarks2.rb core/**/*.rbs ../../ruby/gem_rbs_collection/gems/activerecord/8.0/*.rbs sig/**/*.rbs                          
Benchmarking parsing 177 files...
ruby 3.4.4 (2025-05-14 revision a38531fd3f) +PRISM [arm64-darwin23]
             parsing     1.000 i/100ms
             parsing     19.478 (± 0.0%) i/s   (51.34 ms/i) -     98.000 in   5.032890s
```

src/lexstate.c
```diff
@@ -112,17 +115,60 @@ int rbs_token_bytes(rbs_token_t tok) {
 }
 
 unsigned int rbs_peek(rbs_lexer_t *lexer) {
-    if (lexer->current.char_pos == lexer->end_pos) {
-        lexer->last_char = '\0';
-        return 0;
+    return lexer->current_code_point;
```
soutaro, line 118: The `rbs_peek` simply returns the `current_code_point` that was read in `rbs_next_char`. We can skip calling expensive `rbs_utf8_string_to_codepoint` function.

soutaro:
Merged pull requests that is extracted from this PR. 🎉

PR #2665, opened 2025-09-26
Faster lexical analyzer

soutaro:
Extracted from #2652 

This PR improves the data structure of lexer in RBS.

It improves the parsing performance from ~`14 i/s` to ~`16 i/s` measured by `bin/benchmark-parse.rb`.

<details>
I have `gem_rbs_collection` repository too to load `activerecord` rbs files.

### Baseline

```
➜  rbs git:(38724282) bundle exec ruby bin/benchmark-parse.rb core/**/*.rbs ../../ruby/gem_rbs_collection/gems/activerecord/8.0/*.rbs sig/**/*.rbs
Benchmarking parsing 177 files...
✅ 14.506 i/s (±0.000%)
```

### Fix lexer

```
➜  rbs git:(fix-lexer) bundle exec ruby bin/benchmark-parse.rb core/**/*.rbs ../../ruby/gem_rbs_collection/gems/activerecord/8.0/*.rbs sig/**/*.rbs
Benchmarking parsing 177 files...
✅ 16.667 i/s (±0.000%)
```
</details>

### DIFF
--- a/include/rbs/lexer.h
+++ b/include/rbs/lexer.h
@@ -126,20 +126,26 @@ typedef struct {
  * The lexer state is the curren token.
  *
  * ```
- * ... "a string token"
- *    ^                      start position
- *          ^                current position
- *     ~~~~~~                Token => "a str
+ #.   0.1.2.3.4.5.6.7.8.9.0.1.2.3.4.5.6
+ * ... " a   s t r i n g   t o k e n "
+ *    ^                                   start position (0)
+ *                ^                       current position (6)
+ *                 ^                      current character ('i', bytes = 1)
+ *     ~~~~~~~~~~~                        Token => "a str
  * ```
  * */
 typedef struct {
     rbs_string_t string;
-    int start_pos;            /* The character position that defines the start of the input */
-    int end_pos;              /* The character position that defines the end of the input */
-    rbs_position_t current;   /* The current position */
-    rbs_position_t start;     /* The start position of the current token */
+    int start_pos;          /* The character position that defines the start of the input */
+    int end_pos;            /* The character position that defines the end of the input */
+    rbs_position_t current; /* The current position: just before the current_character */
+    rbs_position_t start;   /* The start position of the current token */
+
+    unsigned int current_code_point; /* Current character code point */
+    size_t current_character_bytes;  /* Current character byte length (0 or 1~4) */
+
     bool first_token_of_line; /* This flag is used for tLINECOMMENT */
-    unsigned int last_char;   /* Last peeked character */
+
     const rbs_encoding_t *encoding;
 } rbs_lexer_t;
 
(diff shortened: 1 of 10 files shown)


include/rbs/lexer.h
```diff
@@ -126,20 +126,26 @@ typedef struct {
...
-    int start_pos;            /* The character position that defines the start of the input */
-    int end_pos;              /* The character position that defines the end of the input */
-    rbs_position_t current;   /* The current position */
-    rbs_position_t start;     /* The start position of the current token */
+    int start_pos;          /* The character position that defines the start of the input */
+    int end_pos;            /* The character position that defines the end of the input */
+    rbs_position_t current; /* The current position: just before the current_character */
+    rbs_position_t start;   /* The start position of the current token */
+
+    unsigned int current_code_point; /* Current character code point */
```
soutaro, line 144: The lexer data structure now stores the code point of next character, so that peeking next character can be implemented really faster than reading next character from buffer.

src/lexstate.c
```diff
@@ -112,17 +114,60 @@ int rbs_token_bytes(rbs_token_t tok) {
 }
 
 unsigned int rbs_peek(rbs_lexer_t *lexer) {
-    if (lexer->current.char_pos == lexer->end_pos) {
-        lexer->last_char = '\0';
-        return 0;
+    return lexer->current_code_point;
+}
+
+bool rbs_next_char(rbs_lexer_t *lexer, unsigned int *codepoint, size_t *byte_len) {
```
soutaro, line 120: This function assigns the next codepoint in the buffer and it's byte length.

src/lexstate.c
```diff
@@ -112,17 +114,60 @@ int rbs_token_bytes(rbs_token_t tok) {
...
+    if (RBS_UNLIKELY(lexer->current.char_pos == lexer->end_pos)) {
+        return false;
+    }
+
+    const char *start = lexer->string.start + lexer->current.byte_pos;
+
+    // Fast path for ASCII (single-byte) characters
+    if ((unsigned int) *start < 128) {
```
soutaro, line 128: We assume the character encoding of RBS files is *ASCII compatible*, like Ruby source file.
If it is ASCII character, it is a single-byte character.

src/lexstate.c
```diff
@@ -112,17 +114,60 @@ int rbs_token_bytes(rbs_token_t tok) {
...
+        *codepoint = (unsigned int) *start;
     } else {
-        rbs_string_t str = rbs_string_new(
-            lexer->string.start + lexer->current.byte_pos,
-            lexer->string.end
-        );
-        unsigned int c = rbs_utf8_string_to_codepoint(str);
-        lexer->last_char = c;
-        return c;
+        *codepoint = 12523; // Dummy data for "ル" from "ルビー" (Ruby) in Unicode
```
soutaro, line 139: Another hack to support encoding other than utf-8.
It doesn't know the exact unicode code point of the next character in other encoding, and it returns a random code point instead. Lexer reads the character, but because the random character doesn't have any meaning for lexer, it works perfectly.

We may want to return a upper case character to support multi-byte class/constant names.

PR #2682, opened 2025-10-17
[Kernel] Add Kernel.trace_var and Kernel.untrace_var

sampersand:
This PR adds `Kernel.trace_var` and `Kernel.untrace_var` in, along with the `Kernel::_Tracer` interface.

This is a pretty simple PR to ease back into things!


ksss:
I agree with adding the methods, but I'm unsure how much management cost we should invest in debugging methods.

sampersand:
I don't see a problem with adding them in. I've also used them in code before, such as:
```ruby
class Logger
  class << self; attr_accessor :debug_lvl end
  trace_var :$DEBUG, &method(:debug_lvl=)
  trace_var :$-d, &method(:debug_lvl=)
end
```

PR #2683, opened 2025-10-17
[Kernel] Tweak signatures for conversion methods, and add unit tests in

sampersand:
This PR tweaks some of the `Kernel` conversion methods' signatures, as well as cleaning up/adding relevant tests.

Of particular note is that the numeric conversions (`Integer`/`Float`/`Complex`/`Rational`)'s final signature (the `(untyped, ?exception: bool) -> type?` one) now marks the `exception: bool` as required.

### DIFF
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -501,7 +501,7 @@ module Kernel : BasicObject
                    | (String real_or_both, ?exception: true) -> Complex
                    | (untyped real_or_both, exception: bool) -> Complex?
                    | (Numeric | String real, Numeric | String imag, ?exception: true) -> Complex
-                   | (Numeric | String real, Integer | Float | Rational | Complex imag, exception: bool) -> Complex
+                   | (Numeric | String real, Integer | Float | Rational | Complex imag, exception: bool) -> Complex?
                    | (Numeric | String real, untyped imag, exception: bool) -> Complex?
 
   # <!--
--- a/core/kernel.rbs
+++ b/core/kernel.rbs
@@ -497,9 +497,12 @@ module Kernel : BasicObject
   #
   def self?.Complex: (_ToC complex_like, ?exception: true) -> Complex
                    | (_ToC complex_like, exception: bool) -> Complex?
-                   | (Numeric | String real, ?Numeric | String imag, ?exception: true) -> Complex
-                   | (Numeric | String real, ?Numeric | String imag, exception: bool) -> Complex?
-                   | (untyped, ?untyped, ?exception: bool) -> Complex?
+                   | (Numeric numeric, ?exception: bool) -> Complex
+                   | (String real_or_both, ?exception: true) -> Complex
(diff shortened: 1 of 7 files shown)


core/kernel.rbs
```diff
@@ -614,7 +617,7 @@ module Kernel : BasicObject
                    | (int | _ToI int_like, exception: bool) -> Integer?
                    | (string str, int base, ?exception: true) -> Integer
                    | (string str, int base, exception: bool) -> Integer?
-                   | (untyped, ?untyped, ?exception: bool) -> Integer?
+                   | (untyped, ?int base, exception: bool) -> Integer?
```
sampersand: If provided, the second argument actually has to be an `int`, otherwise an error will be thrown (regardless of `exception`)

core/kernel.rbs
```diff
@@ -655,14 +658,19 @@ module Kernel : BasicObject
   #
   def self?.Rational: (int | _ToR rational_like, ?exception: true) -> Rational
                     | (int | _ToR rational_like, exception: bool) -> Rational?
-                    | (int | _ToR numer, ?int | _ToR denom, ?exception: true) -> Rational
-                    | (int | _ToR numer, ?int | _ToR denom, exception: bool) -> Rational?
-                    | [T] (Numeric & _RationalDiv[T] numer, Numeric denom, ?exception: bool) -> T
+                    | (int | _ToR numer, int | _ToR denom, ?exception: true) -> Rational
+                    | (int | _ToR numer, int | _ToR denom, exception: bool) -> Rational?
```
sampersand: I simplified the signature of this a bit, collapsing the ` int | _ToR numer, ?...` into the ones above it

test/stdlib/Kernel_test.rb
```diff
@@ -11,73 +11,208 @@ def test_Array
...
+      assert_send_type "(_ToF, exception: true) -> Float",
+                       Kernel, :Float, float_like, exception: true
+      assert_send_type "(_ToF, exception: bool) -> Float",
+                       Kernel, :Float, float_like, exception: false
     end
 
-    assert_send_type "(untyped, ?exception: bool) -> Float?",
-                     Kernel, :Float, :hello, exception: false
+    with_untyped do |untyped|
+      next if defined? untyped.to_f
```
ksss: I can't determine if the subsequent tests are actually being executed.

To determine this, it would be necessary to rely on the implementation of `with_untyped`.

A BasicObject, or some other object that is guaranteed not to have a `to_f` method, should be explicitly specified.

ksss:
@sampersand 
Thank you for your request.
However, due to the extensive number of changes, I am finding it difficult to review.
Would it be possible to split the changes into several Pull Requests?

soutaro:
@sampersand Can you add some `typecheck_test` cases so that Steep gives expected types to the `Complex` calls?

PR #2691, opened 2025-10-29
Accept `nil` state parameter in `to_json` type signatures

vpellan:
This PR adds `nil` state parameter to the type signature for all `to_json` methods. [`cState_from_state_s`](https://github.com/ruby/json/blob/a630388d4dc6f5ddf5d965e4b3844d627f47ceb1/ext/json/ext/generator/generator.c#L1567) will return an unconfigured instance if `nil` is passed. A lot of implementations of `to_json` have `nil` as the default value.

### DIFF
--- a/lib/rbs/ast/annotation.rb
+++ b/lib/rbs/ast/annotation.rb
@@ -21,7 +21,7 @@ module RBS
         self.class.hash ^ string.hash
       end
 
-      def to_json(state = _ = nil)
+      def to_json(state = nil)
         { string: string, location: location }.to_json(state)
       end
     end
--- a/lib/rbs/ast/comment.rb
+++ b/lib/rbs/ast/comment.rb
@@ -21,7 +21,7 @@ module RBS
         self.class.hash ^ string.hash
       end
 
-      def to_json(state = _ = nil)
+      def to_json(state = nil)
         { string: string, location: location }.to_json(state)
       end
     end
--- a/lib/rbs/ast/declarations.rb
+++ b/lib/rbs/ast/declarations.rb
@@ -74,7 +74,7 @@ module RBS
             self.class.hash ^ name.hash ^ args.hash
           end
 
-          def to_json(state = _ = nil)
+          def to_json(state = nil)
             {
               name: name,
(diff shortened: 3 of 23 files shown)


ksss:
@vpellan Thank you for your request.
Please tell me the specific use cases that will be improved by this change.
This is a breaking change, so we need information to make a decision.

vpellan:
Hey @ksss, thanks for your quick reply ! We're currently trying to improve typing in [dd-trace-rb](https://github.com/DataDog/dd-trace-rb/), and one of the first step is to re-enable the disabled rules in our Steepfile. When re-enabling [`Ruby::IncompatibleAssignment`](https://github.com/DataDog/dd-trace-rb/blob/0808070e7f28dad0b5a04e535ac05c376a7d8a45/Steepfile#L25), we stumbled on an error on the parameter of a [`to_json` method](https://github.com/DataDog/dd-trace-rb/blob/0808070e7f28dad0b5a04e535ac05c376a7d8a45/lib/datadog/profiling/collectors/code_provenance.rb#L139) that we've implemented.
If we change [its signature](https://github.com/DataDog/dd-trace-rb/blob/0808070e7f28dad0b5a04e535ac05c376a7d8a45/sig/datadog/profiling/collectors/code_provenance.rbs#L35) to `def to_json: (?JSON::State? state) -> String`, a new `Ruby::ArgumentTypeMismatch` error will appear at [line 140](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/profiling/collectors/code_provenance.rb#L140) saying that `Hash#to_json` does not accept `nil`. Of course we can skip it by changing the parameter to `arg = _ = nil` or adding a `# steep:ignore` comment (and even if this PR gets merged, we'll still have to do it until the next release anyway) but I thought that fixing the root cause could be a great overall improvement.

ksss:
@vpellan 
Thank you for the use case.
We plan to discuss it at our next developer meeting.

ksss:
@vpellan We have decided to accept this PR.
Please wait for RBS v4.
Thanks!

PR #2695, opened 2025-10-31
Deprecate top-level float

sampersand:
This PR deprecates the top-level `float` type, as it's mostly pointless and is not an implicit conversion (like all the other top-level alias types).

Most places you'd expect to use it (eg `Math` functions) actually require `Numeric & _ToF`. There's only a few minor edge-cases in the stdlib that use floats directly (eg `Regexp.timeout=`), but they can instead use `_ToF`.


sampersand:
cc @ParadoxV5 , @soutaro , and @ksss . Thoughts?

ParadoxV5:
> Thoughts?

Perhaps we could use a convenience alias for `Numeric & _ToF`, then.

sampersand:
It'd only really be used in `Math`, which already does this with `type double = Numeric & _ToF`

soutaro (approved):
👍

PR #2697, opened 2025-11-04
[Comparable] Add in Comparable::_CompareToZero

sampersand:
This PR adds in the `Comparable::_CompareToZero` interface, which is the actual return type that `Comparable` expects `<=>` to return.


ksss:
I believe it should return `Integer?` because the document says,

> returning a value less than 0, returning 0, or returning a value greater than 0

https://docs.ruby-lang.org/en/master/Comparable.html

Also, I cannot think of any specific problems that this change would solve.
Could you give me an example?

sampersand:
Yeah, the docs are correct—it actually checks the return valeu for `ret < 0` and `ret > 0`. 

While all of the stdlib returns an `Integer` (`-1`, `0`, or `1`), I believe `Comparable` was designed this way to let people do things like returning the diff between two Floats. For example:
```ruby
class Kilometers
  include Comparable
  
  def initialize(value) = @value = value.to_f
  def to_f = @value
  def <=>(other) = @value - other.to_f
  # ...
end

p Kilometers.new(1.2) < Kilometers.new(3.4)
```

This doesn't typecheck under the current RBS signatures, but does work in Ruby

ParadoxV5:
> ```ruby
>   def <=>(other) = @value - other.to_f
> ```

```ruby
  def <=>(other) = @value <=> other.to_f
```
The point is still valid: `MyNumber` would be a better example.

sampersand:
@ParadoxV5 this is actually the way that [ruby 0.49 did it](https://github.com/sampersand/ruby-0.49):
```shell
% ./fixed/ruby <<RUBY
class X
   include Comparable
   def <=>(r) "A" <=> r end
end

print(X.new() <=> "Q")
RUBY
-16
```

PR #2701, opened 2025-11-07
Remove reference to `JSON.deep_const_get`

byroot:
This was always meant as internal API and was removed in `json 2.16.0`.


ksss:
This PR is necessary to fix the failure of the bundle update job.
https://github.com/ruby/rbs/actions/runs/19250653373

Thanks!

PR #2706, opened 2025-11-12
Skip io/wait tests for methods deleted in Ruby 4

XrXr:
See: https://github.com/ruby/io-wait/pull/49

---

Note that you probably want to also make changes to https://github.com/ruby/rbs/blob/master/core/io/wait.rbs to remove `ready?` and `nread`. I'm not sure the best way to go about it would be, though.

### DIFF
--- a/test/stdlib/IO_test.rb
+++ b/test/stdlib/IO_test.rb
@@ -528,20 +528,22 @@ class IOWaitTest < Test::Unit::TestCase
   def test_readyp
     # This method returns true|false in Ruby 2.7, nil|IO in 3.0, and true|false in 3.1.
 
-    IO.pipe.tap do |r, w|
-      assert_send_type(
-        "() -> untyped",
-        r, :ready?
-      )
-    end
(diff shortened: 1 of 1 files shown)


ksss:
@XrXr 
The next version of RBS v4 is planned to be released after Ruby v4.

The target version for RBS signatures follows the latest Ruby as follows:
https://github.com/ruby/rbs?tab=readme-ov-file#the-target-version

If methods does not exist at Ruby v4, please remove the signature and testing.

ksss:
@XrXr I have taken the liberty of removing the type definition on my end.
https://github.com/ruby/rbs/pull/2710

XrXr:
Thanks!

nobu:
> The next version of RBS v4 is planned to be released after Ruby v4.

Then Ruby 4.0.0 will bundle RBS 3.9.5?

ksss:
@nobu We're planning to bundle RBS **v3.10** with Ruby v4.

nobu:
@ksss Thank you

PR #2707, opened 2025-11-12
[do not merge] io/wait test fix on top of 3.9.5

XrXr:
This exist to fix ruby/ruby CI until #2706 is merged and gems/bundled_gems in ruby/ruby can be bumped to rbs trunk. See https://github.com/ruby/ruby/commit/de6c4b62b00023ef372bae4c434f58014c3c8382 and the CI failures in parent commits.


XrXr:
We've a release past this. Thanks!

PR #2721, opened 2025-11-27
Update rdoc to v6.16

ksss:
Interface of `RDoc::AnyMethod#start_collecting_tokens` has been changed from rdoc v6.16.
https://github.com/ruby/rdoc/pull/1471

I think it would be best to support the latest version.


ksss:
The interpretation of rdoc has changed, resulting in differences in comments, but the fundamental issue has been resolved.

* thread.c => Fixed from https://github.com/ruby/ruby/pull/12662 (v4.0.0)
* openssl => Fixed from https://github.com/ruby/openssl/pull/970

PR #2727, opened 2025-12-03
Add RFC2396_PARSER to URI module

sled:
The URI module defines the constant `URI::RFC2396_PARSER` [here](https://github.com/ruby/uri/blob/master/lib/uri/common.rb#L17C3-L17C17) but it was missing in the RBS declaration, so I got the following warning:

```rb
URI::RFC2396_PARSER.escape("https://example.org")
```

```sh
something.rb:32:29: [warning] Cannot find the declaration of constant: `RFC2396_PARSER`
│ Diagnostic ID: Ruby::UnknownConstant
│
└         URI::RFC2396_PARSER.escape("https://example.org")
```

### DIFF
--- a/stdlib/uri/0/common.rbs
+++ b/stdlib/uri/0/common.rbs
@@ -561,6 +561,11 @@ URI::REL_URI: Regexp
 URI::REL_URI_REF: Regexp
 
 # <!-- rdoc-file=lib/uri/common.rb -->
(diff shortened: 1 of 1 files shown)


sled:
Rubocop causes a segmentation fault, I guess this is unrelated to the change.

ksss:
Thanks!

PR #2733, opened 2025-12-09
Update rdoc and comments

ksss:
The table notation in rdoc v6.17.0 has been corrected, resulting in a more Markdown-like representation.

ref: https://github.com/ruby/rdoc/pull/1483


soutaro:
We will try to update RDoc to 6.17.0 on 3.10 too.

PR #2745, opened 2025-12-12
[Kernel] Narrow `caller_locations`

vpellan:
This PR narrows the type of `caller_locations` if `start` parameter is not given. `caller_locations` can return `nil` only if it has a `start` parameter.


ksss (commented):
@vpellan 
Thank you for the PR. The changes look good.
Could you add some test code?

vpellan:
Hello! Thanks! I added tests in 819bdf017c32665140eabfbd3124b2f07bc3adcb

PR #2749, opened 2025-12-16
Fix `nil` size `Enumerator` test

soutaro:
If the `size` is `nil`, which means the size is unknown:

1. Test with loop but break after 10,000 loops.
2. If the object has more than 10,000 objects, assume it’s an infinite loop.

This is clearly broken for enumerators with more than 10,000 but fixed size. But it works for tests.


soutaro:
This is a fix for https://github.com/ruby/ruby/pull/15541.

PR #2758, opened 2025-12-17
Support BigDecimal v4

ksss:
BigDecimal v4 has been released, with an increase and decrease in methods and changes to signatures.
This PR addresses these changes.

See also:
https://github.com/ruby/bigdecimal/blob/master/CHANGES.md#400

### DIFF
--- a/stdlib/bigdecimal-math/0/big_math.rbs
+++ b/stdlib/bigdecimal-math/0/big_math.rbs
@@ -129,8 +129,6 @@ module BigMath
   #
   def self?.cos: (BigDecimal x, Numeric prec) -> BigDecimal
 
-  # <!--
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -58,9 +58,8 @@ GEM
       rdoc (>= 4.0.0)
       reline (>= 0.4.2)
     json (2.18.0)
-    json-schema (6.0.0)
+    json-schema (5.1.0)
       addressable (~> 2.8)
-      bigdecimal (~> 3.1)
```
ksss: json-schema v6 only supports up to bigdecimal v3. In this change, we will prioritize to support bigdecimal v4 and proceed with bigdecimal v4 support for json-schema separately.

PR #2766, opened 2025-12-19
[Backport] Check tuple type length

ksss:
* https://github.com/ruby/rbs/pull/2765


ksss:
We need 4.0.0-preview3.

PR #2781, opened 2025-12-30
Update {Method,UnboundMethod,Proc}#source_location signatures to match Ruby 4.1

eregon:
* Needed for https://github.com/ruby/ruby/pull/15580

### DIFF
--- a/core/method.rbs
+++ b/core/method.rbs
@@ -400,7 +400,7 @@ class Method
   # This method will return `nil` if the method was not defined in Ruby (i.e.
   # native).
   #
-  def source_location: () -> [String, Integer]?
+  def source_location: () -> ([String, Integer]? | [String, Integer, Integer, Integer, Integer]?)
 
   # <!--
   #   rdoc-file=proc.c
--- a/core/proc.rbs
+++ b/core/proc.rbs
@@ -845,7 +845,7 @@ class Proc
   # This method will return `nil` if the Proc was not defined in Ruby (i.e.
   # native).
   #
-  def source_location: () -> [String, Integer]?
+  def source_location: () -> ([String, Integer]? | [String, Integer, Integer, Integer, Integer]?)
 
   # <!--
   #   rdoc-file=proc.c
--- a/core/unbound_method.rbs
+++ b/core/unbound_method.rbs
@@ -300,7 +300,7 @@ class UnboundMethod
   # This method will return `nil` if the method was not defined in Ruby (i.e.
   # native).
   #
-  def source_location: () -> [String, Integer]?
+  def source_location: () -> ([String, Integer]? | [String, Integer, Integer, Integer, Integer]?)
 
   # <!--
   #   rdoc-file=proc.c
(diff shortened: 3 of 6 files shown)


eregon:
This doesn't work in CRuby CI, see https://github.com/ruby/ruby/actions/runs/20595820151/job/59150151742?pr=15580
What should I do then?
Exclude the tests with `tool/rbs_skip_tests`?
Or rebase this on top of 3.10.0 since the failures might be caused by other changes on master?

eregon:
From https://github.com/ruby/rbs/pull/1203#issuecomment-1376565639

> Generally, the core library RBS definition targets the latest release of Ruby.

So if that's still the case, and it seems to be according to [docs](https://github.com/ruby/rbs/blob/master/docs/CONTRIBUTING.md#introduction) then I should use `tool/rbs_skip_tests` I guess.
Those docs should mention that then.

BTW [these errors](https://github.com/ruby/ruby/actions/runs/20595820151/job/59150151742?pr=15580) mention
> 💡You can skip this test `test_each_paragraph(RBS::AST::Ruby::CommentBlockTest)` by adding the name to `/home/runner/work/ruby/ruby/src/tool/rbs_skip_tests`

But [these errors](https://github.com/ruby/ruby/actions/runs/20593644069/job/59143607016?pr=15580) don't.

I also wonder if it's good to run RBS tests related to the core library in ruby/ruby CI since we know at least some of them will fail, so not sure there is much value.
This repo seems to test against `ruby-head`, not sure how that works though then given those differences.

soutaro:
@eregon Hi! Sorry for the late reply.

We’ve discussed the idea of moving the RBS definitions for core/stdlib classes into the `ruby/ruby` repository, so that committers can fix method types and their tests together with the Ruby implementation. I’m confident we’ll do this, though probably not within the next few weeks. (Maybe within months...)

For the current core/stdlib test failures, please use `rbs_skip_tests` in `ruby/ruby`. I’ll update the documentation as you suggested.

The bundled version of RBS is `3.10`, but the master branch targets RBS `4.0` (unreleased), which is causing the `CommentBlockTest` failure. (Though I’m still not sure why.)

eregon:
> We’ve discussed the idea of moving the RBS definitions for core/stdlib classes into the `ruby/ruby` repository, so that committers can fix method types and their tests together with the Ruby implementation. I’m confident we’ll do this, though probably not within the next few weeks. (Maybe within months...)

That would be great!

> For the current core/stdlib test failures, please use `rbs_skip_tests` in `ruby/ruby`.

Right, it's what I did.

> I’ll update the documentation as you suggested.

Thanks!

eregon:
> This repo seems to test against `ruby-head`, not sure how that works though then given those differences.

Any insight on this? At first sight it seems weird that rbs master + ruby master works in ruby/rbs CI but not in ruby/ruby CI. I suppose they run different tests?

eregon:
The change got reverted in ruby/ruby, so no change needed here.

PR #2867, opened 2026-03-05
update rbs signatures to be more consistent with generics

sampersand:
TL;DR: This migrates RBS signatures to use single-characters instead of multiple charcaters.

RBS is a bit inconsistent about how generic variables (eg `Foo[X, Y]`) are named: Some of them are long (`Array[Elem]`, `Enumerable[Elem]`, `ObjectSpace::WeakMap[Key, Value]`), whereas some of them are short (`Hash[K, V]`, `Set[A]`, and most exceptions like `FrozenError[T]`, `KeyError[K, R]`.)

I talked with @soutaro about this, and he liked the idea of the one-charcater variables (which I'm also partial to—it's how most langauges work, and makes it easier to distinguish them from classnames). This PR implements that

### DIFF
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -546,8 +546,8 @@
 #     given block.
 #
 %a{annotate:rdoc:source:from=array.c}
-class Array[unchecked out Elem] < Object
-  include Enumerable[Elem]
+class Array[unchecked out E] < Object
+  include Enumerable[E]
 
   # <!--
   #   rdoc-file=array.c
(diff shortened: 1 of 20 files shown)


ParadoxV5:
I prefer `E` for Element. Just sayin’

sampersand:
> I prefer `E` for Element. Just sayin’

Done. It also prevents a lot of annoying conflicts where `T` was already used as a generic (eg `Array.find: [T]( ...)`)

soutaro (approved):

PR #2898, opened 2026-03-24
[Class] update class.rbs

### LINKED ISSUE #2124: Ensure all signatures for `class.rbs` are correct
This is part of @sampersand's RBS work, to ensure that all `core/xxx.rbs` files are double-checked, and confirmed correct

sampersand:
This closes https://github.com/soutaro/steep/issues/2124 by updating `class.rbs`. Few important things to note that I'd like input on:

1. I've added a `[self: Class]` to `Class#initialize`—it seems correct, but if we should remove that LMK

And other things to note:
1. Technically `Class` defines both a singleton `Class.allocate` _and_ an instance `Class#allocate`. I've gone ahead and included both for completeness, and it shouldn't matter
2. `inherited` is now private (as it should be) and returns `void` (as nothing in Ruby uses its return value)
3. `new`'s signature now accepts anything
4. `attached_object` still returns `untyped`


ParadoxV5:
> 2. `attached_object` now returns an `instance` instead of `untyped`. Is this correct?

On second thought: It’s not correct.
It should be `instance.instance` (instance²), or `T where self < singleton(T)` (see #1542).

sampersand:
Yeah, i've just removed it entirely and it's now just `untyped` again

PR #2916, opened 2026-04-02
Add signature for `RubyVM::InstructionSequence.of`

Earlopain:
https://docs.ruby-lang.org/en/4.0/RubyVM/InstructionSequence.html#method-c-of

`prism` uses it and currently vendors this change.

The return type is nillable as per the implementation in ruby but I'm not sure how to actually trigger that for valid input. It may happen with some procs, that's all I know.


soutaro (approved):
👍

PR #2924, opened 2026-04-07
fix OpenSSL#session_new_cb signature

HoneyryderChuck:
the callback for #session_new_cb receives two arguments: the ssl socket and the session


soutaro (approved):
👍

PR #2927, opened 2026-04-08
Add RBS::Rewriter and use it in rbs annotate

soutaro:
## Summary

- Add `RBS::Rewriter` class for targeted character-range replacements on `Buffer` content
- Provides `add_comment`, `replace_comment`, `delete_comment` methods for comment manipulation
- Refactor `rbs annotate` to use `Rewriter` instead of `Writer`, preserving non-documentation comments and original formatting (inline comments, multi-line definitions, etc.)
- Deprecate `preserve:` parameter in `annotate_file` since `Rewriter` always preserves formatting

## Motivation

`RBS::Writer` regenerates the entire file from AST, which loses non-documentation comments (inline comments like `# Some comment here` at end of lines). `RBS::Rewriter` only modifies the specific locations that need to change, preserving everything else.

## Test plan

- [x] `ruby -Ilib -Itest test/rbs/rewriter_test.rb` — 20 tests for Rewriter
- [x] `ruby -Ilib -Itest test/rbs/annotate/rdoc_annotator_test.rb` — 10 tests for annotator integration


soutaro:
I plan to implement more APIs on `Rewriter` class, but currently it has only `Comment` related ones.

PR #2933, opened 2026-04-14
Update Module

sampersand:
This PR updates the signature for `Module`, as well as updating all its tests, and adding in an `assert_visibility` method for testing


sig/unit_test/type_assertions.rbs
```diff
@@ -179,6 +179,10 @@ module RBS
       #
       def assert_type: (String | Types::t value_type, untyped value) -> void
 
+      # Asserts if given `value` has a type of `value_type`
```
soutaro:
```suggestion
      # Asserts if RBS definition declares given `method_name` with `visibility`
```

soutaro (approved):

PR #2934, opened 2026-04-15
Add RBS::Ops 

sampersand:
[WIP]

This PR adds in `RBS::Ops`, a uniform location for all operator interfaces.

### DIFF
--- a/core/rbs/ops.rbs
+++ b/core/rbs/ops.rbs
@@ -1,0 +1,154 @@
+module RBS
+  module Ops
+    # An interface for types implementing the `+` operator
+    interface _Add[O, R = self]
+      # Perform the `+` operation
+      def +: (O other) -> R
+    end
+
+    # An interface for types implementing the `-` operator
+    interface _Subtract[O, R = self]
+      # Perform the `-` operation
+      def -: (O other) -> R
+    end
+
+    # An interface for types implementing the `*` operator
+    interface _Times[O, R = self]
+      # Perform the `*` operation
+      def *: (O other) -> R
+    end
+
+    # An interface for types implementing the `/` operator
+    interface _Divide[O, R = self]
+      # Perform the `/` operation
+      def /: (O other) -> R
+    end
+
+    # An interface for types implementing the `%` operator
+    interface _Modulo[O, R = self]
+      # Perform the `%` operation
+      def %: (O other) -> R
+    end
+
+    # An interface for types implementing the `**` operator
+    interface _Power[O, R = self]
+      # Perform the `**` operation
+      def **: (O other) -> R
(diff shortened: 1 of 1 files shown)


sampersand:
Open questions:
1. Do we want the names to be short (`_Add`, `_Sub`, `_Leq`, `_Neq`), medium (`_Add`, `_Subtract`, `_LessThanOrEqual`, `_NotEqual`), or long (`_LessThanOrEqualTo`, `_NotEqualTo`)? I've currently picked the medium ones
2. Do we want to use math operator names like `_Multiply` / `_Modulo`, or do we want to use more use-case-agnostic ones like `_Times` (and what should `%` be in that case?)
3. I've not made interfaces for `eql?` or `equal?`, as they're not formal operators. Do we want interfaces for them here too?
4. I've not made interfaces for `[]` or `[]=` as their arity can vary, and thus doesn't fit well into the type system. (Same with a `_Call` interface, which would be nice eventually, but has similar problems)
5. I've made heavy use of default interface arguments, eg with `<=`, as most of the time you just want a `bool` return value (but you _could_ specify a different one if needed):
```rbs
    # An interface for types implementing the `<=` operator
    interface _LessThanOrEqual[O, R = bool]
      # Perform the `<=` operation
      def <=: (O other) -> R
    end
```
5. (continued, i can't figure out GitHub markdown) I've left the first argument usually required (except for `_Equals`, `_Compare`, and `_NotEquals` where it defaults to `untyped`, as their signatures traditionally are). Do we want to have the `O` parameter default to `self` on most of the other operators?
6. Where and how do we want to test these interfaces? we don't historically have interface tests. Just typechecks?

ParadoxV5:
> 1. Do we want the names to be short (`_Add`, `_Sub`, `_Leq`, `_Neq`), medium (`_Add`, `_Subtract`, `_LessThanOrEqual`, `_NotEqual`), or long (`_LessThanOrEqualTo`, `_NotEqualTo`)? I've currently picked the medium ones
> 2. Do we want to use math operator names like `_Multiply` / `_Modulo`, or do we want to use more use-case-agnostic ones like `_Times` (and what should `%` be in that case?)

I suggest matching `parse.y`/Prism token names.

sampersand:
@ParadoxV5 —while we could, I was thinking we could take this opportunity to be more user-friendly. `parse.y` uses shorthands like `GEQ` ([link](https://github.com/ruby/ruby/blob/634707a7255f132eb486eaf57473925c288ef7bd/parse.y#L2828)) which people outside of parser development might not understand

sampersand:
Cc @soutaro

soutaro (approved):

PR #2935, opened 2026-04-18
Support `def self.method_name` singleton method in inline parser

RYOTA-KOBA:
Previously, the inline parser emitted a diagnostic error for any method definition with a receiver (`def self.foo`, `def obj.foo`, etc.) and skipped processing it entirely. This meant that singleton methods defined with `def self.method_name` could not have their type annotations recognized.

This change adds support for `def self.method_name` by:

- Checking `Prism::SelfNode` specifically in `visit_def_node` and only rejecting non-self receivers (e.g., `def obj.foo`)
- Adding a `kind` attribute (`:instance` | `:singleton`) to `DefMember` to distinguish between instance and singleton methods
- Propagating `kind` through `Environment#resolve_type_names`
- Routing singleton `DefMember`s to `build_singleton` in `MethodBuilder`

All existing inline annotation formats (`#:`, `# @rbs`, `# @rbs param:`) work with singleton methods as well.

## In addition

Could you also add the appropriate milestone to this PR? I do not have permission to set milestones from the GitHub UI. If this PR does not need one, please add the `no-milestone` label instead.

### DIFF
--- a/lib/rbs/ast/ruby/members.rb
+++ b/lib/rbs/ast/ruby/members.rb
@@ -549,15 +549,25 @@ module RBS
 
           attr_reader :name
           attr_reader :node
+          attr_reader :kind
           attr_reader :method_type
           attr_reader :leading_comment
 
-          def initialize(buffer, name, node, method_type, leading_comment)
+          def initialize(buffer, name, node, method_type, leading_comment, kind: :instance)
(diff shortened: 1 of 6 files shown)


soutaro (approved):
Thanks! 👍

PR #2938, opened 2026-04-22
Fix rbs test use cases

HoneyryderChuck:
In the process of porting usage of rbs runtime checks via env vars onto `rbs test` in the `httpx` repository, there were some setbacks that were addressed by these fixes:

* using `--log-level` wasn't working
* manifest.yaml (a recommendation of RBS docs for libs) isn't being used when running tests from the lib repo. this fixes it by searching for a "manifest.yaml" file in the sig dirs.
* I initially added a test to support RBS_TEST_RAISE via a --raise parameter, only to find out that it's now raising by default. I left the test around, but if you think it's not adding much, lmk and I can remove it.

### DIFF
--- a/lib/rbs/cli.rb
+++ b/lib/rbs/cli.rb
@@ -1061,7 +1061,7 @@ EOB
       env_hash = {
         'RUBYOPT' => "#{ENV['RUBYOPT']} -rrbs/test/setup",
         'RBS_TEST_OPT' => test_opt(options),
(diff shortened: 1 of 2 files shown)


soutaro (changes requested):
Thanks!

I cannot approve the *fix* related to `manifest.yaml`. I'm not sure if reading `manifest.yaml` is good for `-I` directories, and the implementation has some regression.

Can you extract that change to another PR so that we can merge other two fixes?

HoneyryderChuck:
@soutaro done ([PR](https://github.com/ruby/rbs/pull/2986))

PR #2939, opened 2026-04-24
Auto-generate rbs_ast_ruby_annotations_t union from config.yml

ksss:
## Summary

- The `rbs_ast_ruby_annotations_t` union in `include/rbs/ast.h` was hand-written with only 6 members, while `ast.h` defines 14 `rbs_ast_ruby_annotations_*_t` structs and `src/parser.c` casts all 14 to `rbs_ast_ruby_annotations_t *`. The code worked only because downstream consumers (e.g. `ext/rbs_extension/ast_translation.c`) just read `.base` and re-cast to the concrete struct — the union value itself is never instantiated.
- Teach `templates/template.rb` to collect nodes under `RBS::AST::Ruby::Annotations::*` and render the union body from them in `templates/include/rbs/ast.h.erb`, so future annotation additions stay in sync automatically.
- As a side benefit, replace the 13 hand-maintained `allowlist_type` entries in `rust/ruby-rbs-sys/build.rs` with a single regex `rbs_ast_ruby_annotations_.*`, so bindgen picks up every annotation struct (and the union itself) regardless of whether we build against the pinned vendor or the current master. This also silently closes the missing-entry for `rbs_ast_ruby_annotations_module_self_annotation_t`.

## Why the Rust drift went unnoticed

`rust/ruby-rbs-sys/build.rs` was last updated on 2026-03-12; `RBS::AST::Ruby::Annotations::ModuleSelfAnnotation` was added to `config.yml` on 2026-04-06. The Rust CI workflow pins `rbs_version = v4.0.2` and vendors that tag via `rake rust:rbs:sync`, so the CI build consumes a `config.yml` that predates the new annotation. `cargo build` therefore only failed for developers who use `rake rust:rbs:symlink` to work against the current master. After this PR, the regex-based allowlist stays in sync with `config.yml` on both sides (pinned vendor and master), so adding a new annotation only requires editing `config.yml`.

## Test plan

- [x] `bundle exec rake compile` (clean, no warnings)
- [x] `bundle exec rake test` — 936 tests / 0 failures
- [x] `cargo clean && cargo test` under `rake rust:rbs:sync` (CI-equivalent setup, pinned v4.0.2 vendor)
- [x] `cargo clean && cargo test` under `rake rust:rbs:symlink` (master `HEAD` vendor — previously failed with `cannot find type 'rbs_ast_ruby_annotations_module_self_annotation_t'`)
- [x] `cargo clippy` — no warnings in both modes
- [x] Generated `include/rbs/ast.h` matches template output after `clang-format` (equivalent to `rake confirm_templates`)


soutaro (approved):
👍

PR #2942, opened 2026-04-29
Update File::Constants and File::Stat

sampersand:
This PR cleans up miscellaneous `File` RBS definitions: `File`'s top-level constants, `File::Constants` and `File::Stat` (which were mostly correct, barirng a few nilable values). This is in preparation for another subsequent PR which will update the core `File` class.

Also note that this PR splits `File::Constants` and `File::Stat` into their own `core/file_{constants,stat}.rbs` files. If that's not desired, I can stick them back in `core/file.rbs`, but I think it's cleaner to have them as separate files. (And, we already have separate test files for them anyways, so it makes sense.)


soutaro (approved):
Thanks! 👍

PR #2946, opened 2026-05-04
Reduce Hash allocation per `Node` creation

amomchilov:
* 1% speed up on `bundle exec ruby bin/benchmark-parse.rb core`
* 3% on `activerecord.rbs`


soutaro (approved):
Thanks! 👍

PR #2953, opened 2026-05-13
Update docs/inline.md to match current inline parser behavior

ksss:
Comparing `lib/rbs/inline_parser.rb` against `docs/inline.md` surfaced several
items that were stale or missing in the documentation:

- Removed outdated method limitations (singleton methods and splat/block
  parameters were already supported).
- Added a note that singleton methods (`def self.NAME`) are supported, with
  an example in the Method Definitions intro.
- Documented the trailing `#:` annotation for method return types.
- Fixed constant type inference: a bare symbol literal infers to `Symbol`,
  not the singleton symbol type; also added `nil` to the inferred list.
- Added a Skip Annotation section for `@rbs skip`, which was previously
  undocumented despite being a first-class feature.
- Added limitations: `class << self`, top-level `def`, non-self receiver
  (`def obj.foo`), top-level `attr_*`.
- Fixed two typos: "hyphones" → "hyphens", "infered" → "inferred".

### DIFF
--- a/docs/inline.md
+++ b/docs/inline.md
@@ -163,6 +163,8 @@ Inline RBS supports methods defined using the `def` syntax in Ruby.
 ```ruby
 class Calculator
   def add(x, y) = x+y
+
+  def self.zero = 0
 end
 ```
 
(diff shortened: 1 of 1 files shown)


soutaro (approved):
🙏

PR #2957, opened 2026-05-16
Flyweight TypeName / Namespace cached in a shared trie

tk0miya:
Cache canonical TypeName / Namespace instances behind two positional factory methods, and route every allocation site — Ruby helpers, the C parser, the resolver, and the environment — through them. Repeated construction of structurally equal values returns the same object instead of allocating fresh.

  Namespace[path, absolute]
  TypeName[namespace, name]

Namespace interns into a per-`absolute` trie of nested Hashes keyed on path Symbols. TypeName interns into a two-level Hash keyed by canonical Namespace identity and name Symbol. Both fast paths are lock-free; cache misses take a mutex.

The C parser calls `Namespace[]` / `TypeName[]` via `rb_funcallv`; an in-C trie walk that read intern internals through `rb_hash_lookup` was identical in wall time on Ruby 4.0+ and was dropped to avoid coupling the parser to the cache layout.

Measured against the kaigionrails/conference-app project's RBS collection (3,903 type names), 1 warmup + 5 runs, median:

  Ruby 3.4.9   master 1.247s wall / 0.63s user
               branch 1.158s wall / 0.57s user   (-7% wall, -10% user)
  Ruby 4.0.4   master 1.175s wall / 0.57s user
               branch 1.146s wall / 0.55s user   (-2.5% wall, -4% user)

Live RBS::Namespace 51,436 → 6,904; RBS::TypeName 78,493 → 22,819 after `rbs list` on Ruby 4.0.

### DIFF
--- a/ext/rbs_extension/ast_translation.c
+++ b/ext/rbs_extension/ast_translation.c
@@ -177,6 +177,30 @@ VALUE rbs_type_param_variance_to_ruby(enum rbs_type_param_variance value) {
     rb_class_new_instance(argc, argv, receiver)
 #endif
 
+// Route Namespace / TypeName construction through the Ruby-side
+// flyweight cache (`RBS::Namespace.[]` / `RBS::TypeName.[]`) so that
+// structurally equal values produced by the parser share canonical
+// instances. An in-C trie walk was tried but did not beat
+// `rb_funcallv` on Ruby 4.0+, where method dispatch is well optimized.
+static ID id_intern_brackets;
+
+static inline ID intern_brackets(void) {
+    if (!id_intern_brackets) id_intern_brackets = rb_intern("[]");
+    return id_intern_brackets;
+}
+
+static VALUE rbs_intern_namespace(rbs_translation_context_t ctx, rbs_namespace_t *node) {
+    VALUE args[2];
+    args[0] = rbs_node_list_to_ruby_array(ctx, node->path);
+    args[1] = node->absolute ? Qtrue : Qfalse;
+    return rb_funcallv(RBS_Namespace, intern_brackets(), 2, args);
+}
+
+static VALUE rbs_intern_type_name(VALUE namespace, VALUE name) {
+    VALUE args[2] = { namespace, name };
+    return rb_funcallv(RBS_TypeName, intern_brackets(), 2, args);
+}
+
 VALUE rbs_struct_to_ruby_value(rbs_translation_context_t ctx, rbs_node_t *instance) {
     if (instance == NULL) return Qnil;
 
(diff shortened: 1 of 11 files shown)


tk0miya:
This also improves the performance of steep (with kaigionrails/conference-app):
(-4.2% wall, -5.5% user)

```
$ ruby -v
ruby 4.0.4 (2026-05-12 revision b89eb1bcbf) +PRISM [arm64-darwin24]
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  32.32s user 1.84s system 570% cpu 5.991 total
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  32.40s user 1.84s system 599% cpu 5.707 total
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  32.45s user 1.81s system 597% cpu 5.730 total
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  32.22s user 1.82s system 599% cpu 5.676 total
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  32.37s user 1.89s system 593% cpu 5.771 total
```
```
$ ruby -v
ruby 4.0.4 (2026-05-12 revision b89eb1bcbf) +PRISM [arm64-darwin24]
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  30.96s user 1.94s system 541% cpu 6.072 total
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  30.74s user 1.83s system 586% cpu 5.551 total
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  30.71s user 1.81s system 592% cpu 5.488 total
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  30.06s user 1.81s system 593% cpu 5.371 total
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  29.93s user 1.90s system 577% cpu 5.510 total
$ time bundle exec steep check > /dev/null
bundle exec steep check > /dev/null  30.48s user 1.88s system 591% cpu 5.472 total
```

soutaro (approved):
Hmm. This requires a difficult architectural decision. What if we want to multi-ractor processing in RBS?

Let's merge this for now, but we may revisit this when we want to make things in RBS parallel for more performance later.

tk0miya:
@soutaro  Thank you for reviewing & merging.

> What if we want to multi-ractor processing in RBS?

Sorry for the lack of consideration for Ractor. AFAIK, to support Ractor, it's better to move these classes to the C extension.

In this PR, I did not change the implementation of TypeName and Namespace as much as possible to keep the proposal simple. Additionally, I'm not sure it is a better way to implement the whole of them as C.

If you need, I will work on another PR to reimplement them as C. What do you think?

PR #2959, opened 2026-05-20
Update `StringScanner` signatures to match strscan 3.1.6

zonuexe:
`StringScanner#[]` accepts a capture-group name as a String or Symbol in addition to an Integer index — the method's own RD doc in this file documents and demonstrates it ("`specifier` symbol or string. returns the named subgroup", with `scanner[:wday]` / `scanner['wday']` examples). The signature only listed `Integer`, so named-capture lookups type-check as an argument error.

see https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L1705-L1715

---

## Update StringScanner signatures to match strscan 3.1.6

The follow-up commit aligns the rest of `StringScanner` with strscan 3.1.6 (bundled with Ruby 4.0): it widens the matcher arguments to accept `String` patterns, corrects several return types, and adds methods introduced in recent releases. The table lists one method per row; the pattern-matching methods that share a single C helper are described below it.

| Ruby Method | Change | C function |
|-------------|--------|------------|
| `[]` | Accept a capture-group name (`Integer` → `Integer \| String \| Symbol`) | [`strscan_aref`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L1695-L1726) |
| `initialize` | Drop the obsolete `dup` parameter; return type `void` | [`strscan_initialize`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L254-L282) |
| `named_captures` | New method (`() -> Hash[String, String?]`) | [`strscan_named_captures`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L2176-L2189) |
| `peek_byte` | New method (`() -> Integer?`) | [`strscan_peek_byte`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L1173-L1183) |
| `post_match` | Return type `String` → `String?` | [`strscan_post_match`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L1917-L1927) |
| `pre_match` | Return type `String` → `String?` | [`strscan_pre_match`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L1880-L1890) |
| `reset` | Return type `void` → `self` | [`strscan_reset`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L364-L373) |
| `scan_byte` | New method (`() -> Integer?`) | [`strscan_scan_byte`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L1148-L1165) |
| `scan_integer` | New method (`(?base: Integer) -> Integer?`) | [`lib/strscan/strscan.rb`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/lib/strscan/strscan.rb#L4-L24) |
| `size` | Return type `Integer` → `Integer?` | [`strscan_size`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L1752-L1760) |
| `terminate` | Return type `void` → `self` | [`strscan_terminate`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L380-L389) |
| `unscan` | Return type `void` → `self` | [`strscan_unscan`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L1399-L1410) |
| `values_at` | Accept capture-group names (`*Integer` → `*Integer \| String \| Symbol`) | [`strscan_values_at`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L1837-L1853) |

### Pattern-matching methods

`check`, `check_until`, `exist?`, `match?`, `scan`, `scan_until`, `skip`, `skip_until`, `scan_full`, and `search_full` are not listed individually above because they all dispatch through the shared helper [`strscan_do_scan`](https://github.com/ruby/ruby/blob/v4.0.5/ext/strscan/strscan.c#L689-L755). That helper accepts both `Regexp` and `String` patterns, so the first parameter of each method is widened from `Regexp` to `Regexp | String`. It also produces their results, so the return types are corrected at the same time: `check_until` to `String?`, and `scan_full` / `search_full` from `untyped` to `(String | Integer)?`.

Removal of the obsolete methods (`clear`, `empty?`, `getbyte`, `peep`, `restsize`) is intentionally out of scope here and handled separately in #2961.

### DIFF
--- a/stdlib/strscan/0/string_scanner.rbs
+++ b/stdlib/strscan/0/string_scanner.rbs
@@ -498,7 +498,7 @@ class StringScanner
   # scanner[0]           # => nil
   # scanner[1]           # => nil
   #
-  def []: (Integer) -> String?
+  def []: (Integer | String | Symbol) -> String?
 
   # <!--
   #   rdoc-file=ext/strscan/strscan.c
--- a/stdlib/strscan/0/string_scanner.rbs
+++ b/stdlib/strscan/0/string_scanner.rbs
@@ -625,7 +625,7 @@ class StringScanner
   #     scanner.check(/nope/)          # => nil
   # match_values_cleared?(scanner) # => true
   #
-  def check: (Regexp) -> String?
+  def check: (Regexp | String) -> String?
 
   # <!--
   #   rdoc-file=ext/strscan/strscan.c
@@ -673,7 +673,7 @@ class StringScanner
   #     scanner.check_until(/nope/)    # => nil
   # match_values_cleared?(scanner) # => true
   #
-  def check_until: (Regexp) -> String
+  def check_until: (Regexp | String) -> String?
 
   # <!--
   #   rdoc-file=ext/strscan/strscan.c
@@ -760,7 +760,7 @@ class StringScanner
   #     scanner.exist?(/nope/)         # => nil
   # match_values_cleared?(scanner) # => true
   #
-  def exist?: (Regexp) -> Integer?
+  def exist?: (Regexp | String) -> Integer?
 
   # <!--
   #   rdoc-file=ext/strscan/strscan.c
@@ -913,7 +913,7 @@ class StringScanner
   #     scanner.match?(/nope/)         # => nil
   # match_values_cleared?(scanner) # => true
   #
-  def match?: (Regexp) -> Integer?
+  def match?: (Regexp | String) -> Integer?
 
   # <!--
   #   rdoc-file=ext/strscan/strscan.c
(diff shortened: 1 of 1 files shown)


stdlib/strscan/0/string_scanner.rbs
```diff
@@ -1288,7 +1331,22 @@ class StringScanner
   #
   # "full" means "#scan with full parameters".
   #
-  def scan_full: (Regexp pattern, bool advance_pointer_p, bool return_string_p) -> untyped
+  def scan_full: (Regexp | String pattern, bool advance_pointer_p, bool return_string_p) -> (String | Integer)?
```
ksss: `return_string_p` seems to be an option that changes the return type, as the name suggests.
Can we separate the overloads based on whether `return_string_p` is `true` or `false`?

stdlib/strscan/0/string_scanner.rbs
```diff
@@ -1352,7 +1410,7 @@ class StringScanner
   # `return_string_p` is true, otherwise returns the number of bytes advanced.
   # This method does affect the match register.
   #
-  def search_full: (Regexp pattern, bool advance_pointer_p, bool return_string_p) -> untyped
+  def search_full: (Regexp | String pattern, bool advance_pointer_p, bool return_string_p) -> (String | Integer)?
```
ksss: ditto

ksss (commented):
Once the conflicts are resolved, it should be ready to merge.

stdlib/strscan/0/string_scanner.rbs
zonuexe: Thanks for the review. I rebased it, resolved the conflicts, and added the overload.

ksss (approved):
✨

PR #2966, opened 2026-05-21
Replace deterministic `untyped` return types with concrete types

zonuexe:
## Summary

Several methods were typed `-> untyped` even though the implementation makes the return value deterministic. This replaces those with concrete types.

| Ruby Method | `untyped` → | Basis |
|-------------|-------------|-------|
| `Zlib::GzipFile#sync=` | `boolish` | setter returns its argument |
| `MonitorMixin::ConditionVariable#wait_for_cond` | `bool` | C implementation returns `Qtrue` / `Qfalse` |
| `MonitorMixin::ConditionVariable#wait` | `bool` | delegates to `#wait_for_cond` |
| `MonitorMixin::ConditionVariable#wait_until` | `nil` | body is an `until` loop expression |
| `MonitorMixin::ConditionVariable#wait_while` | `nil` | body is a `while` loop expression |
| `ERB#run` | `nil` | `print self.result(b)` — `print` returns `nil` |
| `ERB#initialize` | `void` | constructor |
| `ERB#def_method` | `Symbol` | `module_eval { eval("def …\nend") }` returns the method name |
| `ERB::DefMethod.def_erb_method` | `Symbol` | delegates to `ERB#def_method` |

### `putc`

`IO#putc`, `StringIO#putc`, and `Zlib::GzipWriter#putc` return their argument. They were typed `-> untyped` (`StringIO` / `GzipWriter`) or `(Numeric | String) -> (Numeric | String)` (`IO`); this types them as `(real) -> real | (String) -> String` so the return type tracks the argument. `real` is also more precise than `Numeric` — `putc` does not accept `Complex`. (`IO#putc` was not `untyped`, but is aligned here for consistency since the others document themselves as "Same as IO".)

## Out of scope

- `ObjectSpace.internal_super_of` — although the doc says it returns "Class or Module", it actually returns the internal super entry, which can be a `T_ICLASS` object. That is not a real `Module`, so `untyped` is the correct type (the `stdlib_test` for it confirms this). Left unchanged.
- The `core/class.rbs` (`allocate`, `new`) and `core/struct.rbs` (`self.new`) `untyped` → `instance` proposals were left out: `Class#new` is special-cased by the type checker, and `Struct.new(:a, :b)` returns a new struct *class* (not an instance), so `instance` would be incorrect there.

## Test plan

- [x] `bundle exec rbs -r stringio -r zlib -r monitor -r erb validate`
- [x] `bundle exec rake stdlib_test`
- [x] `bundle exec rubocop` on the changed files

### DIFF
--- a/stdlib/stringio/0/stringio.rbs
+++ b/stdlib/stringio/0/stringio.rbs
@@ -1271,7 +1271,8 @@ class StringIO
   # -->
   # See IO#putc.
   #
-  def putc: (Numeric | String arg0) -> (Numeric | String)
+  def putc: (real) -> real
+          | (String) -> String
 
   def puts: (*untyped arg0) -> nil
 
--- a/stdlib/stringio/0/stringio.rbs
+++ b/stdlib/stringio/0/stringio.rbs
@@ -1271,7 +1271,7 @@ class StringIO
   # -->
   # See IO#putc.
   #
-  def putc: (Numeric | String arg0) -> untyped
+  def putc: (Numeric | String arg0) -> (Numeric | String)
 
   def puts: (*untyped arg0) -> nil
 
--- a/stdlib/stringio/0/stringio.rbs
+++ b/stdlib/stringio/0/stringio.rbs
@@ -1272,7 +1272,8 @@ class StringIO
   # -->
   # See IO#putc.
   #
-  def putc: (Numeric | String arg0) -> (Numeric | String)
+  def putc: (real) -> real
+          | (String) -> String
 
   def puts: (*untyped arg0) -> nil
 
--- a/stdlib/stringio/0/stringio.rbs
+++ b/stdlib/stringio/0/stringio.rbs
@@ -1272,7 +1272,7 @@ class StringIO
   # -->
   # See IO#putc.
   #
-  def putc: (Numeric | String arg0) -> untyped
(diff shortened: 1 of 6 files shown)


stdlib/stringio/0/stringio.rbs
```diff
@@ -1271,7 +1271,7 @@ class StringIO
   # -->
   # See IO#putc.
   #
-  def putc: (Numeric | String arg0) -> untyped
+  def putc: (Numeric | String arg0) -> (Numeric | String)
```
ksss: How about to like this?


```suggestion
  def putc: (real) -> real
            (String) -> String
```

stdlib/zlib/0/gzip_writer.rbs
```diff
@@ -125,7 +125,7 @@ module Zlib
     # -->
     # Same as IO.
     #
-    def putc: (Numeric | String arg0) -> untyped
+    def putc: (Numeric | String arg0) -> (Numeric | String)
```
ksss: ditto

stdlib/stringio/0/stringio.rbs
zonuexe: I applied it to `IO` as well.

ksss (approved):

PR #2967, opened 2026-05-21
Correct core/stdlib signatures to match Ruby 4.0

zonuexe:
## Summary

Several core/stdlib signatures were inaccurate against the Ruby 4.0 implementation. This corrects them.

| Ruby Method | Change |
|-------------|--------|
| `IO#pathconf` | Add the method — `(Integer name) -> Integer?` |
| `Digest::Instance#finish` | Return `String`, not `self` |
| `Digest::Class.digest` / `.hexdigest` / `.bubblebabble` | Accept trailing `*args` (passed through to the constructor) |
| `Digest::Class.file` / `.base64digest` | Accept trailing `*args` |
| `Etc.sysconf` | Return type is nilable — `-> Integer?` |
| `Etc.systmpdir` | Return type is nilable — `-> String?` |
| `Etc::Group.each` / `Etc::Passwd.each` | Add block / `Enumerator` overloads instead of `-> untyped` |
| `Etc::Passwd` | Add the `age`, `comment`, and `quota` members (with their setters) |
| `StringIO.open` | Add the non-block overload — `(?String, ?String?) -> StringIO` |
| `StringIO#fcntl` | Always raises `NotImplementedError` — return `bot` |
| `StringIO#fsync` | Always returns `0` — `-> Integer` (not `Integer?`) |
| `StringIO#internal_encoding` | Always returns `nil` |
| `StringIO#external_encoding` | Return type is nilable — `-> Encoding?` |
| `StringIO#pread` | Add the method |
| `StringIO#read_nonblock` | Accept `exception:`; return type is nilable |
| `StringIO#write_nonblock` | Add the method |
| `StringIO#sysread` | `outbuf` is optional |
| `StringIO#ungetc` | Also accepts an `Integer` |
| `StringIO#set_encoding_by_bom` | Add the method |

## Notes

- `Monitor#wait_for_cond` / `MonitorMixin::ConditionVariable#wait` (`-> bool`) are intentionally excluded here — they are covered by #2966.
- rdoc comments for the newly added methods match `rbs annotate` output (`confirm_annotation`).

## Test plan

- [x] `bundle exec rbs -r digest -r etc -r stringio validate`
- [x] `bundle exec rubocop core/io.rbs stdlib/digest/0/digest.rbs stdlib/etc/0/etc.rbs stdlib/stringio/0/stringio.rbs`

### DIFF
--- a/stdlib/digest/0/digest.rbs
+++ b/stdlib/digest/0/digest.rbs
@@ -303,7 +303,7 @@ module Digest::Instance
   # uninitialized.  Do not call this method from outside.  Use #digest!() instead,
   # which ensures that internal data be reset for security reasons.
   #
-  def finish: () -> self
+  def finish: () -> String
 end
 
 # <!-- rdoc-file=ext/digest/digest.c -->
@@ -319,7 +319,7 @@ class Digest::Class
   # Returns the base64 encoded hash value of a given *string*.  The return value
   # is properly padded with '=' and contains no line feeds.
   #
-  def self.base64digest: (string str) -> String
+  def self.base64digest: (string str, *untyped args) -> String
 
   # <!--
   #   rdoc-file=ext/digest/bubblebabble/bubblebabble.c
@@ -327,7 +327,7 @@ class Digest::Class
   # -->
   # Returns the BubbleBabble encoded hash value of a given *string*.
   #
-  def self.bubblebabble: (string) -> String
+  def self.bubblebabble: (string, *untyped) -> String
 
   # <!--
   #   rdoc-file=ext/digest/digest.c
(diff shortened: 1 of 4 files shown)


ksss (approved):

stdlib/digest/0/digest.rbs
```diff
@@ -319,15 +319,15 @@ class Digest::Class
   # Returns the base64 encoded hash value of a given *string*.  The return value
   # is properly padded with '=' and contains no line feeds.
   #
-  def self.base64digest: (string str) -> String
+  def self.base64digest: (string str, *untyped args) -> String
```
ksss, line 322: I have already merged it, but I believe there is an issue with these changes.
For example, the following code will cause an ArgumentError in actual behavior, but it passes the type check.

```rb
Digest::SHA256.base64digest("aaa", :bbb) #=> ArgumentError
```

While these methods are indeed delegated to the constructor, since it is not possible to describe this delegation in RBS, I believe we should add appropriate overloads in each class.

I will fix it soon.

We will address the issue that the existing test cases failed to detect the problem as a problem with the test harness.

ksss: Done https://github.com/ruby/rbs/pull/2980

PR #2973, opened 2026-05-25
Fix lexer infinite loop / abort on invalid UTF-8 byte

ksss:
## Summary

`rbs_next_char` left `byte_len = 0` when the active encoding's `char_width()` rejected a byte. Depending on where the byte appeared, the lexer either:

- looped forever inside a comment (GVL held, `SIGINT` ineffective), or
- tripped `RBS_ASSERT(current_character_bytes > 0, ...)` in `rbs_skip` at top level and called `exit(1)`.

This PR makes `rbs_next_char` advance one byte in that case so the lexer always makes progress; the invalid byte then flows into the usual parser error path.

## Reproducer

```ruby
buf = RBS::Buffer.new(content: "# \xC2".force_encoding("UTF-8"), name: "x.rbs")
RBS::Parser._parse_signature(buf, 0, buf.content.bytesize)
# hangs forever; only `kill -9` stops the process
```

## Fix

`src/lexstate.c` (`rbs_next_char`): when `encoding->char_width()` returns `0`, treat the byte as a 1-byte garbage character and advance one byte. The lexer's invariant "cursor advances by at least one byte per step" is now preserved on every code path. Valid UTF-8 input is unaffected because `char_width()` never returns `0` for valid sequences.

---

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,6 +48,7 @@ gem "net-smtp"
 gem 'csv'
 gem 'ostruct'
 gem 'pstore'
+gem "timeout"
 
 group :minitest do
   gem "minitest"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -231,6 +231,7 @@ DEPENDENCIES
   steep!
   tempfile
   test-unit
+  timeout
 
 BUNDLED WITH
   4.0.1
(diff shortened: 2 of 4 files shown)


soutaro (approved):
It would be better if we can report an error when invalid token is included in the comment, but I cannot figure out how to do that right now...

ksss:
@soutaro I had claude consider how to handle this.
https://github.com/ruby/rbs/pull/2983

PR #2974, opened 2026-05-25
Reject reversed position ranges in parser/lexer entrypoints

ksss:
## Summary

`alloc_parser_from_buffer` and `alloc_lexer_from_buffer` already rejected negative positions but accepted ranges where `start_pos > end_pos`. Such a range made the lexer's main loop condition (`current.byte_pos == end_pos`) unsatisfiable: `current` advances past `end_pos` and the equality is never reached, producing an infinite loop in the C extension with the GVL held (`SIGINT` ineffective).

## Reproducer

```ruby
RBS::Parser.parse_type("", byte_range: 1..0)
# hangs forever; only `kill -9` stops the process
```

The same hang is reachable from `RBS::Parser.parse_method_type` via its `byte_range:` keyword, and from the low-level `_parse_signature` / `_parse_type` / `_parse_method_type` directly with reversed `start_pos` / `end_pos` arguments. The `_lex` entrypoint always passes `start_pos = 0`, so the new guard on the lexer side is defensive rather than a present-day bug fix.

## Fix

Add a reversed-range guard alongside the existing negative-position check, and extract both into a `validate_position_range` helper shared by `alloc_lexer_from_buffer` and `alloc_parser_from_buffer`:

```c
static void validate_position_range(int start_pos, int end_pos) {
    if (start_pos < 0 || end_pos < 0) {
        rb_raise(rb_eArgError, "negative position range: %d...%d", start_pos, end_pos);
    }
    if (start_pos > end_pos) {
        rb_raise(rb_eArgError, "invalid position range: %d...%d", start_pos, end_pos);
    }
}
```

Reversed ranges now raise `ArgumentError` immediately instead of reaching the lexer.

---

### DIFF
--- a/test/rbs/parser_test.rb
+++ b/test/rbs/parser_test.rb
@@ -1,5 +1,4 @@
 require "test_helper"
-require "timeout"
 
 class RBS::ParserTest < Test::Unit::TestCase
   def buffer(source)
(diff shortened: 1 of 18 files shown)


test/rbs/parser_test.rb
```diff
@@ -1028,4 +1029,23 @@ class Foo[T < Integer] < Bar # Comment
     assert_equal [:tTRIVIA, "\n", 56...57], tokens.shift.then { |t| [t[0], t[1].source, t[1].range] }
     assert_equal [:pEOF, '', 57...57], tokens.shift.then { |t| [t[0], t[1].source, t[1].range] }
   end
+
+  def test_invalid_position_range_raises
+    # Regression: start_pos > end_pos used to cause an infinite loop in the lexer.
+    Timeout.timeout(5) do
```
soutaro: Does this `Timeout.timeout` call make any sense? It cannot interrupt loop in C, and then it anyways cannot leave from the loop without the fix.

soutaro (approved):

test/rbs/parser_test.rb
ksss: Exactly. I will remove the timeout.

Issue #2954, opened 2026-05-14
Proposal: Extend `ruby-rbs` crate features


dak2:
## Motivation

The `ruby-rbs` crate currently provides only the parser layer (AST + `Visit` trait).
The `rbs` gem, in contrast, ships several higher-level layers on top of the parser — environment construction, name resolution, and definition (ancestor / method) resolution.

Until those layers exist on the Rust side, downstream tooling (type checkers, LSP servers, code generators) cannot be built using `ruby-rbs` alone. This issue proposes to gradually port those upper layers into the crate.

## Features I want

I'm building a static type checker in Rust for my own gem.
Currently, RBS type definitions are loaded on the Ruby side and the result is then handed over to Rust for analysis.
I'd like to move the loading step into Rust as well, so the whole pipeline can be driven through a Rust interface — which requires the following layers in `ruby-rbs`:

- **`TypeName` / `Namespace`** — structured type names like `::Foo::Bar`; prerequisite for every layer below
- **`EnvironmentLoader`** — discovering `.rbs` files from gem paths / libdirs, equivalent to `rbs` gem's loader
- **`Environment`** — merging declarations from multiple files (class reopens, module aliases) and exposing a lookup API
- **`DefinitionBuilder`** — resolving inherited methods via MRO so callers can ask "what methods does `Foo::Bar` expose?" without re-implementing ancestor traversal in every consumer

Of course, things won’t always go as planned.

## Willingness to contribute

If the RBS core team is planning to expand `ruby-rbs` with these upper layers, I'd love to contribute actively — not just file this as a wishlist.

soutaro:
https://github.com/ruby/rbs/pull/2964 I tried to implement some basic infrastructure -- string interner and typename definition.

dak2:
Thanks for getting right to work on the prototype! I’ve also created a draft version of NameSpace. 
https://github.com/ruby/rbs/pull/2975

I’d like to incorporate this as well as a first step, if possible.

PR #2975, opened 2026-05-25
[Rust] Add NameSpace

dak2:
https://github.com/ruby/rbs/issues/2954


dak2:
After talking with Soutaro, he said that since NameSpace was originally scheduled for removal, it might not be included in the crate either.

PR #2987, opened 2026-06-03
Update `Array`

sampersand:
This PR completely updates `Array`'s signatures, and rewrites the test harness to work with it.

Note that there's some `# where E: ...` comments strewn about in the code. This marks places where, when we eventually get around to implementing the `where` clause, we should implement them in the Array.

### DIFF
--- a/core/array.rbs
+++ b/core/array.rbs
@@ -1397,7 +1397,7 @@ class Array[unchecked out E]
   #
   # Related: see [Methods for Assigning](rdoc-ref:Array@Methods+for+Assigning).
   #
-  def concat: (*::Array[E] arrays) -> self
+  def concat: (*array[E] other_arrays) -> self
 
   # <!--
   #   rdoc-file=array.c
@@ -2750,7 +2750,7 @@ class Array[unchecked out E]
   # Deleting](rdoc-ref:Array@Methods+for+Deleting).
   #
   def pop: () -> E?
-         | (int n) -> ::Array[E]
+         | (int count) -> Array[E]
 
   # <!-- rdoc-file=array.c -->
   # Prepends the given `objects` to `self`:
@@ -2838,7 +2838,7 @@ class Array[unchecked out E]
   #
   # Related: see [Methods for Assigning](rdoc-ref:Array@Methods+for+Assigning).
   #
-  def push: (*E obj) -> self
+  def push: (*E objects) -> self
 
   # <!--
   #   rdoc-file=array.c
@@ -3001,7 +3001,7 @@ class Array[unchecked out E]
   #
   # Related: see [Methods for Combining](rdoc-ref:Array@Methods+for+Combining).
   #
-  def reverse: () -> ::Array[E]
+  def reverse: () -> Array[E]
 
   # <!--
   #   rdoc-file=array.c
@@ -3015,7 +3015,7 @@ class Array[unchecked out E]
   #
   # Related: see [Methods for Assigning](rdoc-ref:Array@Methods+for+Assigning).
   #
-  def reverse!: () -> ::Array[E]
+  def reverse!: () -> self
 
   # <!--
   #   rdoc-file=array.c
@@ -3133,7 +3133,7 @@ class Array[unchecked out E]
   #
   # Related: see [Methods for Fetching](rdoc-ref:Array@Methods+for+Fetching).
   #
-  def rotate: (?int count) -> ::Array[E]
+  def rotate: (?int count) -> Array[E]
 
   # <!--
   #   rdoc-file=array.c
@@ -3305,7 +3305,7 @@ class Array[unchecked out E]
   # Related: see [Methods for Deleting](rdoc-ref:Array@Methods+for+Deleting).
   #
   def shift: %a{implicitly-returns-nil} () -> E
-           | (int n) -> ::Array[E]
+           | (int count) -> Array[E]
 
   # <!--
   #   rdoc-file=array.rb
@@ -3746,7 +3746,7 @@ class Array[unchecked out E]
   #
   # Related: see [Methods for Converting](rdoc-ref:Array@Methods+for+Converting).
   #
-  def to_a: () -> ::Array[E]
+  def to_a: () -> Array[E]
 
   # <!--
   #   rdoc-file=array.c
@@ -3902,7 +3902,7 @@ class Array[unchecked out E]
   # Related: Array#shift; see also [Methods for
   # Assigning](rdoc-ref:Array@Methods+for+Assigning).
   #
-  def unshift: (*E obj) -> self
+  def unshift: (*E objects) -> self
 
   # <!--
   #   rdoc-file=array.c
@@ -4124,7 +4124,7 @@ class Array[unchecked out E]
   #
   def |: [T] (array[T] other_array) -> Array[E | T] # TODO: where E: _Eql[T]
 
-  def freeze: ->__todo__
+  def freeze: () -> self
   alias detect find
 
   private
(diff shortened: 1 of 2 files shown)


core/array.rbs
```diff
@@ -1155,7 +1161,9 @@ class Array[unchecked out E] < Object
   #
   # Related: see [Methods for Querying](rdoc-ref:Array@Methods+for+Querying).
   #
-  alias any? all?
+  def any?: () -> bool
+          | (_Pattern[E] pattern) -> bool
+          | () { (E element) -> boolish } -> bool
```
sampersand, line 1166: This was probably originally `alias`ed because they share the same signature? but they shouldn't have been

core/array.rbs
```diff
@@ -1155,7 +1161,9 @@ class Array[unchecked out E] < Object
   #
   # Related: see [Methods for Querying](rdoc-ref:Array@Methods+for+Querying).
   #
-  alias any? all?
+  def any?: () -> bool
+          | (_Pattern[E] pattern) -> bool
```
sampersand: TODO: Use `RBS::Ops:_Matches`

core/array.rbs
```diff
@@ -2215,7 +2219,7 @@ class Array[unchecked out E] < Object
   #
   # Related: see [Methods for Querying](rdoc-ref:Array@Methods+for+Querying).
   #
-  def include?: (top object) -> bool
+  def include?: (top object) -> bool # TODO: E: _Equals
```
sampersand: Should this be `top` or `untyped`? And should other places be using `top`?

core/array.rbs
```diff
@@ -2502,10 +2506,8 @@ class Array[unchecked out E] < Object
   #
   # Related: see [Methods for Fetching](rdoc-ref:Array@Methods+for+Fetching).
   #
-  def max: %a{implicitly-returns-nil} () -> E
-         | %a{implicitly-returns-nil} () { (E a, E b) -> ::Integer? } -> E
-         | (int n) -> ::Array[E]
-         | (int n) { (E a, E b) -> ::Integer? } -> ::Array[E]
+  def max: %a{implicitly-returns-nil} () ?{ (E a, E b) -> Comparable::_CompareToZero? } -> E # TODO: where `E: Comparable::_WithSpaceshipOperator`
+         | (int count) ?{ (E a, E b) -> Comparable::_CompareToZero? } -> Array[E] # TODO: i dont think these shoudl have `?` in _CompareToZero. But it would make signatures nicer?
```
sampersand: TODO: Should we have `?` here in the signature? Actually having it will always cause an exception, but most `<=>` have `nil` in their return type.

core/array.rbs
```diff
@@ -3292,7 +3302,7 @@ class Array[unchecked out E] < Object
   # Related: see [Methods for Deleting](rdoc-ref:Array@Methods+for+Deleting).
   #
   def shift: %a{implicitly-returns-nil} () -> E
```
sampersand, line 3351: It feels weird to have this be an `implicitly-returns-nil`

core/array.rbs
```diff
@@ -3676,8 +3682,7 @@ class Array[unchecked out E] < Object
   # *   Array#sum method may not respect method redefinition of "+" methods such
   #     as Integer#+.
   #
-  def sum: (?untyped init) -> untyped
-         | (?untyped init) { (E e) -> untyped } -> untyped
+  def sum: (?untyped init) ?{ (E e) -> untyped } -> untyped # This would be usable if we had `where E < Complex`, etc
```
sampersand, line 3737: I wish there was a better way to solve `sum`

core/array.rbs
```diff
@@ -4130,11 +4140,12 @@ class Array[unchecked out E] < Object
...
+  alias initialize_copy replace
 end
 
+%a{deprecated: Use Array::_Rand}
 interface _Rand
-  def rand: (::Integer max) -> ::Integer
+  def rand: (Integer max) -> Integer
 end
 
 interface Array::_Pattern[T]
```
sampersand, line 4203: TODO: deprecate this and just use `RBS::Ops` when needed

core/array.rbs
```diff
@@ -1225,9 +1233,8 @@ class Array[unchecked out E] < Object
   #
   # Related: see [Methods for Fetching](rdoc-ref:Array@Methods+for+Fetching).
   #
-  def bsearch: () -> ::Enumerator[E, E?]
-             | () { (E) -> (true | false) } -> E?
-             | () { (E) -> ::Integer } -> E?
+  def bsearch: () { (E element) -> (Numeric | bool?) } -> E?
```
soutaro: I'm not sure if this works for actual type checkers, but these definitions are not equivalent.
Returning both `true | false` and `Integer` from a block is *not* allowed to `bsearch`, I think.

core/array.rbs
```diff
@@ -3607,8 +3649,7 @@ class Array[unchecked out E] < Object
   #
   # Related: see [Methods for Assigning](rdoc-ref:Array@Methods+for+Assigning).
   #
-  def sort!: () -> self
-           | () { (E a, E b) -> ::Integer } -> self
+  def sort!: () ?{ (E a, E b) -> Comparable::_CompareToZero } -> Array[E]
```
soutaro: Does it return `self` instead of `Array[E]`?

core/array.rbs
sampersand: yep, do this

core/array.rbs
ParadoxV5:
```ruby
[…].bsearch { it.positive? and it } # false | Integer
```

sampersand (commented):
Did final check, should good

core/array.rbs
```diff
@@ -621,7 +627,7 @@ class Array[unchecked out E] < Object
   # Related: see [Methods for Creating an
   # Array](rdoc-ref:Array@Methods+for+Creating+an+Array).
   #
-  def self.[]: [U] (*U) -> ::Array[U]
+  def self.[]: [U] (*U) -> Array[U] # this technically returns a subclass, but `instance(self)` isn't a thing
```
soutaro, line 630: I think `instance` would work. However `instance[U]` isn't a thing.

core/array.rbs
```diff
@@ -640,7 +646,9 @@ class Array[unchecked out E] < Object
   # Related: see [Methods for Creating an
   # Array](rdoc-ref:Array@Methods+for+Creating+an+Array).
   #
-  def self.try_convert: [U] (untyped) -> ::Array[U]?
+  def self.try_convert: [A < Array[U], U] (A array) -> A
+                      | [U] (array[U] ary) -> Array[U]
+                      | [U] (untyped) -> Array[U]?
```
soutaro:
```suggestion
                      | (untyped) -> Array[untyped]?
```

soutaro: Wow, the original definition was broken...

core/array.rbs
soutaro: https://docs.ruby-lang.org/en/4.0/language/bsearch_rdoc.html#label-Binary+Searching

> The block should not mix the modes by sometimes returning true or false and other times returning a numeric value, but this is not checked.

core/array.rbs
```diff
@@ -640,7 +646,9 @@ class Array[unchecked out E] < Object
   # Related: see [Methods for Creating an
   # Array](rdoc-ref:Array@Methods+for+Creating+an+Array).
   #
-  def self.try_convert: [U] (untyped) -> ::Array[U]?
+  def self.try_convert: [A < Array[U], U] (A array) -> A
+                      | [U] (array[U] ary) -> Array[U]
```
soutaro:
```suggestion
                      | [U] (array[U] ary) -> Array[U]?
```

soutaro (approved):
👍

PR #2988, opened 2026-06-03
Add the protected visibility

sampersand:


soutaro:
Can you confirm (and explain) that this PR provides sufficient APIs/information for implementing type checking for them?

Issue #2943, opened 2026-05-01
Add `wasm32-wasip1` target support for `ruby-rbs-sys` (cross-platform bindgen issues)


rubys:
## Summary

`ruby-rbs-sys` 0.3.0's `build.rs` has no `wasm32-wasi*` target awareness — there's no way to point `cc` at a wasi-targeting compiler or propagate the same flags to bindgen. I drafted a build refactor mirroring [`ruby-prism-sys`'s pattern](https://github.com/ruby/prism/blob/main/rust/ruby-prism-sys/build/vendored.rs) (split `build.rs` into `build/main.rs` + `build/vendored.rs`, detect the wasm target, route everything through wasi-sdk). The C compile of vendored librbs to `wasm32-wasip1` then succeeds cleanly. **However, the bindgen step fails differently across three different libclang implementations**, and the failure modes suggest a deeper bindgen-on-wasm issue that may also affect `ruby-prism-sys` — but isn't surfacing there because [its wasm CI doesn't actually run](#why-prisms-pattern-alone-may-not-be-a-sufficient-precedent).

This issue documents what I found in case maintainers (or anyone with deeper bindgen-on-wasm experience) can spot the underlying cause. I'm happy to send a PR for the build-refactor portion regardless — that part is mechanical and parallels prism cleanly. The bindgen question is what I'd like guidance on before proceeding.

## Goal

Build `ruby-rbs` (and consumer crates) for `wasm32-wasip1` so it can run in browser environments via `wasi-libc` shims. Used in a downstream project that ingests RBS signatures alongside Ruby source for a transpile-time type-resolution pipeline.

## What works after the build refactor

- Vendored librbs C compiles to `wasm32-wasip1` cleanly with `WASI_SDK_PATH` set, using `-D_WASI_EMULATED_MMAN=1` (defensive `<sys/mman.h>` includes in `rbs_allocator.c`; no actual mmap calls, so no link against `libwasi-emulated-mman` needed).
- Native build path is unaffected; the wasm code is gated on `target.contains("wasm32") && target.contains("wasi")`.
- Bindgen is invoked with the same `-I`, `-D`, and `--sysroot` flags as `cc` via `BINDGEN_EXTRA_CLANG_ARGS_<target>` propagation — same convention `ruby-prism-sys` uses.

## Three libclangs, three failure modes

| Platform / libclang | Failure |
|---|---|
[...]

PR #2992, opened 2026-06-14
ruby-rbs-sys: support wasm32 targets

rubys:
## What

Teach `ruby-rbs-sys`'s build script to support `wasm32-*` targets:

- Compile the vendored RBS C parser with the WASI SDK's clang (via `WASI_SDK_PATH`).
- Generate the FFI bindings against the **host** target (`--target=$HOST`, `layout_tests(false)`) rather than the wasm target.

This mirrors `ruby-prism-sys`'s existing wasm build support.

## Why

Running bindgen against the wasm target is libclang-version-fragile. Across libclangs I see two distinct failures, both on `wasm32-wasip1`:

- some libclangs drop *all* `rbs_*` function declarations;
- others emit the forward-referenced node structs (`rbs_namespace`, `rbs_type_name`, `rbs_ast_symbol`, `rbs_types_block`, `rbs_ast_declarations_class_super`) as opaque, so the generated layout assertions overflow (`E0080`).

The generated `#[repr(C)]` declarations are layout-portable, so generating them for the host and compiling them for `wasm32` is both correct and reliable. The C parser is still compiled for wasm by the WASI SDK clang.

## Testing

- `cargo build -p ruby-rbs-sys --target wasm32-wasip1` and `cargo build -p ruby-rbs --target wasm32-wasip1` both succeed.
- Runtime: a cdylib calling `rbs_parser_new` → `rbs_parse_signature` → walking `signature.declarations` (and the high-level `ruby_rbs::node::parse`) runs correctly under a WASI runtime.
- Native builds are unchanged — the wasm-specific configuration is gated on the target.

### DIFF
--- a/rust/ruby-rbs-sys/build.rs
+++ b/rust/ruby-rbs-sys/build.rs
@@ -11,21 +11,45 @@ fn main() -> Result<(), Box<dyn Error>> {
     let include = vendor_rbs.join("include");
     let c_src = vendor_rbs.join("src");
 
-    build(&include, &c_src)?;
+    let target = env::var("TARGET").unwrap_or_default();
+    let is_wasm = target.contains("wasm32");
 
-    let bindings = generate_bindings(&include)?;
+    build(&include, &c_src, is_wasm)?;
+
+    let bindings = generate_bindings(&include, is_wasm)?;
     write_bindings(&bindings)?;
 
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


pvcresin:
+1, there's demand for using `ruby-rbs-sys` on `wasm32-wasip1`. Looking forward to this landing — thanks for the work here!

ref: https://github.com/ruby/rbs/issues/2943

PR #1080, opened 2022-08-08
added type of Float::INFINITY

HoneyryderChuck:

PR #2994, opened 2026-06-16
Remove stale Float constants and add a constant drift test

zonuexe:
## Summary

Removes two `Float` constants that are declared in `core/float.rbs` but raise `NameError` at runtime, and adds a test guarding against this kind of drift for platform-invariant core classes.

## Removed constants

- **`Float::ROUNDS`** — removed from Ruby in 3.0 (ruby/ruby#2953); accessing it raises `NameError`.
- **`Float::Infinity`** — added in #1095 (which superseded #1080). Ruby does not define this constant: looking at the Ruby of that period (3.1), `numeric.c` defines `Float::INFINITY` but not `Float::Infinity`, and it still raises `NameError` today.

## Constant drift test

`test/stdlib/constant_drift_test.rb` compares `Module#constants(false)` against the constants declared in the loaded RBS environment, for a curated set of platform/build-invariant core classes (`Float`, `Integer`, `String`, …). It fails in either direction:

- **stale** — declared in RBS but no longer defined at runtime (e.g. the two constants above), or
- **missing** — defined at runtime but absent from RBS.

Classes whose constant set varies by OS or build options (`Process`, `Socket`, `Errno`, …) are intentionally excluded, with a `SKIP` map for per-constant exceptions. Singleton constant type assertions for the remaining `Float` constants are added as well.


soutaro (approved):
Thanks!

PR #2995, opened 2026-06-16
Update `Integer`, phase 1

sampersand:
This updates `Integer`'s signatures. Since `Integer` is such an important class, I've broken it up into multiple phases.

This first phase does two things:
1. Modernized the test suite for `Integer_test.rb`. Note that since not all methods were implemented, test methods for missing ones are no-ops.
2. Updated signatures for all non-complicated, non-`coerce`-reliant methods.

### DIFF
--- a/core/integer.rbs
+++ b/core/integer.rbs
@@ -93,6 +93,8 @@
 #     given value.
 #
 class Integer < Numeric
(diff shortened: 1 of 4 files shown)


soutaro (approved):
🙏

PR #3001, opened 2026-06-16
Update `Integer`, phase 2

sampersand:
This is the second phase for updating `Integer` (phase one is #2995).

Here, we finish off the rest of the methods, including the ones which required `coerce`. The `_Coerce` interface (which is on `Numeric`) actually slots quite nicely:

```rbs
class Numeric
  interface _Coerce[O, Other, Self]
    def coerce: (O other) -> [Other, Self]
  end
end

class Integer < Numeric
  def +: (Integer other) -> Integer
       | [S, R] (Numeric::_Coerce[self, RBS::Ops::_Add[S, R], S] other) -> R
end
```

We don't even need separate cases for other numerics, as coerce works!

### DIFF
--- a/core/integer.rbs
+++ b/core/integer.rbs
@@ -93,6 +93,8 @@
 #     given value.
 #
 class Integer < Numeric
+  # TODO: `undef self.new`
+
   # <!--
   #   rdoc-file=numeric.c
   #   - Integer.sqrt(numeric) -> integer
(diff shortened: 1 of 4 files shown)


soutaro:
@sampersand I don't think it actually works with type checkers. Can you confirm with `rake typecheck_test`?

```rb
1 + 2     #=> Integer
1 + 2.0   #=> Float
1 + 2i    #=> (I don't remember)
1 + 2r    #=> (I don't remember)
1 + SomeCustomIntegerConvertible.new(2) #=> ??
```

PR #3002, opened 2026-06-17
Allow nil output buffers for reader methods

Strech:
Ruby accepts explicit `nil` for output buffer arguments on reader methods and allocates a new `String` in that case. Update signatures and stdlib tests to match.

Updated signatures:

* `_Reader#read`
* `_ReaderPartial#readpartial`
* `IO#{pread,read_nonblock,readpartial,sysread}`
* `ARGF#{read,read_nonblock,readpartial}`
* `StringIO#{read,pread,read_nonblock,readpartial,sysread}`
* `Zlib::GzipReader#read,readpartial`
* `OpenSSL::Buffering#{read,read_nonblock,readpartial}`
* `OpenSSL::SSL::SSLSocket#sysread`

### DIFF
--- a/core/builtin.rbs
+++ b/core/builtin.rbs
@@ -179,11 +179,11 @@ interface _EachEntry[out T]
 end
 
 interface _Reader
-  def read: (?int? length, ?string outbuf) -> String?
+  def read: (?int? length, ?string? outbuf) -> String?
 end
 
 interface _ReaderPartial
-  def readpartial: (int maxlen, ?string outbuf) -> String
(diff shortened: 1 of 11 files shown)


Strech:
@soutaro Not sure how can I make those 2 failing actions to be green as one of them require a label and another one is failing on bundle install with

```
An error occurred while installing strscan (3.1.8), and Bundler cannot continue.

In Gemfile:
  steep was resolved to 2.1.0.dev, which depends on
    strscan
```

soutaro (approved):
Thanks!

PR #3004, opened 2026-06-18
Fix flaky DirSingletonTest GC failures in test_fchdir and test_for_fd

hsbt:
The `Dir` created by `Dir.new(Dir.pwd).fileno` in `test_fchdir` and `test_for_fd` was only referenced long enough to read its fileno, so it became collectable right away. When GC ran before the descriptor was used it got closed, making the later `Dir.fchdir(fd)` and `Dir.for_fd(fd)` calls fail with `Errno::EBADF`.

This surfaced on `ruby/ruby` ZJIT CI running `make test-bundled-gems` with `--enable-zjit=dev --zjit-call-threshold=1`, where allocation and GC timing shifted enough to expose it. Keep the Dir in a local and close it in ensure so the descriptor stays valid for the whole test.

https://github.com/ruby/ruby/actions/runs/27735384476/job/82051097840

### DIFF
--- a/test/stdlib/Dir_test.rb
+++ b/test/stdlib/Dir_test.rb
@@ -104,7 +104,8 @@ class DirSingletonTest < Test::Unit::TestCase
   end
 
   def test_fchdir
-    fd = Dir.new(Dir.pwd).fileno
+    dir = Dir.new(Dir.pwd)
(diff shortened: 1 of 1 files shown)


soutaro (approved):
🙏

PR #3007, opened 2026-06-18
ci: clear MAKEFLAGS so the mswin build's nmake works

soutaro:
## Problem

The `compile (mswin)` job in **Ruby on Windows** has been failing on `master` since around Jun 16 (it passed on Jun 10). It dies during `bundle install` while building native gem extensions (`ffi`, `racc`, `zlib`, `strscan`, …):

```
current directory: D:/ruby-mswin/lib/ruby/gems/4.1.0+3/gems/ffi-1.17.4/ext/ffi_c
nmake sitearchdir\=./.gem.20260616-6852-l6mmqk sitelibdir\=./.gem.20260616-6852-l6mmqk clean

Microsoft (R) Program Maintenance Utility Version 14.51.36246.0
NMAKE : fatal error U1065: invalid option '-'
Stop.
```

Example run: https://github.com/ruby/rbs/actions/runs/27622377214/job/81674542835

## Root cause

Only the **mswin** matrix entry fails; `ucrt` and `3.4` pass. The difference is the make program: mswin uses MSVC's `nmake`, the others use GNU `make`.

The echoed command `nmake sitearchdir=... clean` contains no `-`, so the `-` that nmake rejects comes from an inherited **`MAKEFLAGS`** environment variable. GNU `make` accepts its own GNU-style `MAKEFLAGS`; `nmake` parses the leading `-` and aborts with `U1065`. This is the same `MAKEFLAGS` + `nmake` incompatibility documented in [rust-lang/cargo#4156](https://github.com/rust-lang/cargo/issues/4156).

`ruby/setup-ruby` sets `MAKE=nmake` for mswin but does **not** export `MAKEFLAGS`, so the value comes from the base environment and a workflow-level `env:` reliably overrides it. This is independent of any particular PR — `master` itself is affected.

## Fix

Clear `MAKEFLAGS` for the `compile` job so `nmake` receives no flags. It's a harmless no-op for the mingw/UCRT GNU make builds, and it also protects the `rake compile` step (which builds the RBS C extension via `nmake` on mswin).

```yaml
  compile:
    runs-on: "windows-latest"
    env:
      MAKEFLAGS: ""
```

https://claude.ai/code/session_014RENNXiMZ7ihuquCwKM4Ht

---
_Generated by [Claude Code](https://claude.ai/code/session_014RENNXiMZ7ihuquCwKM4Ht)_

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -13,6 +13,12 @@ permissions:
 jobs:
   compile:
     runs-on: "windows-latest"
+    env:
+      # The mswin (MSVC) build uses `nmake`, which aborts with
+      # "NMAKE : fatal error U1065: invalid option '-'" when it inherits a
+      # GNU-make-style MAKEFLAGS from the environment. Clear it so nmake gets no
+      # flags; this is a harmless no-op for the mingw/UCRT GNU make builds.
+      MAKEFLAGS: ""
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -13,12 +13,6 @@ permissions:
 jobs:
   compile:
     runs-on: "windows-latest"
-    env:
-      # The mswin (MSVC) build uses `nmake`, which aborts with
-      # "NMAKE : fatal error U1065: invalid option '-'" when it inherits a
-      # GNU-make-style MAKEFLAGS from the environment. Clear it so nmake gets no
-      # flags; this is a harmless no-op for the mingw/UCRT GNU make builds.
-      MAKEFLAGS: ""
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


soutaro:
Closing — this turned out to be an **upstream RubyGems bug that is already fixed**, so the workaround here is unnecessary.

**Root cause:** RubyGems' new *make jobserver* feature ([`b4d1813` "Implement a make jobserver"](https://github.com/rubygems/rubygems), 2026-06-16) injects `MAKEFLAGS` (with `-j`) into the `make` child process before building native extensions. On the mswin matrix entry `make` is `nmake`, which can't parse the GNU-style `-j` and aborts with `NMAKE : fatal error U1065: invalid option '-'`. (That's also why the value never showed up in the shell's `MAKEFLAGS` — RubyGems sets it on the child process, not the parent shell.)

**Already fixed upstream:** current `lib/rubygems/ext/builder.rb` special-cases nmake (`unless is_nmake`), so `nmake` is no longer given `-j`/jobserver flags. That fix has now propagated into the prebuilt mswin head Ruby.

**Verification:** re-running the previously-failed `compile (mswin)` job on `master` (commit a8f8f27, **no fix at all**) now passes — `bundle install` and `compile` both green (run `27738194928`, attempt 2). So `master` is green on its own and this workaround is not needed.

---
_Generated by [Claude Code](https://claude.ai/code)_

hsbt:
🙏

PR #3016, opened 2026-06-23
Deprecate RDoc plugin parser

st0012:
## Problems

- The dedicated parser tests are now break against the latest RDoc in Ruby core CI, preventing a RDoc version bump.
- However, removing the RBS-owned RDoc parser plugin immediately would break users who still depend on it before RDoc 8.0 is broadly available.

## Solutions

- Keep the plugin registration, parser implementation, and RBS signatures in place.
- Warn when the RBS plugin parser is actually used so users know RDoc 8.0 will provide this parsing directly.
- Remove only the old RDoc plugin parser test file to unblobk Ruby core CI.


soutaro:
@st0012 Thanks! Can you try using [rbs_skip_tests](https://github.com/ruby/ruby/blob/master/tool/rbs_skip_tests) file (and [_windows](https://github.com/ruby/ruby/blob/master/tool/rbs_skip_tests_windows))? to skip the failing tests in ruby CI?

PR #3019, opened 2026-06-27
Fetch Chicory/ASM jars from Maven instead of bundling in gem

### LINKED ISSUE #3018: Don't ship jars in the gem
JRuby-based gems that depend on other Java JAR files should ideally not ship those jars in the gem, for several reasons:

* The gem size will be greatly increased.
* Gems with conflicting copies of the same jars will cause errors if loaded at the same time.
* Jars should be sourced from the canonical source, Maven.

As part of the WASM work done by Claude in #2998, it appears that all of the related jars are now being shipped in the gem, under lib/rbs/wasm/jars. This adds almost 1MB of size to the released gem.

For this purpose, such gems use [jar-dependencies](https://github.com/jruby/jar-dependencies), included with JRuby, to fetch jars from Maven as part of the gem installation. This guarantees that all libraries use the same canonical source (Maven) and do not load conflicting versions (because everyone should go through jar-dependencies which is aware of those versions).

See [Psych](https://github.com/ruby/psych) for an example of a gem that uses jar-dependencies (in that case to fetch the jar for SnakeYAML Engine).

soutaro:
The `-java` gem bundled the Chicory/ASM jars (~1 MB) under `lib/rbs/wasm/jars`. This PR declares the dependencies with `jar-dependencies` requirements instead, so they're fetched from Maven Central at install time — the published gem stays small, everyone resolves the same versions from the canonical source, and conflicting copies can't be loaded at once.

Fixes [#3018](https://github.com/ruby/rbs/issues/3018).

### DIFF
--- a/Steepfile
+++ b/Steepfile
@@ -13,6 +13,8 @@ target :lib do
     "lib/rbs/wasm/location.rb",
     "lib/rbs/wasm/runtime.rb",
     "lib/rbs/wasm/parser.rb",
+    # jar-dependencies require_jar calls for the JRuby runtime; not type-checked.
+    "lib/rbs_jars.rb",
   )
 
   library "pathname", "json", "logger", "monitor", "tsort", "uri", 'dbm', 'pstore', 'singleton', 'shellwords', 'fileutils', 'find', 'digest', 'prettyprint', 'yaml', "psych", "securerandom"
--- a/lib/rbs_jars.rb
+++ b/lib/rbs_jars.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# This is the file jar-dependencies generates from the `jar` requirements in
+# rbs.gemspec, kept with hand edits:
+#   1. the `com.dylibso.chicory:runtime` require_jar line is corrected -- the
+#      generator mangles that artifact id to `jar` (it collides with a Maven
+#      scope keyword), which breaks loading;
+#   2. the "# this is a generated file" marker is dropped so jar-dependencies
+#      does not regenerate (and re-break) this file at gem install;
+#   3. a frozen_string_literal magic comment is added (rubocop).
+# `rake wasm:install_jars` downloads the jars into ~/.m2; to refresh the list
+# after a version bump, regenerate the file and re-apply these edits.
+begin
+  require 'jar_dependencies'
+rescue LoadError
+  require 'com/dylibso/chicory/compiler/1.7.5/compiler-1.7.5.jar'
+  require 'com/dylibso/chicory/runtime/1.7.5/runtime-1.7.5.jar'
+  require 'com/dylibso/chicory/wasm/1.7.5/wasm-1.7.5.jar'
+  require 'org/ow2/asm/asm/9.9.1/asm-9.9.1.jar'
+  require 'org/ow2/asm/asm-commons/9.9.1/asm-commons-9.9.1.jar'
+  require 'org/ow2/asm/asm-tree/9.9.1/asm-tree-9.9.1.jar'
+  require 'org/ow2/asm/asm-util/9.9.1/asm-util-9.9.1.jar'
+  require 'org/ow2/asm/asm-analysis/9.9.1/asm-analysis-9.9.1.jar'
+  require 'com/dylibso/chicory/wasi/1.7.5/wasi-1.7.5.jar'
+  require 'com/dylibso/chicory/log/1.7.5/log-1.7.5.jar'
+end
+
+if defined? Jars
+  require_jar 'com.dylibso.chicory', 'compiler', '1.7.5'
+  require_jar 'com.dylibso.chicory', 'runtime', '1.7.5'
+  require_jar 'com.dylibso.chicory', 'wasm', '1.7.5'
+  require_jar 'org.ow2.asm', 'asm', '9.9.1'
+  require_jar 'org.ow2.asm', 'asm-commons', '9.9.1'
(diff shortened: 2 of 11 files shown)


lib/rbs_jars.rb
```diff
@@ -0,0 +1,21 @@
...
+# overwriting this file at gem install. Keep this list in sync with the `jar`
+# requirements in rbs.gemspec.
+require "jar_dependencies"
+
+require_jar "com.dylibso.chicory", "wasm", "1.7.5"
+require_jar "com.dylibso.chicory", "runtime", "1.7.5"
+require_jar "com.dylibso.chicory", "log", "1.7.5"
+require_jar "com.dylibso.chicory", "wasi", "1.7.5"
+require_jar "com.dylibso.chicory", "compiler", "1.7.5"
+require_jar "org.ow2.asm", "asm", "9.9.1"
```
headius: Only top-level dependencies are required here; jar-dependencies will load any transitive dependencies necessary. I believe only `runtime` and `compiler` are needed from experimenting locally.

rbs.gemspec
```diff
@@ -38,18 +38,37 @@ Gem::Specification.new do |spec|
...
+    # jar-dependencies (bundled with JRuby) downloads these jars from Maven when
+    # the gem is installed, keeping the gem small and avoiding conflicting copies.
+    # lib/rbs_jars.rb require_jars them at runtime; keep the two lists in sync.
+    spec.add_dependency "jar-dependencies", ">= 0.1.7"
+    spec.requirements << "jar com.dylibso.chicory:wasm, 1.7.5"
+    spec.requirements << "jar com.dylibso.chicory:runtime, 1.7.5"
+    spec.requirements << "jar com.dylibso.chicory:log, 1.7.5"
+    spec.requirements << "jar com.dylibso.chicory:wasi, 1.7.5"
+    spec.requirements << "jar com.dylibso.chicory:compiler, 1.7.5"
+    spec.requirements << "jar org.ow2.asm:asm, 9.9.1"
```
headius: Only need `runtime` and `compiler` here as well I believe.

These dependencies are also being duplicated multiple places despite Claude saying that the jars file is the master list.

Steepfile
```diff
@@ -13,6 +13,8 @@ target :lib do
     "lib/rbs/wasm/location.rb",
     "lib/rbs/wasm/runtime.rb",
     "lib/rbs/wasm/parser.rb",
+    # Generated by `rake wasm:install_jars` (jar-dependencies require_jar calls).
```
headius: This documentation says it is generated, but the comments in that file say "Hand-maintained, NOT auto-generated". Claude is getting confused here.

headius (commented):
This is basically correct but a mess of conflicting comments from the LLM.

* Only the direct dependency jars need to be specified. I believe that's just `runtime` and `compiler`.
* Jar specifications are duplicated several places even though the comments claim they should live in the jars source file.
* rbs_jars.rb says it is hand-maintained but other places say it is generated. LLM is getting confused and contradicting itself.

This matches the work I did locally as an experiment, but quite a bit messier with nonsense comments.

headius:
LLM is making up nonsense, or getting confused...

> Renamed jars_dir to local_jars_dir and made it return nil for installed gems (only returns a path when jars are vendored locally for development)

I think in all scenarios jars are downloaded to local .m2 for use, so I'm not sure what this means.

> The jars.rb module serves as the single source of truth for jar coordinates, consumed by:

This file does not exist. The actual file, rbs_jars.rb, does seem to be the single source of truth... except its contents get duplicated several other places.

> Added graceful error handling for optional AOT compiler jars

Probably no need for this. The WASM interpreter will probably be unusably slow without the compiler, and there's no reason it won't load now.

--

The actual changes to support jar-dependencies look fine but the documentation has a lot of nonsense in it.

ParadoxV5:
Has this bundled gem degenerated from methodical development to prompt engineering and botsitting?

If the bot cannot do it right, throw its slop out and do it ourselves, and consult human experts rather than a black box of probabilities.

Relying on bot output is not “hand-maintained”; it’s _vibed_.
And vibe coding is already well infamous for its poor quality and inäbility to learn post-training.

rbs.gemspec
soutaro: Thanks! I keep `runtime` and `compiler` here, and delete everything else.

lib/rbs_jars.rb
soutaro: Requiring only the top-level jars resulted in `NoClassDefFoundError`. `jar-dependencies` gem also generates the `require_jar` lines. So, I think we need these lines.

soutaro:
@ParadoxV5

Fair concern — this PR was a complicated and confused draft, so I understand the reaction.

Either way, I'm responsible for the code that lands here, whether I write it myself, an LLM does, or another contributor does. And I'll be honest: I have almost no JRuby context, and I don't think I could have gotten this one done without the AI's help.

The history is squashed and the misleading description is being rewritten. Thanks for keeping the quality bar high.

lib/rbs_jars.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# This is the file jar-dependencies generates from the `jar` requirements in
+# rbs.gemspec, kept with hand edits:
+#   1. the `com.dylibso.chicory:runtime` require_jar line is corrected -- the
```
soutaro, line 5: Opened a PR to fix this: https://github.com/jruby/jar-dependencies/pull/110

PR #3028, opened 2026-07-21
Allow non-ASCII characters in identifiers

ksss:
## Summary

Ruby accepts identifiers with non-ASCII code points: `Ú` and `Ω` are valid constants, `únicos` and `日本語` are valid local identifiers. The RBS lexer only recognised ASCII, so the same names could not be written in a signature file. See [rubocop-on-rbs#151](https://github.com/ksss/rubocop-on-rbs/issues/151).

## Lexer

`rbs_next_char` already folded every non-ASCII character into a single sentinel code point. It now reports **one of two**, chosen by the encoding's `isupper_char` — the test Ruby's `rb_sym_constant_char_p` uses to tell a constant from a local — and the grammar matches those two classes directly:

```re2c
mb_upper = [ル];
mb_other = [ビ];
word     = [a-zA-Z0-9_] | mb_upper | mb_other;

([a-z] | mb_other) word* { return rbs_next_token(lexer, tLIDENT); }
([A-Z] | mb_upper) word* { return rbs_next_token(lexer, tUIDENT); }
```

Widening `word` also lets every existing rule accept non-ASCII **continuation** characters.

Deciding this in `rbs_next_char` rather than in a lexer action covers encodings where a non-ASCII character is a single byte: in ISO-8859-1, `À` is 0xC0 and `à` is 0xE0, and the old `byte_len == 1` branch passed both through as raw bytes that matched no identifier rule. It costs one `isupper_char` call per non-ASCII character rather than one per identifier — over the bundled core and stdlib that is below the run-to-run noise floor, and even a corpus that is 46% Japanese prose is only ~12% slower.

**Builds on #2983.** The two sentinels are distinct from the invalid-byte sentinel, so an invalid byte inside an identifier still surfaces as an ErrorToken rather than being absorbed into the token.

**ASCII-8BIT needs an exception.** It carries bytes rather than characters and its `char_width` reports 1 for every byte, so folding "every valid non-ASCII character" would turn arbitrary binary into identifiers.

**Depends on #3040.** A character straddling the end of a `byte_range` makes the lexer read on to the end of the string. That is a `master` bug — reachable today whenever the straddling character sits somewhere non-ASCII is already accepted, such as a string literal — and it is fixed separately in #3040. Letting identifiers hold non-ASCII widens the set of inputs that reach it, so this should land after that one.

## `TypeName#kind`

It answers the same question for names that never pass through the lexer (`TypeName.parse`), so it has to agree. It now asks Prism — already a hard dependency — instead of restating Ruby's rule with regexps:

```ruby
@kind = case
        when name.start_with?("_")
          :interface
        when Prism::StringQuery.constant?(name.name)
          :class
        else
          :alias
        end
```

That covers what a regexp got wrong here: titlecase letters (`class ǅFoo` used to parse into a `Declarations::Class` whose own `name.class?` was `false`, then raise from `Environment`), and names read from a non-UTF-8 signature (`\p{...}` pins a regexp to UTF-8 and raises). `Symbol#name` returns a frozen String, so unlike `to_s` this allocates nothing. Only the leading `_` case stays, since that is RBS's own interface convention. `rust/ruby-rbs` has no Prism to defer to and keeps an explicit test.

## Behaviour

Accepted: `class ServicioÚltimaVez`, `class Ωmega`, `class ǅFoo`, `def enviar_últimas_interacciones`, `def 日本語`, `@日本語`, `@@クラス変数`, `$グローバル`, and `class Àbc` in ISO-8859-1.

Rejected, matching Ruby: `class αlpha` (Greek `α` is lowercase), `class 日本語` (kanji is `Other_Letter`), any identifier containing a byte invalid under the active encoding, and `\x80` under ASCII-8BIT.

## Test plan

- [x] `rake test` — 973 tests, 0 failures. Two errors also present on `master`, unrelated.
- [x] `rake validate`, `rake confirm_lexer` (re2c 4.5.1)
- [x] `cargo test --lib type_name` — 16 tests, 0 failures. Building it at all needs a temporary local patch for a pre-existing missing match arm in `src/ast/convert.rs`, unrelated to this PR and not part of this branch.
- [x] New tests cover each row of the behaviour table above, and assert `name.kind` — `TypeName#==` does not compare it, which is how the titlecase defect survived review once.

## Known gaps, not addressed here

- The vendored uppercase table in `src/util/rbs_encoding.c` predates the host Ruby's Onigmo, so ~55 recently added `Lu` code points lex as `tLIDENT` and cannot name a class. Prism agrees with the host Ruby, so the lexer is the odd one out. This grows with each Unicode release.
- For EUC-JP and Windows-31J, the C per-encoding tables mark fullwidth capitals as uppercase, so the lexer and `TypeName#kind` can still disagree.

### Before

```
$ bundle exec ruby benchmark/benchmark_new_env.rb
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [arm64-darwin25]
             new_env     2.000 i/100ms
       new_rails_env     1.000 i/100ms
             new_env     24.444 (± 4.1%) i/s   (40.91 ms/i) -    246.000 in  10.063843s
       new_rails_env      3.612 (± 0.0%) i/s  (276.84 ms/i) -     37.000 in  10.242942s

$ bundle exec ruby benchmark/memory_new_env.rb
Total allocated: 15205605 bytes (186884 objects)
Total retained:  10439464 bytes (97929 objects)
```

### After

```
$ bundle exec ruby benchmark/benchmark_new_env.rb
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM [arm64-darwin25]
             new_env     2.000 i/100ms
       new_rails_env     1.000 i/100ms
             new_env     25.025 (± 4.0%) i/s   (39.96 ms/i) -    252.000 in  10.070082s
       new_rails_env      3.629 (± 0.0%) i/s  (275.58 ms/i) -     37.000 in  10.196490s

$ bundle exec ruby benchmark/memory_new_env.rb
Total allocated: 15205605 bytes (186884 objects)
Total retained:  10439464 bytes (97929 objects)
```


soutaro:
Implemented in https://github.com/ruby/rbs/pull/3082. 🙇‍♂️

PR #3029, opened 2026-07-21
Fix nested declarations in rbs prototype rbi

infiton:
## Summary

Fix `rbs prototype rbi` so nested Sorbet RBI declarations produce valid, semantically accurate RBS.

The converter previously flattened every class and module into a top-level declaration while leaving relative references unchanged. For example:

```ruby
module Demo
  class Parent; end
  class Child < Parent; end
end
```

was emitted as `class Demo::Child < Parent`, where `Parent` no longer resolves relative to `Demo`. The same problem affected mixins and types in method signatures.

This PR preserves the RBI's lexical declaration nesting using the existing nested-declaration representation in the RBS AST. It also fixes the related conversion bugs found in the same reproduction.

## Changes

- Preserve nested class and module declarations instead of flattening them.
  - Relative superclasses, mixins, and signature types retain their lexical resolution context.
  - Absolute and explicitly qualified declaration names remain intact.
- Emit nested constant declarations in their containing declaration.
  - This prevents duplicated paths such as `Demo::Demo::Modes::VALUE`.
- Ignore the Sorbet-only `T::Helpers` extension, alongside `T::Sig` and `T::Generic`.
- Convert `T::Class[...]` to `untyped`.
  - RBS has no exact generic equivalent for an arbitrary class object whose instances have the specified type.
  - Falling back avoids emitting an unresolved `T::Class` while not inventing an unsupported structural constraint.
- Track `class << self` context so contained definitions become singleton methods.
- Convert `attr_reader`, `attr_writer`, and `attr_accessor`, consuming their pending `sig`.
  - This prevents an attribute signature from becoming an overload on the following method.
- Track declaration visibility and emit `private`/`public` transitions.
  - RBS does not model `protected`, so it uses the conservative `private` fallback.

## Compatibility

This only changes RBI-to-RBS prototype conversion. It does not change the RBS language, parser, validator, or the meaning of existing `.rbs` files.

`RBS::Prototype::RBI#decls` now returns a nested AST for nested RBI declarations instead of a flattened list. For an empty declaration such as `module Foo; module Bar; end; end`, both forms define the same RBS constants, but callers that inspect the AST structure directly will observe the corrected shape. For declarations containing relative references, retaining that nesting is required for semantic correctness.

## Tests

Regression tests cover each behavior independently and load the complete reported reproduction into an RBS environment, building its instance and singleton definitions.

```text
bundle exec rake test TEST=test/rbs/rbi_prototype_test.rb
33 tests, 33 assertions, 0 failures, 0 errors

bundle exec rubocop lib/rbs/prototype/rbi.rb test/rbs/rbi_prototype_test.rb
2 files inspected, no offenses detected
```

### DIFF
--- a/lib/rbs/prototype/rbi.rb
+++ b/lib/rbs/prototype/rbi.rb
@@ -9,10 +9,14 @@ module RBS
       attr_reader :modules
       attr_reader :last_sig
 
+      Context = Struct.new(:singleton, :visibility, keyword_init: true)
+
       def initialize
         @decls = []
 
         @modules = []
+        @contexts = []
+        @emitted_visibility = {}
       end
 
       def parse(string)
@@ -20,19 +24,17 @@ module RBS
         process RubyVM::AbstractSyntaxTree.parse(string), comments: comments
       end
 
-      def nested_name(name)
-        (current_namespace + const_to_name(name).to_namespace).to_type_name.relative!
-      end
-
-      def current_namespace
-        modules.inject(Namespace.empty) do |parent, mod|
-          parent + mod.name.to_namespace
+      def append_decl(decl)
+        if current_module
+          current_module.members << decl
+        else
+          decls << decl
         end
       end
 
       def push_class(name, super_class, comment:)
         class_decl = AST::Declarations::Class.new(
-          name: nested_name(name),
+          name: const_to_name(name),
           super_class: super_class && AST::Declarations::Class::Super.new(name: const_to_name(super_class), args: [], location: nil),
           type_params: [],
           members: [],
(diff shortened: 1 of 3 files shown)


soutaro:
@infiton Thank you for opening this PR. 🙏 

I pushed a fix for visibility handling of singleton members. RBS visibility sections (`private`/`public` on their own line) don't apply to singleton members — only member level visibility does — so the sections this PR emitted around `def self.` had no effect, and the `private` from a `class << self` body was misleading. Singleton members now get their own visibility instead, as `private def self.foo: () -> void`, and a test covering methods and attributes comes with it.

PR #3050, opened 2026-07-30
Add EnvironmentLoader for ruby-rbs crate

dak2:
## Summary

Mirrors RBS::EnvironmentLoader: resolves core, library, and explicit-directory sources in Ruby's load order, then parses each .rbs file once into an Environment.

https://github.com/ruby/rbs/issues/2954

## Architecture differences from `RBS::EnvironmentLoader`

- Ruby's loader both *resolves* what to load (gems, versions, manifest dependencies) and *reads* it.
- The Rust one only reads: it takes resolved directories, so everything needing RubyGems knowledge stays in Ruby, and Rust does the walking, parsing, and interning.
- That drops the dependency machinery, and makes the name pool per-`Environment` state instead of a process-global one.

| | Ruby | Rust |
|---|---|---|
| Library resolution | resolves name/version → directory via `gem_sig_path` / `Repository`, and expands `manifest.yaml` dependencies | `add_library(name, path)` takes an already-resolved directory; no `Repository`, no manifest expansion, no implicit `stringio` |
| Registration API | one `add(path:/library:/version:)` with keyword modes | `add_library` / `add_dir`, the modes separated by type |
| Duplicate registration | `libs` is a `Set`, to stop dependency expansion from recursing | plain `Vec`; with no expansion, first-wins on `seen_files` is enough |
| Name pool | process-global symbols and `TypeName` objects | `Environment` owns the `Interners`, so loaded names stay resolvable for its lifetime |
| File walk | `Pathname.glob("**/*.rbs")`, sorted as `/`-joined strings | recursive `read_dir`, sorted by platform path; symlinked directories are not followed, and only `PermissionDenied` / `NotFound` are skipped |

## Usage

Register the directories on the loader, then load them into an `Environment`:

```rust
let loader = EnvironmentLoader::new(Some(core_root))
    .add_library("pathname", pathname_sig_dir) // resolved on the Ruby side
    .add_dir(PathBuf::from("sig"));

let env = Environment::from_loader(&loader)?;

for source in env.sources() {
    // source.path, source.kind (Core / Library { name, path } / Dir { path }),
    // source.directives, source.declarations
}
```

`load` can also fill an existing `Environment`, and returns what it read, in the
order it read it:

```rust
let mut env = Environment::new();
let loaded: Vec<LoadedFile> = loader.load(&mut env)?; // { path, kind } per file
```

### Notes

- Directories are visited core → libraries → dirs, and a path already read is skipped, so the first source registering a file wins.
- `LoadError` is `Io { path, .. }` or `Parse { path, message }`; the sources read before the failure are already in the `Environment`.
- The `Environment` owns the interners, so `env.interners()` is what resolves and displays the interned names held by the loaded declarations.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/rust/ruby-rbs/src/environment/source.rs
+++ b/rust/ruby-rbs/src/environment/source.rs
@@ -1,0 +1,35 @@
+use std::path::PathBuf;
+
+use crate::ast::{Declaration, Directive};
+
+/// Corresponds to the `source` values yielded by
+/// `RBS::EnvironmentLoader#each_dir`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SourceKind {
+    Core,
+    /// `path` is the library's version directory, already resolved by the
+    /// caller (Ruby's `Repository#lookup` / `gem_sig_path`), holding the
+    /// library's RBS files directly. Version alone could not recover where a
+    /// library's signatures came from; the path can.
+    Library {
+        name: String,
+        path: PathBuf,
+    },
+    Dir {
+        path: PathBuf,
+    },
+}
+
+impl SourceKind {
+    pub(crate) fn skips_hidden(&self) -> bool {
+        !matches!(self, SourceKind::Dir { .. })
+    }
+}
+
+#[derive(Debug)]
+pub struct Source {
+    pub path: PathBuf,
+    pub directives: Vec<Directive>,
+    pub declarations: Vec<Declaration>,
+    pub kind: SourceKind,
+}
--- a/rust/ruby-rbs/src/environment/source.rs
+++ b/rust/ruby-rbs/src/environment/source.rs
@@ -1,0 +1,33 @@
+use std::path::PathBuf;
+
+use crate::ast::{Declaration, Directive};
+use crate::buffer::Buffer;
+
+/// Where a loaded signature file came from, corresponding to the `source`
+/// values yielded by `RBS::EnvironmentLoader#each_dir`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SourceKind {
+    /// Core library signatures (`:core` in Ruby).
+    Core,
+    /// A library resolved through the repository or a `GemSigResolver`.
+    ///
+    /// Carries the requested version as well as the name: Ruby keys its
+    /// library set on the `(name, version)` pair, so `uri` and `uri` 1.0 are
+    /// distinct entries yielded separately by `each_dir`.
+    Library {
+        name: String,
+        version: Option<String>,
+    },
+    /// An explicitly added signature directory.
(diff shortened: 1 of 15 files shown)


rust/ruby-rbs/src/loader/mod.rs
```diff
@@ -0,0 +1,397 @@
...
+    core_root: Option<PathBuf>,
+    stdlib_root: Option<PathBuf>,
+    repository: Repository,
+    libs: Vec<Library>,
+    dirs: Vec<PathBuf>,
+    resolver: Box<dyn GemSigResolver>,
+}
+
+impl EnvironmentLoader {
+    pub fn new() -> Self {
```
soutaro: Make `core_root` and `stdlib_root` required arguments.

rust/ruby-rbs/src/environment/source.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// Core library signatures (`:core` in Ruby).
+    Core,
+    /// A library resolved through the repository or a `GemSigResolver`.
+    ///
+    /// Carries the requested version as well as the name: Ruby keys its
+    /// library set on the `(name, version)` pair, so `uri` and `uri` 1.0 are
+    /// distinct entries yielded separately by `each_dir`.
+    Library {
+        name: String,
+        version: Option<String>,
```
soutaro:
```suggestion
        path: Path,  // or some Path-ish attribute
```

Ruby runtime will fill the path for the library root, where adding `sig` dir name gives the path for the directory of RBS files.

rust/ruby-rbs/src/loader/gem_sig_resolver.rs
```diff
@@ -0,0 +1,27 @@
...
+///
+/// The Ruby implementation asks `Gem::Specification`; Rust tools inject their
+/// own strategy here (spawning Ruby, reading a cache, ...). The default
+/// [`NoGemSigs`] never finds anything, so only the repository is consulted.
+///
+/// Implementations must be `Send + Sync`. Loading is planned to run its parse
+/// stage on worker threads, which requires the loader holding this resolver to
+/// be `Sync`; a supertrait cannot be added later without breaking downstream
+/// implementations, so the bound is here from the start.
+pub trait GemSigResolver: Send + Sync {
```
soutaro: I think this `trait` is unnecessary. The library client would pass the resolved paths of each loaded gems.

dak2:
Remove Buffer. In details, https://github.com/ruby/rbs/pull/3050/changes/2b3df5a413e0f1022c7a7ece6ed8bc6bfde70a4f

soutaro (approved):
👏

PR #3063, opened 2026-08-03
Say how the changelog is assembled without `gh`

soutaro:
`gem:changelog` and `gem:changelog:json` go through `gh`, which a Claude Code on the web session cannot reach. `api.github.com` is blocked at the agent proxy for anything the shell does — `gh` is not installed, installing it does not help, and rewriting the tasks against REST or Net::HTTP would be refused the same way, since the 403 is keyed on the session rather than on the client. The rest of the release is unaffected: `gem:check_release` and `gem:tag` read git and the working tree, and `gem:gh_release` runs on a runner.

So `docs/release.md` gains a section for the route that does work there — the GitHub MCP server — written as the three steps the task itself takes: resolve the base tag with `git describe`, list the commits with `git log`, then match the pull requests by `head.sha` against `list_pull_requests`. That intersection is what the GraphQL `associatedPullRequests` query answers, reached from the other side.

Most of the section is the ways it goes wrong, because the obvious shortcut is a trap. Reading the numbers out of `Merge pull request #N` subjects looks like it works on this repository and loses pull requests silently: the path filter the task uses (`--full-history --simplify-merges -- . ':(exclude)rust'`) drops the merge commits while keeping the commits they merged, so five of the eight numbers of the 4.1.2 cycle disappear with them, and a squashed or rebased pull request never writes that subject at all. Also documented: `head.sha` only lands in the history because this repository merges with merge commits, the listing reports `merged: false` for merged pull requests so `merged_at` is the field to read, backports need their `(cherry picked from commit <sha>)` trailer resolved first as `changelog_origins` does, and the `rust/` exclusion belongs on the pull requests rather than on the commit list.

Verified against the 4.1.2 cycle: the procedure yields exactly #3053–#3061, with #3052 correctly falling outside `v4.1.1`. This is the list #3062 was written from.

Docs only, no code change.

---
_Generated by [Claude Code](https://claude.ai/code/session_014QUAgW2m9wsYE1cLM1fFbw)_

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -77,6 +77,10 @@ on their own GitHub releases.
 Sort the list into the sections below. `rake gem:changelog:json` prints the same pull requests with
 the changed files, labels, and body of each, which is what the sorting is based on.
 
+Both tasks reach GitHub through `gh`, which a Claude Code on the web session cannot do. See
+[Assembling the changelog without `gh`](#assembling-the-changelog-without-gh) for how the same list
+is produced there.
+
 ```markdown
 ## X.Y.Z (YYYY-MM-DD)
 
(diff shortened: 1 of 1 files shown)


docs/release.md
```diff
@@ -77,6 +77,10 @@ on their own GitHub releases.
 Sort the list into the sections below. `rake gem:changelog:json` prints the same pull requests with
 the changed files, labels, and body of each, which is what the sorting is based on.
 
+Both tasks reach GitHub through `gh`, which a Claude Code on the web session cannot do. See
```
ParadoxV5, line 80: Ruby RBS has officially switched to AI vibing and botsitting.
No longer is Claude merely a tool, for the dev process now revolves around how it takes the wheel.

PR #3069, opened 2026-08-06
Exclude CR from comment tokens to fix an off-by-one Location#end_line on CRLF files

tufusa:
## Background

Currently, the pattern for comment tokens is `"#" (. \ [\x00\uFFFD])*`:
https://github.com/ruby/rbs/blob/d89e4ad62914119466910c2cf742d3a3ceb572a2/src/lexer.re#L61-L66

This works correctly in LF environments, but the `.` pattern excludes only LF, not CR; therefore `\r` is included at the end of the comment token in CRLF environments such as Windows.

As a result, the value of `Location#end_line` for comments ends up being one line greater than actually is. This occurs because while `Buffer#ranges`
https://github.com/ruby/rbs/blob/d89e4ad62914119466910c2cf742d3a3ceb572a2/lib/rbs/buffer.rb#L42-L49
correctly removes `\r\n` (L44), `Buffer#pos_to_loc` which is passed a value for `pos` that is one greater than the actual value
https://github.com/ruby/rbs/blob/d89e4ad62914119466910c2cf742d3a3ceb572a2/lib/rbs/buffer.rb#L56-L59
incorrectly finds the next line.

This will cause incorrect behavior in tools built on `RBS::Parser.lex`, for example, `RBS/Layout/ExtraSpacing` cop in rubocop-on-rbs (https://github.com/ksss/rubocop-on-rbs/issues/152; in fact, I discovered this issue while investigating the false detection of this cop).

This token behavior itself existed prior to RBS 4, but it looks to become apparent because, starting with RBS 4, `String#chomp` is called before the character count is performed within `Buffer#ranges`.

## Reproduction

```ruby
require 'rbs'

src = "class Klass\r\n  # Comment\r\nend\r\n"
buffer = RBS::Buffer.new(name: Pathname("test.rbs"), content: src)
tokens = RBS::Parser.lex(buffer).value

comment = tokens.find { _1.type == :tLINECOMMENT }
puts comment.location.end_line # => 3 (if LF only: 2)
```

- ruby 4.0.5
- rbs 4.1.2

## Changes

- Added `\r` to the pattern's subtraction group: `"#" (. \ [\x00\r\uFFFD])*`. This causes the comment token to stop reading as soon as `\r` appears.
- Added the test `RBS::ParserTest#test__lex_crlf` next to `test__lex`. This verifies that `\r` is treated as a single independent `TRIVIA` token as usual, even in the case of comments.

Thank you!

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -129,7 +129,7 @@ rbs_token_t rbs_lexer_next_token(rbs_lexer_t *lexer) {
         }
     yy1:
         rbs_skip(lexer);
-#line 175 "src/lexer.re"
+#line 184 "src/lexer.re"
         {
             return rbs_next_eof_token(lexer);
         }
@@ -137,7 +137,7 @@ rbs_token_t rbs_lexer_next_token(rbs_lexer_t *lexer) {
     yy2:
         rbs_skip(lexer);
     yy3:
-#line 176 "src/lexer.re"
+#line 185 "src/lexer.re"
     {
         return rbs_next_token(lexer, ErrorToken);
     }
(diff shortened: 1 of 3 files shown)


soutaro (approved):
Thanks!

PR #3070, opened 2026-08-07
Extract per-declaration type param alignment into entry-level align_params

soutaro:
Stacked on #3067.

The substitution that renames a declaration's type parameters to the entry's type parameters was built inline in five places: `MethodBuilder#build_instance`, `DefinitionBuilder#define_instance`, `AncestorBuilder#mixin_ancestors`, `ModuleEntry#self_types`, and `AncestorBuilder#validate_super_class!`. This PR defines it once as `ModuleEntry#align_params` / `ClassEntry#align_params` and uses it from all of them, making the decl-to-entry name translation a defined entry-level operation instead of a copy-pasted idiom. The method returns `nil` when the declaration already uses the entry's type parameter names — the common case — so callers skip the substitution entirely.

https://claude.ai/code/session_01TBY8ct4HpkVkZNPdsNHEDE

---
_Generated by [Claude Code](https://claude.ai/code/session_01TBY8ct4HpkVkZNPdsNHEDE)_

### DIFF
--- a/lib/rbs/definition_builder.rb
+++ b/lib/rbs/definition_builder.rb
@@ -124,10 +124,13 @@ module RBS
       end
 
       entry = env.class_decls[type_name] or raise "Unknown name for build_instance: #{type_name}"
-      args = entry.type_params.map {|param| Types::Variable.new(name: param.name, location: param.location) }
 
       entry.each_decl do |decl|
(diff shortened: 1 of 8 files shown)


soutaro:
Folded into #3067.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3081, opened 2026-08-11
Remove deprecated `json/add/*` types

byroot:
I honestly don't know much about RBS, so unsure what exactly should be done.

But all these `json/add/*.rb` methods are gone in JSON 3.x: https://github.com/ruby/json/releases/tag/v3.0.0.rc1

So it's causing issues on ruby's master branch:  https://github.com/ruby/ruby/pull/18295


soutaro:
Thanks @byroot!
I think skipping tests with JSON 3.x would be better because the RBS tests should run with released versions of Ruby with JSON 2.x. https://github.com/ruby/rbs/pull/3084

byroot:
Up to you, I don't have much context on RBS.

PR #3102, opened 2026-08-26
Use RUBY for CLI test subprocesses

soutaro:
`RbConfig.ruby` points at Ruby's configured installation path, which may not exist while ruby/ruby tests an uninstalled build. Prefer the `RUBY` environment variable supplied by its test runner, with `RbConfig.ruby` as the fallback.

### DIFF
--- a/test/rbs/cli_test.rb
+++ b/test/rbs/cli_test.rb
@@ -33,7 +33,8 @@ class RBS::CliTest < Test::Unit::TestCase
   # with a version mismatch error when `RUBYLIB` points at the standard library of the running Ruby.
   # That is what happens in ruby/ruby CI, where the tests run with a freshly built `ruby`.
   #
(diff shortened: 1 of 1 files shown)


test/rbs/cli_test.rb
```diff
@@ -33,7 +33,8 @@ def refute_cli_success(exit_status = nil, &)
   # with a version mismatch error when `RUBYLIB` points at the standard library of the running Ruby.
   # That is what happens in ruby/ruby CI, where the tests run with a freshly built `ruby`.
   #
-  BUNDLE_COMMAND = [RbConfig.ruby, "-S", "bundle"]
+  RUBY_EXECUTABLE = ENV["RUBY"] || RbConfig.ruby
```
amomchilov, line 36: Can you update the comment above to explain (what you already have in the PR description)

Issue #348, opened 2020-07-30
RubyVM usage


eregon:
Hello, it seems that currently `rbs` uses `RubyVM::AbstractSyntaxTree` in a couple places:
https://github.com/ruby/rbs/search?q=RubyVM&unscoped_q=RubyVM

However, RubyVM by design is MRI-only and is expected to not exist as a constant on alternative Ruby implementations ([documentation](https://rubydoc.info/stdlib/core/RubyVM)).

Do you plan to address this somehow?
If `AbstractSyntaxTree` is needed by `rbs` then I think it's time to move it somewhere outside of `RubyVM`.

For instance, when running on TruffleRuby:
```
% rbs prototype rb lib/person.rb lib/email.rb lib/phone.rb
/Users/bfish/Documents/truffleruby-ws/truffleruby/mxbuild/truffleruby-native/jre/languages/ruby/lib/gems/gems/rbs-0.7.0/lib/rbs/prototype/rb.rb:55:in `const_missing': uninitialized constant RBS::Prototype::RB::RubyVM (NameError)
	from /Users/bfish/Documents/truffleruby-ws/truffleruby/mxbuild/truffleruby-native/jre/languages/ruby/lib/gems/gems/rbs-0.7.0/lib/rbs/prototype/rb.rb:55:in `parse'
	from /Users/bfish/Documents/truffleruby-ws/truffleruby/mxbuild/truffleruby-native/jre/languages/ruby/lib/gems/gems/rbs-0.7.0/lib/rbs/cli.rb:629:in `block in run_prototype_file'
	from /Users/bfish/Documents/truffleruby-ws/truffleruby/mxbuild/truffleruby-native/jre/languages/ruby/lib/gems/gems/rbs-0.7.0/lib/rbs/cli.rb:628:in `each'
	from /Users/bfish/Documents/truffleruby-ws/truffleruby/mxbuild/truffleruby-native/jre/languages/ruby/lib/gems/gems/rbs-0.7.0/lib/rbs/cli.rb:628:in `run_prototype_file'
	from /Users/bfish/Documents/truffleruby-ws/truffleruby/mxbuild/truffleruby-native/jre/languages/ruby/lib/gems/gems/rbs-0.7.0/lib/rbs/cli.rb:526:in `run_prototype'
[...]

PR #3105, opened 2026-08-26
Port prototype rbi generation to prism

Earlopain:
Apart from porting to prism, this also does the following:

1. Remove `variable` tracking for `type_of0`.
  It contains `AST::TypeParam` but checked for inclusion of a Symbol. There's no difference in the output even when this is fixed, so I just removed it entirely
1. Have the parse method return declarations, make it a class method.
  Just more convenient with the new structure. Eventually `rb prototype` will do this as well
1. Allow to use it on jruby/truffleruby.
1. Split comment parsing from comment processing in the helper.
  In rbi the comments now come from a plain parse. When `prototype rb` uses prism as well, `parse_comments` can be removed (as well as most of the other helper methods there)

I tested this against code samples from https://github.com/Shopify/tapioca/blob/d029cc9c3f76865f61fdbaff68d75e18a0014764/spec/tapioca/gem/pipeline_spec.rb
The output is largely the same, and improved in some areas. For example, `type_member` with no paren is no longer considered an untyped constant. Previously that was only the case for `type_member()` or when an argument was passed like `type_member(:out)`.

For https://github.com/ruby/rbs/issues/348

### DIFF
--- a/lib/rbs/cli.rb
+++ b/lib/rbs/cli.rb
@@ -108,7 +108,8 @@ module RBS
       opts
     end
 
-    def has_parser?
+    def has_parser?(format)
+      return true if format == "rbi"
       defined?(RubyVM::AbstractSyntaxTree) ? true : false
     end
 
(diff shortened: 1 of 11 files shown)


Earlopain:
cc @eregon

ksss (approved):
👍

PR #3106, opened 2026-08-27
Bump net-protocol from 0.2.2 to 0.3.0

dak2:
## Problem

`test (head, stdlib_test rubocop)` fails on ruby-head with 8 errors, all in Net::HTTP / open-uri ([run](https://github.com/ruby/rbs/actions/runs/33056526924/job/98464589988)):

```
ArgumentError: wrong number of arguments (given 3, expected 1..2)
  net-protocol-0.2.2/lib/net/protocol.rb:194:in 'Net::BufferedIO#readuntil'
  ruby-head/lib/ruby/4.1.0+4/net/http/response.rb:164:in 'Net::HTTPResponse.read_line'
```

## Cause

ruby/net-http#325 (in ruby/ruby as https://github.com/ruby/ruby/commit/331568b2a) made `Net::HTTPResponse` use `readuntil(limit:)`, which requires net-protocol >= 0.3.0.

ruby-head ships 0.3.0 as a default gem, so it is consistent on its own. But `net/http` is a default gem too, so its dependency never reaches Bundler's resolution — our lock pinned 0.2.2 through `net-smtp`, which carries no lower bound, and the installed 0.2.2 shadowed ruby-head's 0.3.0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3107, opened 2026-08-27
Declare Net::ReadLimitExceeded added in net-protocol 0.3.0

dak2:
## What

net-protocol 0.3.0 adds `Net::ReadLimitExceeded`, raised when `Net::BufferedIO#readuntil` does not find the terminator within the byte limit it was given. Declare it in `stdlib/net-protocol/0/net-protocol.rbs`.

Added in ruby/net-protocol#67.

## Notes

`readuntil` itself gained a `limit:` keyword in the same release, but it needs no signature update: `Net::BufferedIO` is `:nodoc: internal use only` and is not declared in RBS.

The other addition in 0.3.0, `Net::Protocol::TCP_SOCKET_NEW_HAS_OPEN_TIMEOUT`, is a `private_constant`.

Independent of #3106, which bumps the lock to 0.3.0.

### DIFF
--- a/stdlib/net-protocol/0/net-protocol.rbs
+++ b/stdlib/net-protocol/0/net-protocol.rbs
@@ -27,6 +27,13 @@ module Net
   class ProtoRetriableError < ProtocolError
   end
 
+  # <!-- rdoc-file=lib/net/protocol.rb -->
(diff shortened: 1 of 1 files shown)


ksss (approved):
Thanks!

PR #3111, opened 2026-08-29
Use portable collection paths in cleaner

OskarEichler:
Summary: derive collection gem and version names through Pathname components instead of splitting paths on a slash. This preserves cleaning behavior on Windows-style path implementations.

### DIFF
--- a/lib/rbs/collection/cleaner.rb
+++ b/lib/rbs/collection/cleaner.rb
@@ -11,9 +11,8 @@ module RBS
 
       def clean
         lock.repo_path.glob('*/*') do |dir|
-          *_, gem_name, version = dir.to_s.split('/')
-          gem_name or raise
-          version or raise
+          gem_name = dir.parent.basename.to_s
+          version = dir.basename.to_s
           next if needed? gem_name, version
 
           case


pocke:
Thank you for the patch. I would like to merge this, but as a refactoring rather than as a portability fix, so let me record why.

I tried to build the case the description points at and could not find one. `Dir.glob` returns `/`-separated paths on Windows too, and `Pathname#glob` assembles its results through `Pathname#+`, which also joins with `/`. Even if `path:` in `rbs_collection.yaml` were written with backslashes, only the leading component would carry them, and `*_, gem_name, version = dir.to_s.split("/")` takes the last two components, which are always separated by `/`. As far as I can tell the old and the new code produce the same `gem_name` and `version` on every platform we support.

That said, the new form reads better. `dir.parent.basename` and `dir.basename` say what they mean directly, and the two bare `raise`s that go away could only have fired in situations the `*/*` glob pattern already rules out. That is reason enough to take it, so I am merging it on those grounds.

I have rebased the branch onto the current master. The `test (head, stdlib_test rubocop)` failure was an unrelated `Net::HTTP` test on ruby-head and is now fixed there.

If you do have a case where the old code actually breaks, please tell me. That would change how we describe the change, and it would be worth a test.

PR #3112, opened 2026-08-29
Handle disabled collection configuration

OskarEichler:
Summary: let collection help work with disabled collection configuration and return a clear failure for commands that require configuration, instead of raising from a missing path.

### DIFF
--- a/lib/rbs/cli.rb
+++ b/lib/rbs/cli.rb
@@ -1084,7 +1084,15 @@ EOB
       opts = collection_options(args)
       params = {} #: Hash[Symbol, untyped]
       opts.order args.drop(1), into: params
-      config_path = options.config_path or raise
+      if %w[help hel he h].include?(args[0])
+        stdout.puts opts.help
+        return 0
+      end
+
+      unless config_path = options.config_path
+        stderr.puts "`rbs collection` cannot be used with `--no-collection`."
+        return 1
+      end
       lock_path = Collection::Config.to_lockfile_path(config_path)
 
       case args[0]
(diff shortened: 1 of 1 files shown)


pocke:
Thank you for the patch, and sorry for the slow reply. I am going to close this one.

You are right that the current code is bad: `rbs --no-collection collection install` dies with a bare `RuntimeError` carrying an empty message, and I reproduced that. But I do not think the combination is worth supporting.

`--no-collection` is not an option of the `collection` subcommand. It belongs to `LibraryOptions`, and its only job is to make `LibraryOptions#loader` skip `add_collection`, so that `rbs list`, `rbs ast`, `rbs validate` and friends can build an environment without pulling in the gem RBS collection. `run_collection` only sees the nil because every subcommand is handed the same options object. `--no-collection` and the `config_path = options.config_path or raise` line were in fact added in the same commit (38fd8b2d, the one that introduced `rbs collection`), so combining the two was never something we meant to support. Asking `rbs collection` to run while telling rbs to ignore the collection configuration does not describe anything a user wants.

On the `help` branch specifically: I do not see what it buys us. `rbs collection help` already works without `--no-collection`, and nobody is going to pass `--no-collection` in order to read the help text. Lifting one branch out of the `case` to serve that combination costs more in readability than the behaviour is worth.

So we would be adding code, and a user-facing error message, for an invocation that has no meaning. I would rather leave the surface as it is.

If you have run into this for real -- a wrapper or an alias that always passes `--no-collection`, say -- please tell me. A concrete case would change my mind, and then I would want the fix to be just the nil check, without the `help` special case.

PR #3121, opened 2026-08-29
Preserve missing accessor halves when subtracting

OskarEichler:
Summary: split an accessor when only its reader or writer already exists, so subtraction preserves the missing half and its metadata.

### DIFF
--- a/lib/rbs/subtractor.rb
+++ b/lib/rbs/subtractor.rb
@@ -50,11 +50,44 @@ module RBS
 
       context = _ = [context, decl.name]
       children = call(decl.each_decl.to_a, context: context) +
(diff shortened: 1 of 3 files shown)


OskarEichler:
The existing accessor test exposed the intended changed output but still asserted that the whole accessor disappeared. I updated it in 8e1d830f and expanded the case to cover reader-only, writer-only, and complete subtraction. The focused subtractor suite passes (23 tests), targeted RuboCop is clean, and `git diff --check` passes. The remaining `ruby-head` stdlib failures are shared upstream drift.

ksss:
I am against subtracting only a part of attr_accessor, as it may lead to unintended behavior.

Please let me know the problem you are trying to solve.

PR #3122, opened 2026-08-29
Preserve partial accessor visibility in prototypes

OskarEichler:
Summary: split an accessor when a visibility change targets only its reader or writer, preserving the unaffected half and the requested visibility.


ksss:
Please add test code for the issue you want to achieve.

PR #3128, opened 2026-08-29
Preserve runtime prototype keyword rest names

OskarEichler:
Summary: retain named keyword-rest parameters in runtime prototypes while continuing to normalize the anonymous forwarding sentinel.

### DIFF
--- a/lib/rbs/prototype/runtime.rb
+++ b/lib/rbs/prototype/runtime.rb
@@ -211,7 +211,8 @@ module RBS
             name or raise
             optional_keywords[name] = Types::Function::Param.new(name: nil, type: untyped)
           when :keyrest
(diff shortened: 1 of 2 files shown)


OskarEichler:
The implementation correctly changes runtime prototype output for named keyword-rest parameters, but the existing focused expectations still asserted the old anonymous form. I updated both merge and non-merge expectations in f7fa06f6. The focused runtime prototype suite now passes (18 tests, 27 assertions, 1 expected omission), targeted RuboCop is clean, and `git diff --check` passes. The remaining `ruby-head` stdlib errors are shared upstream drift and unrelated to this change.

ksss:
`rbs prototype runtime` is deprecated. See https://github.com/ruby/rbs/issues/2841

PR #3129, opened 2026-08-29
Preserve visibility when sorting members

OskarEichler:
Summary: keep instance attributes in their public/private groups and encode singleton visibility on the singleton members that it governs, avoiding invalid scope transfer.


OskarEichler:
I updated the sorter expectation to match the intended inline singleton visibility output and added explicit coverage for both private singleton and private instance attributes in 864b5ad1. The focused sorter test now passes, targeted RuboCop is clean, and `git diff --check` passes. This addresses the real cross-version test failures; the separate `ruby-head` stdlib failures are upstream drift shared by the other RBS PRs.

PR #3152, opened 2026-09-20
Update singleton

ksss:
Update the singleton module's type definitions to reflect actual usage.

In the current definitions, even the `#instance` method in standard use cases with `include Singleton` does not work correctly.

This is a breaking change for calls to `Singleton.instance`, but this method does not exist in the first place and type checks should not pass for it.

To provide correct type definitions, `extend Singleton::SingletonClassMethods` is required alongside `include Singleton` in each case, which is also the correct behavior.


test/stdlib/singleton_test.rb
```diff
@@ -11,11 +11,6 @@ class TestClass
     include Singleton
   end
 
-  def test_instance
-    assert_send_type  '() -> SingletonSingletonTest::TestClass',
-                      TestClass, :instance
-  end
-
```
ksss, line 18: This test will fail, but making it pass requires breaking changes to the test harness.

